PR #26, opened 2025-10-14
rework autosegs to use global constructors instead of scanning segments, removing all platform-dependent code

RicardoLuis0:
(TODO rebase)


Boondorl:
Looks like it just needs to be rebased atm

PR #35, opened 2025-10-14
[Feature] Image Calibration Screen

the-phinet:
TODO (for this PR):

- [x] Reset relevant settings upon update from older version
- [x] Fix gles
- [x] Verify 3d outputs
- [x] Verify GrayFormula=0/2 pipelines
- [x] Look into HDR
- [x] Add translation strings to language files (#797)
- [x] Verify mac gamma on an apple panel (2.2 is correct. 1.8 was used on old macs aparently)

TODO (for future PR):

- ~~Nicer menu?~~ It's pretty decent now

---

After a lot of discussion in the discord, it has become clear that a proper image calibration screen would be very useful.

Right now, gzdoom's gamma correction is very "Vibes Based", where the user sets the gamma based on what they feel looks nice. In most other applications, gamma correction is used to calibrate the image to the output device, in order to display things as the author intended.

WIP Video Demo (play video in a new tab at its native scale for the test pattern to look correct):

[Screencast_20251214_175556.webm](https://github.com/user-attachments/assets/2b759955-9349-4a75-9dfa-3efdc89c080f)

<details><summary>Proposal / Old mockups</summary>

I propose adding a "Image calibration screen" instead of the gamma slider we currently have. The screen would take control of the renderer directly to display a test pattern (and maybe a demo render), allowing the user to adjust the output parameters accordingly.

The screen would have 3 sliders: Gamma, Black Point, White Point
- Gamma controls the gamma correction. (default 2.2)
- Black point adjust what the engine considers the darkest black. This is to fix the image on monitors with bad black levels.  (default 0)
- White point does the same for as black point, but for the brightest white. (default 1)

Here is a quick-and-dirty demo of how I think the params should work. 
[Demo](https://phinet.ca/projects/gamma/)
[Source](https://github.com/the-phinet/Image-Calibration-Toy/)

Here is an image of how I think the screen should kinda look:

Image

(if you see a pattern in this mock-screenshot, github probably resized the image. try opening it in a new tab)

Often games give you one slider and three images. "Adjust until you cant see the dark logo". I've always disliked how this is implemented, as it A: hides what it is actually doing (is it setting the black point? adjusting gamma? just boosting brightness?); B: is extremely imprecise. I think the method proposed above is robust and gameplay agnostic (dark maps vs light maps).

Imported from https://github.com/ZDoom/gzdoom/pull/3332

</details>

### DIFF
--- a/src/common/rendering/gl/gl_postprocess.cpp
+++ b/src/common/rendering/gl/gl_postprocess.cpp
@@ -202,6 +202,8 @@ void FGLRenderer::DrawPresentTexture(const IntRect &box, bool applyGamma)
 		mPresentShader->Uniforms->Contrast = 1.0f;
 		mPresentShader->Uniforms->Brightness = 0.0f;
 		mPresentShader->Uniforms->Saturation = 1.0f;
+		mPresentShader->Uniforms->BlackPoint = 0.0f;
+		mPresentShader->Uniforms->WhitePoint = 1.0f;
 	}
 	else
 	{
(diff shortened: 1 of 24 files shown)


Ralgor:
This seems to break HDR. The colorspace in HDR mode is linear, but this PR causes it to still apply a gamma of 2.2, which washes the image out.

RicardoLuis0 (approved):

PR #46, opened 2025-10-15
Revert commit 0891ab2c6ff5adca7fd131927be874888f12998c

RicardoLuis0:


MajorCooke:
Thank you. I really didn't like VKDoom's menu at all.

Boondorl:
Tbh I never felt the simple options menu worked either, it completely screws over mods that couldn't account for it. This being on by default makes it even nuttier. I do think something needs to be done about this, however. There's a lot of really...not useless features in there, but it feels like you really have to dive deep into the menus to find what you want. Definitely needs some streamlining

RicardoLuis0:
oh yeah, a new/reworked optionsmenu is needed, just not vkdoom's

PR #48, opened 2025-10-15
[Nit] Editor Configuration

the-phinet:
Adds an [.editorconfig](https://editorconfig.org/) file to allow for everyone's text editors to automatically agree on things like indentation and line endings.

Adds a (yet unused) [.clang-format](https://clang.llvm.org/docs/ClangFormat.html) file, to reduce code churn and enforce a consistent style

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,11 @@
+BasedOnStyle: Microsoft
+UseTab: AlignWithSpaces
+AlignArrayOfStructures: Right
+AlignConsecutiveAssignments: Consecutive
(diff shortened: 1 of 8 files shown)


RicardoLuis0:
can clang-format be enabled to work only on the currently-being-edited function (and not on the whole file)?

RicardoLuis0:
ah, so it's per-editor, was hoping it'd be something doable in the clang-format config itself

TehRealSalt (approved):

PR #52, opened 2025-10-15
Default autostart / skill cvar to MAPINFO's defaultskill

TehRealSalt:
Changes the following to default to MAPINFO's `defaultskill` setting:
 - Warping to maps
 - Resetting the "skill" cvar

I did this because I have a project that has 4 skills instead of 5, so the hardcoded default of 2 doesn't match. I can manually change the default in an IWAD with DEFCVARS, but I've seen lots of gameplay mods define their own skill lists, so it'd be nice to let these mods define the command line's default skill, too.

# Testing instructions

1. Run any game with any autostart parameter (`-warp`, `-episode`, etc)
2. Check "skill" in console, current skill and the default should both be 2.
3. Close, re-run with this mod: [default-skill-test.zip](https://github.com/user-attachments/files/22918130/default-skill-test.zip)
4. Check "skill" again, should be 1 now.

### DIFF
--- a/src/d_main.cpp
+++ b/src/d_main.cpp
@@ -321,6 +321,7 @@ int NoWipe;				// [RH] Allow wipe? (Needs to be set each time)
 bool singletics = false;	// debug flag to cancel adaptiveness
 FString startmap;
 bool setmap;
+int setskill = -1;
 bool autostart;
 bool advancedemo;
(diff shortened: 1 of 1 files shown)


Boondorl:
Bless up, this has annoyed me for so long

PR #54, opened 2025-10-15
KexDoom-style menu fades

Cacodemon345:
Implements KexDoom-style menu enter/exit animations, and fading in when first opening the menu.

### DIFF
--- a/src/common/2d/v_2ddrawer.cpp
+++ b/src/common/2d/v_2ddrawer.cpp
@@ -222,6 +222,7 @@ DEFINE_ACTION_FUNCTION_NATIVE(DShape2D, PushTriangle, Shape2D_PushTriangle)
 int F2DDrawer::AddCommand(RenderCommand *data) 
 {
 	data->mScreenFade = screenFade;
+	data->mAlpha = data->noAlpha ? 1.0 : Alpha;
 	if (mData.Size() > 0 && data->isCompatible(mData.Last()))
 	{
 		// Merge with the last command.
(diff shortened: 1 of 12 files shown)


mc776:
These can be turned off by the user, right?

mc776:
I just looked it up. Timing-wise it's not *too* bad https://www.youtube.com/watch?v=LAsY_z02fy0

but that movement physically makes me ill.

EDIT: Also two points of concern:

1. Aesthetically this can end up branding everything, even independent non-Doom-related projects, with the branding of the KEX Engine stuff, which can be inappropriate and confusing.
2. The KEX Engine is proprietary so hard-coding this independent of which game the player is actually running with it raises concerns.

Cacodemon345:
The initial fade-in animation as mentioned here happens rather very quickly so it shouldn't be a huge visual concern for mods implementing their own in my opinion, but I'm happy to hear others' opinion on this.

Cacodemon345:
Making it opt-in by default has a much higher chance of making the animation inconsistent across menus, since I don't think some menus (especially the Yes/No ones) are actually defined in MENUDEF.

mc776:
> Personally I think this needs to work on inverse logic, with menus needing to opt into this + having it has a user feature in the options menu.

Seconding this.

The inconsistency is tolerable pending more MENUDEFization... certainly more tolerable than having this forced on the user anyway.

RicardoLuis0:
> > Personally I think this needs to work on inverse logic, with menus needing to opt into this + having it has a user feature in the options menu.
> 
> Seconding this.
> 
> The inconsistency is tolerable pending more MENUDEFization... certainly more tolerable than having this forced on the user anyway.

it's not that they haven't been converted to menudef, is that they can't - they're messagebox menus, and as such they're written only in zscript (and handled by the MessageBoxMenu/CustomMessageBoxMenu classes)

Kroc:
It'll never be a single binary toggle, they never are. Consider down-stream distributors like devs shipping on Steam, they neither want this hard-baked into their game, nor a fixed effect. They'd need fade speed control, colour etc.

Cacodemon345:
If you need more control over the fade effects beyond fade speed control and color, I'm sorry but it's going to bloat the 2D drawer code with even more stuff most people are not going to consider.

the-phinet:
I really enjoy this, but I think the other's concerns are very valid. I'd love to see this continue being worked on

the-phinet:
﻿
﻿

PR #87, opened 2025-10-16
Actual rebrand

nashmuhandes:
Program icons are temporary placeholders, banner art is also to-be-updated in another PR

### DIFF
--- a/tools/AppImageBuilder.yml
+++ b/tools/AppImageBuilder.yml
@@ -4,9 +4,9 @@ version: 1
 
 AppDir:
   app_info:
-    id: org.zdoom.UZDoom
+    id: org.uzdoom.UZDoom
     name: UZDoom
-    icon: org.zdoom.UZDoom
+    icon: org.uzdoom.UZDoom
     version: '{{GIT_DESCRIBE}}'
     exec: ./usr/bin/uzdoom
     exec_args: $@


RicardoLuis0:
> > > Nice, I had noticed that the version.h still mentioned GZDoom and was wonder when that'd be changed. Also damn why are all the checks failing?
> > 
> > 
> > I took a minimal approach here, as in: only changes that are outwardly visible. Superficial stuff like comments are left alone. I'm not here to erase GZDoom history
> > Yeah not sure what's up with the CI, will have to investigate later
> 
> Pretty simple (I think), you renamed the project in cmake to UZdoom and the CI is expecting GZdoom to be outputted.
> 
> https://github.com/UZDoom/UZDoom/blob/bbdee80b3426934d8910b46b4d4034287bf2080b/.github/workflows/continuous_integration.yml#L99-L129

seems that was it indeed, thanks 👍

src/posix/freedesktop/org.uzdoom.UZDoom.png
the-phinet: If we're switching to a raster, src/CMakeLists.txt:1468 needs to be updated (will break install / appimage generation)

nashmuhandes:
Closing because I wanted to redo the PR from scratch

Issue #14, opened 2025-10-14
[Feature] Add a way to lift ACS VM instruction limit


TDRRmk2:
### Which game are you running with GZDoom?

None

### What Operating System are you using?

None

### Is your feature request related to a problem? Please describe.

This is a re-post of one of my old tracker issues from the Zandro tracker, to allow for discussion and decision of what the best way to handle this would be:
Mods, especially more modern ones, can very easily hit the runaway script limit with more complex code. It's intended to help stop infinitely looping scripts, but the fact it's an instruction limit means it doesn't just limit the amount of loops, but also the amount of work done in a tic in general, so more than detecting infinite loops it's just a hindrance. This is also exacerbated by more powerful and modern compilers like GDCC allowing for more sophisticated implementations of features potentially requiring more instructions, and new features like ACS lump reading being heavily limited by this.

It's a rather arbitrary limit to still have around, too, given how neither DECORATE nor ZScript have any similar limits (DECORATE does have one, but only in Zandronum and it's purely clientside).

### Describe the solution you'd like

Some ways to lift the limit have been suggested, such as:
-A "BUSY" script flag, suggested by Dusk. Adds a script flag that removes the instruction limit only for the flagged script. Would require modifying ACC and would introduce more differences with GZDoom, but allows the most granular control.

-Disabling by default, but keeping a compatflag to enable it if an old mod needs it (or for debugging), suggested by Sean. Doesn't require modifying ACC and there's much less difference with GZDoom.
(I have implemented this already, but it got no'd by Graf)
https://github.com/ZDoom/gzdoom/pull/2486

-Making it a branch limit instead, suggested by DavidPH. This would vastly increase the amount of work allowed and would address the issue much more directly. It would pretty much just limit the amount of loops and not the amount of work done in said loops. I kinda like this one but don't like that it basically just increases the limit, which would mean we'd have the same issue in the future anyway.

-Removing it entirely. Some old mods do have runaway script issues, particularly those made for Skulltag and old Zandronum. I would prefer one of the other ways instead.

Boondorl:
Tbh I might just bump this to INT_MAX since I doubt any remotely modern CPU is gonna struggle to hit that limit quickly. People should always be fixing their infinite loops so a huge stutter isn't an issue and will more or less force them to fix it.

(I meant to do this a while ago but it completely slipped my mind, sorry)

Boondorl:
[...]

PR #98, opened 2025-10-16
Add support for the "BUSY" script flag.

TDRRmk2:
This flag allows the script to skip the runaway limit. This addresses issue https://github.com/UZDoom/UZDoom/issues/14
The relevant ACC PR is here: https://github.com/UZDoom/acc/pull/1

I deliberately tried to keep changes as minimalistic as possible for easy backporting into Zandronum. If there's anything that could be tweaked please let me know.

### DIFF
--- a/src/playsim/p_acs.cpp
+++ b/src/playsim/p_acs.cpp
@@ -6961,7 +6961,7 @@ int DLevelScript::RunScript()
 	ACSLocalArrays noarrays;
 	ACSLocalArrays *localarrays = &noarrays;
 	ScriptFunction *activeFunction = NULL;
-	ScriptPtr *ptr;
+	ScriptPtr *ptr = nullptr;
 	FRemapTable *translation = 0;
 	int resultValue = 1;
 	int transi = -1;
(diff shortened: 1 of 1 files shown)


src/playsim/p_acs.cpp
```diff
@@ -6961,13 +6961,14 @@ int DLevelScript::RunScript()
 	ACSLocalArrays noarrays;
 	ACSLocalArrays *localarrays = &noarrays;
 	ScriptFunction *activeFunction = NULL;
+	ScriptPtr *ptr;
```
kcat: I don't know about the overall logic for the function, but this is left uninitialized when `InModuleScriptNumber` is negative, so the later `ptr` check on line 7052 isn't valid in that case.

src/playsim/p_acs.cpp
```diff
@@ -7048,7 +7049,7 @@ int DLevelScript::RunScript()
 
 	while (state == SCRIPT_Running)
 	{
-		if (++runaway > 2000000)
+		if ( (++runaway > 2000000) && ptr && !(ptr->Flags & SCRIPTF_Busy) )
```
marrub--: I feel like this shouldn't increment the runaway counter when the script is marked as BUSY.

PR #99, opened 2025-10-16
[Fix] Re-license to GPLv2

the-phinet:
Re-license file to GPLv2, in order to allow Raze to use it


Moth-Tolias:
uh, isn't this not legally possible without the permission of everyone who has previously contriibuted to this file?

RicardoLuis0:
yep, this file was authored by @the-phinet in its entirety, which is why it can be so simply relicensed

PR #107, opened 2025-10-17
Fix "NoMBF21" behavior

### LINKED ISSUE #36: [BUG] Remove `nombf21` behavior in favor of matching `comp_reservedlineflag`
The original implementation of the MBF21 spec didn't account for the old borked editors that placed erroneous "ON" bits for the upper byte in linedef flags. This was corrected in [v1.4](https://github.com/kraflab/mbf21/commit/a2eecae9b695d146572b8ad85318dc5f27dab2d0) of the spec. This is a particularly egregious issue, as MBF21 did not need to be entirely disabled at all, only the upper byte of the linedef flags needed clearing, and even then, only when running under MBF21 compatibility. This has silently cause many issues that went unreported by the community at large, as many MBF21 maps/wads now fail to load correctly unless `compat_nombf21` is set to false on ZMAPINFO or an "MBF21" string is provided on a COMPLVL lump.

elf-alchemist:
Fixes #36
Early attempt at fixing a poor Graf implementation and better matching the MBF21 spec. I've not been able to consistently test this PR properly, while E2M7's step-up crates in the northeast corner of the map _aren't_ blocking the player, I've not yet consistently reproduced a scenario that actually marks them as player-blocking. I'll test further, soon.


Boondorl:
Just a heads up, the language file itself shouldn't be changed. Historically this was done through Graf's spreadsheet on Google and then pulled in later, but we'll be looking into different options for the future. For now you can ignore it

src/gamedata/options.cpp
```diff
@@ -74,6 +75,7 @@ static dsda_option_t option_list[] = {
   { "comp_stairs", &dsda_options::comp_stairs, 0, 1 },
   { "comp_ledgeblock", &dsda_options::comp_ledgeblock, 0, 1 },
   { "comp_friendlyspawn", &dsda_options::comp_friendlyspawn, 0, 1 },
+  { "comp_reservedlineflag", &dsda_options::comp_friendlyspawn, 0, 1 },
```
BlueSpeedsterYT: why do I feel like this is just an oversight?

the-phinet:
﻿
﻿

the-phinet:
What remains for this to be ready for review?

elf-alchemist:
I've successfully reproduced the original MBF21 issue pertaining to the `ML_RESERVED` linedef flag, turns out the map translation definitions weren't even allowing it through into the map translation code. This PR can finally be considered functionality-wise done, ready for any review point to be brought up. Is there an issue with the commit history? I am used to not worrying about commit messages in PRs due to most squash-merging branches. This PR being done, marks UZD as another step closer to matching other ports, and thus help reduce another compatibility pain point for mappers.

RicardoLuis0:
> Is there an issue with the commit history? I am used to not worrying about commit messages in PRs due to most squash-merging branches.

we mostly merge with rebases not squashes, the commit history would need to be either linearized or squashed

RicardoLuis0 (approved):

PR #124, opened 2025-10-17
constexpr instead of macro defines for Console defaults

River-Salmon:
As a small start to a general modernization pass, this change uses constexpr rather than macro defines to set default values for the console. These defaults have also been moved into namespaces to avoid ambiguity.


src/common/console/c_console.cpp
```diff
@@ -62,9 +62,22 @@
...
+	static inline constexpr uint8_t bottom_margin = 12;
+
+	static inline constexpr uint8_t max_hist_size = 50;
+
+	static inline constexpr uint8_t scroll_up = 1;
+	static inline constexpr uint8_t scroll_down = 2;
+	static inline constexpr uint8_t scroll_no = 0;
+	
+	static inline constexpr uint8_t min_con_lines = 20;
+	static inline constexpr uint8_t ms_between_cursor_ticks = 500;
```
kcat: This is too large for `uint8_t`, which maxes out at 255. It might be better to use braced initialization:
```c++
static inline constexpr uint8_t ms_between_cursor_ticks{500};
// or
static inline constexpr auto ms_between_cursor_ticks = uint8_t{500};
```
which would catch/error on narrowing conversions like these instead of implicitly truncating.

src/common/console/c_console.cpp
```diff
@@ -62,9 +62,23 @@
...
 
-#define LEFTMARGIN 8
-#define RIGHTMARGIN 8
-#define BOTTOMARGIN 12
+
+namespace Console::Defaults
+{
+	static inline constexpr uint8_t left_margin = 8;
+	static inline constexpr uint8_t right_margin = 8;
+	static inline constexpr uint8_t bottom_margin = 12;
```
the-phinet: This (bottom_margin) should be 4

the-phinet (changes requested):

src/common/console/c_console.cpp
River-Salmon: It was 12 in the change that it was forked from

the-phinet: Sorry, I should have explained better. I think that's an oversight. The only place `Defaults::bottom_margin` is used is replacing the magic number 4. (`BOTTOMARGIN` was never actually used)

```diff
@@ -623,17 +635,17 @@ void C_DrawConsole ()
...
- round((float)ConBottom / textScale) - CurrentConsoleFont->GetHeight() - 4,
+ round((float)ConBottom / textScale) - CurrentConsoleFont->GetHeight() - Defaults::bottom_margin,
...
- round((float)ConBottom / textScale) - CurrentConsoleFont->GetHeight() - 4,
+ round((float)ConBottom / textScale) - CurrentConsoleFont->GetHeight() - Defaults::bottom_margin,
...
```

River-Salmon: updated

the-phinet (approved):

TehRealSalt (approved):

PR #129, opened 2025-10-18
only allow numbers for `discordappid` and `steamappid`

madame-rachelle:


RicardoLuis0:
might need to use a per-character check instead to only allow numbers, i'm not sure if ids are always small enough to fit in an int

madame-rachelle:
Alright, I will work on that, probably use a different approach to address this.

RicardoLuis0 (approved):

Issue #128, opened 2025-10-18
[BUG?] DropInventory keeping Item if amount is zero without bKeepDepleted


XLightningStormL:
### UZDoom version

Gzdoom g4.15pre-589-g24e7bee0f

### Which game are you running with UZDoom?

Doom 2

### What Operating System are you using?

Windows 10

### Have you checked that no other similar issue already exists?

- [x] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

When using Drop console command, DropInventory, or A_DropInventory; items without the bKeepDepleted flag will remain in the inventory even if the amount taken is equal to that of what is in the inventory.

I have a private solution, but I want to make sure this is an actual bug, and not an intended feature first.

### Steps to reproduce the behaviour.

Explain how to reproduce
1.  Open Console, type: give `shell 20` and hit enter
2. Open console, type `printinv`, and hit enter; Shell count should be 20
3. Open console, type `drop shell 20`, and hit enter
4. Open console, type `printinv`; shell count will be 0.

### Your configuration

```ini

```

RicardoLuis0:
ah, seems the amount check only checks for `Amount == 1`, that's definitely wrong https://github.com/UZDoom/UZDoom/blob/main/wadsrc/static/zscript/actors/inventory/inventory.zs#L785-L791
(seems to be a holdout from before CreateTossable took a configurable amount)

XLightningStormL:
> ah, seems the amount check only checks for `Amount == 1`, that's definitely wrong https://github.com/UZDoom/UZDoom/blob/main/wadsrc/static/zscript/actors/inventory/inventory.zs#L785-L791 (seems to be a holdout from before CreateTossable took a configurable amount)

That complicates my original bandaid solution which was to just throw
```
if(item.amount < 1&&item.bKeepDepleted==false)
{
	TakeInventory(item.GetClassName(), 1);
}
```
into DropInventory. Nice to know it's a more "root cause" issue, though it's out of my skill department.

RicardoLuis0:
yeah, the fix would be to, in Inventory::CreateTossable put the `amt = clamp(...)` before the bKeepDepleted check, and do an `Amount == amt` check instead of `Amount == 1` check

RicardoLuis0:
fixed with #131

PR #131, opened 2025-10-18
Fix "CreateTossable" keeping certain items at zero amount

XLightningStormL:
Fixed a bug in CreateTossable where certain items would be kept even if their amounts where at zero despite not having the bKeepDepletedFlag.

In my [original issue report](https://github.com/UZDoom/UZDoom/issues/128) I erroneously used an ammo item as an example; note that by default ammo is kept at zero for various reasons, thus custom mod items or those from Heretic/Hexen would be more clear testing examples.

With help from Jay.

### DIFF
--- a/wadsrc/static/zscript/actors/inventory/inventory.zs
+++ b/wadsrc/static/zscript/actors/inventory/inventory.zs
@@ -782,7 +782,8 @@ class Inventory : Actor
 		{
 			return NULL;
 		}
-		if (Amount == 1 && !bKeepDepleted)
+		amt = clamp(amt, 1, Amount);
+		if (Amount == amt && !bKeepDepleted)
 		{
 			BecomePickup ();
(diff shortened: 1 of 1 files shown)


RicardoLuis0 (approved):

advnau:
https://zdoom.org/wiki/Classes:Inventory
States 
image

so if we pass -1 as amt to the method, we are setting the amount to 1 which i suspect would be an issue.

RicardoLuis0:
> https://zdoom.org/wiki/Classes:Inventory States image
> 
> so if we pass -1 as amt to the method, we are setting the amount to 1 which i suspect would be an issue.

A_DropItem has nothing to do with CreateTossable

RicardoLuis0:
CreateTossable before amt was added, always dropped 1, and -1 being the default means it keeps that behavior

PR #138, opened 2025-10-18
Fix crushers restarting looping sequences

### LINKED ISSUE #7: [BUG] Crushers restart their sounds instead of looping
Whenever a crusher changes directions, they'll restart the whole sound sequence instead of letting it keep looping like how Doom does it. 

I think this change occurred around the time the sound rework was done, which was about November/December 2019.

Plug this into SNDSEQ. 
```cs
:CeilingNormal
	play		plats/pt1_stop
	playrepeat	plats/pt1_mid
	stopsound	plats/pt1_stop
end
```

The expected result, for me at least, is to have the crushers keep looping the repeat sound without playing the starting or stopping sound until it's actually stopped. But instead, it goes through the whole thing and is rather jarring. Even Star Wars Dark Forces does what Doom does, it plays the looping sound without restarting it until it's stopped.

Another way to check is to do this:

```cs
:CeilingNormal
	playrepeat	plats/pt1_strt
end
```

You'll definitely tell it's repeating since Doom's normal movement sound is difficult to distinguish.

mohin-io:
## Summary
- track whether a sector sound sequence is currently executing a playrepeat command
- use that loop state to skip re-triggering the ceiling sequence when crushers reverse
- persist the loop flag for savegames and script adjustments to avoid regressions

## Testing
- not run (engine build not available in environment)

Closes #7

### DIFF
--- a/src/sound/s_sndseq.cpp
+++ b/src/sound/s_sndseq.cpp
@@ -305,6 +305,7 @@ DSeqNode::DSeqNode ()
 : m_SequenceChoices(0)
 {
 	m_Next = m_Prev = m_ChildSeqNode = m_ParentSeqNode = nullptr;
+	m_IsLooping = false;
 }
 
 void DSeqNode::Serialize(FSerializer &arc)
@@ -318,6 +319,7 @@ void DSeqNode::Serialize(FSerializer &arc)
 	float atten = ATTN_NORM;
 	int seqnum;
 	unsigned int numchoices;
+	bool looping = false;
 
 	// copy these to local variables so that the actual serialization code does not need to be duplicated for saving and loading.
 	if (arc.isWriting())
@@ -329,6 +331,7 @@ void DSeqNode::Serialize(FSerializer &arc)
 		id = m_CurrentSoundID;
 		seqName = Sequences[m_Sequence]->SeqName;
 		numchoices = m_SequenceChoices.Size();
+		looping = m_IsLooping;
 	}
 	Super::Serialize(arc);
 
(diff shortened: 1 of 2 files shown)


MajorCooke:
Will try.

MajorCooke:
Ignore my previous comment, I just realized my build package had screwed up. Gonna test one more time and find out.

MajorCooke:
It works. This is good to go.

Boondorl (approved):

MajorCooke (approved):

Boondorl:
I'm gonna approve this for now, but I'd actually like to take a peak into the sound engine first to see if this is something going on in the backend since Cooke mentioned it happened around the rework. It's possible this is a bug, but it could also be expected behavior and sound sequences were missed in the update. If it ends up taking too long or it's too complicated to resolve, this can just be merged

MajorCooke:
Either way, it definitely wasn't behaving to how it used to be with regular doom when the sound rework happened. I have a [sound package](https://drive.google.com/file/d/1GQl9ZImn--mDx4rkr6VRnZwhMmhDU3Wc/view?usp=drive_link) you can use to test it out. 

You can also simply rename some of the sounds to their replacement sounds and it'd still produce the exact same result with vanilla sequences too.

inkoalawetrust:
Status on this? It has been sitting for 3 weeks now.

RicardoLuis0:
it's still waiting on testing (and for a further look into the sound engine)

MajorCooke:
> it's still waiting on testing (and for a further look into the sound engine)

I already tested and it works. Also Boon say it can be merged if nothing else comes of it.

Boondorl:
I don't think I'll be able to get to this any time soon. I think we merge this and create a task to investigate it later, especially in light of #510 potentially having a lead on caching not working properly.

@mohin-io This will need to be rebased so we can run the appropriate CI and merge it in.

PR #140, opened 2025-10-18
Indicate shader compilation phase during startup when possible

Cacodemon345:
This PR indicates shader compilation phase during startup when possible, especially if start screens are displayed.

https://github.com/user-attachments/assets/02a76124-6a84-4e97-ad5b-e2b8475ec1ec

### DIFF
--- a/src/common/startscreen/startscreen.cpp
+++ b/src/common/startscreen/startscreen.cpp
@@ -709,6 +709,7 @@ void FStartScreen::Render(bool force)
 		if (ShaderComp && NewConsoleFont)
 		{
 			FString compStr = GStrings.localize("$SHADERCOMPILING");
+			compStr.Truncate(compStr.Len() - (3 - ShaderComp));
 			auto strWidth = NewConsoleFont->StringWidth(compStr);
 			if (ShaderComp == 3)
 				compStr.AppendFormat("..");
--- a/src/common/startscreen/startscreen.cpp
+++ b/src/common/startscreen/startscreen.cpp
@@ -708,18 +708,15 @@ void FStartScreen::Render(bool force)
 
 		if (ShaderComp && NewConsoleFont)
 		{
-			FString compStr("Compiling shaders.");
-			if (GStrings.CheckString("SHADERCOMPILING"))
-			{
-				compStr = GStrings.localize("$SHADERCOMPILING");
-			}
+			FString compStr = GStrings.localize("$SHADERCOMPILING");
+			auto strWidth = NewConsoleFont->StringWidth(compStr);
 			if (ShaderComp == 3)
 				compStr.AppendFormat("..");
 			else if (ShaderComp == 2)
 				compStr.AppendFormat(".");
 
 			Dim(twod, PalEntry(0, 0, 0, 255), 0.5, 0, 0, screen->GetWidth(), screen->GetHeight());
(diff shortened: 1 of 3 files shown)


RicardoLuis0 (changes requested):
this should probably indicate every step of the loading process, not just shader compilation, because for asset-heavy mods it may not at all be the longest part

Cacodemon345:
If you mean texture loading (which would cover majority of the load times of those mods), I don't think that really needs to be covered here because it's easily visible in the progress bar. Shader loading, however, isn't really easily visible there, especially for Vulkan. There's also the problem of the texture manager still having to be properly initialized here to not run into problems when rendering texts. Hence the reason to only display this in the shader compilation stage so that the user knows it has finished texture loading.

RicardoLuis0:
> If you mean texture loading (which would cover majority of the load times of those mods), I don't think that really needs to be covered here because it's easily visible in the progress bar. Shader loading, however, isn't really easily visible there, especially for Vulkan. There's also the problem of the texture manager still having to be properly initialized here to not run into problems when rendering texts. Hence the reason to only display this in the shader compilation stage so that the user knows it has finished texture loading.

all initialization phases, loading files/compiling zscript/etc

Cacodemon345:
In that case it will require redoing this PR completely since no start screens were ever made to accommodate that kind of thing. And honestly ZScript compilation shouldn't take that much time as compared to asset loading.

src/common/startscreen/startscreen.cpp
```diff
@@ -701,6 +701,22 @@ void FStartScreen::Render(bool force)
 			DrawTexture(twod, StartupTexture, 0, 0, DTA_VirtualWidthF, displaywidth, DTA_VirtualHeightF, displayheight, TAG_END);
 		}
 
+		if (ShaderComp && NewConsoleFont)
+		{
+			FString compStr("Compiling shaders.");
```
rheit: Why do this? You also created a localizable string. Just define its default and use that.

src/common/startscreen/startscreen.cpp
```diff
@@ -701,6 +701,22 @@ void FStartScreen::Render(bool force)
...
+			if (GStrings.CheckString("SHADERCOMPILING"))
+			{
+				compStr = GStrings.localize("$SHADERCOMPILING");
```
rheit: You can just use `GStrings.GetString("SHADERCOMPILING")`

Cacodemon345:
> It would look nicer if the text didn't move around as dots are added and removed. Or maybe a spinner "graphic", looping through a sequence of `-` `\` `|` `/` characters instead of building up and taking away dots.

It's deliberately using spinning dots to minimize the impact on shader compilation times since it's not done in parallel.

src/common/startscreen/startscreen.cpp
```diff
@@ -703,18 +703,15 @@ void FStartScreen::Render(bool force)
...
 		if (ShaderComp && NewConsoleFont)
 		{
-			FString compStr("Compiling shaders.");
-			if (GStrings.CheckString("SHADERCOMPILING"))
-			{
-				compStr = GStrings.localize("$SHADERCOMPILING");
-			}
+			FString compStr = GStrings.localize("$SHADERCOMPILING");
+			auto strWidth = NewConsoleFont->StringWidth(compStr);
 			if (ShaderComp == 3)
```
rheit: If you're going to bake a part of the ticker into the localized string, you should bake it all in, not just the first character. Not all languages use '.' like this. Then you can just draw a portion of the string starting at the beginning and going for between (length - 2) and (length) characters without doing any string composition here.

Or do the other suggestion and use a spinner character: `char spinChar = "\|/-"[ShaderComp];` (assuming `ShaderComp` ranges between 0 and 3).

the-phinet:
﻿
﻿

Issue #100, opened 2025-10-16
[BUG] Inconsistent treatement of alpha values > 1.0 breaks certain older mods


electricbrass:
### UZDoom version

GZDoom 1.14.2 (Also checked a UZDoom commit yesterday but I deleted it and don't remember exactly what it was)

### Which game are you running with UZDoom?

Doom 2

### What Operating System are you using?

Windows 11

### Please describe your specific OS version

Windows 11 Pro

### Relevant hardware info

Core i5-9300H + GTX 1650

### Have you checked that no other similar issue already exists?

- [x] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

This is a continuation of https://github.com/ZDoom/gzdoom/issues/3163. The appearance of alpha values greater than 1.0 differ between hardware and software rendering, with Zandronum and pre-2.2.0 GZDoom matching the post-2.2.0 GZ/UZDoom software renderer in both modes. This causes visual issues with certain older wads, such as Pirate Doom. It appears that in software mode (and hardware mode in older versions/Zandronum), values above 1.0 were treated as a fraction of 255. Alpha values of 0.5 and 128 appear identical as far as i can tell, except in post-2.2.0 hardware mode, where it ends up solid white.

It looks like between 2.1.1 and 2.3.0, `AActor::alpha` was changed from a `fixed_t` to a `double` in 4e60ea0252d432be5770ddbf61d67937876285f6. This may not be the source of the issue but it seems like a place to look.

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Make sure hardware rendering is on
2. Go to map48 of Pirate Doom
3. Pick up a cannon
4. Fire it

### Your configuration

```ini
Can be reproduced with a default config.
```

the-phinet:
Closing this, since the only known mod with this issue has patched the behaviour on their end. If you think this is a mistake, feel free to open a new issue.

PR #143, opened 2025-10-18
emulate legacy behavior of alpha > 1 treated as fraction of 255

River-Salmon:
In response to https://github.com/UZDoom/UZDoom/issues/100, this does the "treat alpha > 1.0 as a fraction of 255" behavior for the hardware sprite rendering.

### DIFF
--- a/src/d_main.cpp
+++ b/src/d_main.cpp
@@ -736,6 +736,7 @@ CVAR (Flag, compat_nombf21,				compatflags2, COMPATF2_NOMBF21);
 CVAR (Flag, compat_voodoozombies,		compatflags2, COMPATF2_VOODOO_ZOMBIES);
 CVAR (Flag, compat_fdteleport,			compatflags2, COMPATF2_FDTELEPORT);
 CVAR (Flag, compat_novdolllockmsg,		compatflags2, COMPATF2_NOVDOLLLOCKMSG);
+CVAR (Flag, compat_alphawraparound,		compatflags2, COMPATF2_ALPHAWRAPAROUND);
 
 CVAR(Bool, vid_activeinbackground, false, CVAR_ARCHIVE | CVAR_GLOBALCONFIG)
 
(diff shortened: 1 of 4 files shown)


RicardoLuis0 (dismissed):
haven't tested with pirate doom itself yet, but this looks good 👍

River-Salmon:
Thanks. I tested it with Pirate Doom and got the expected behavior

nashmuhandes:
Dynamic and lightmap light Actors in both UZD and VKDoom also can have their Alpha's go beyond 1.0 for overbrights. This definitely needs to be thrown under a compat setting

RicardoLuis0:
> Dynamic and lightmap light Actors in both UZD and VKDoom also can have their Alpha's go beyond 1.0 for overbrights. This definitely needs to be thrown under a compat setting

that's on map loading though, it isn't passed along to the HWSprite::Process, so it wouldn't at all be affected by this

kcat:
`ALPHAWRAPAROUND` doesn't seem the right terminology. "Wrap around" would suggest modulo-like behavior (going above one wraps it back around to 0, where it can increase to 1 again and wrap back to 0 again, etc). This is more like auto-normalize behavior, where it expects a value between 0...1, and if it goes above 1, it's scaled down by 255 to bring it back to the 0...1 range.

TehRealSalt (approved):

River-Salmon:
It's worth noting this PR does not include an actual addition of Pirate Doom to the compat flags file, just the definition of the compat flag itself

RicardoLuis0 (changes requested):
> `ALPHAWRAPAROUND` doesn't seem the right terminology. "Wrap around" would suggest modulo-like behavior (going above one wraps it back around to 0, where it can increase to 1 again and wrap back to 0 again, etc). This is more like auto-normalize behavior, where it expects a value between 0...1, and if it goes above 1, it's scaled down by 255 to bring it back to the 0...1 range.

i second this, i'd prefer some different naming for the compat, ex. `int8alpha` or something

River-Salmon:
Are there other examples of this than Pirate Doom? Pirate Doom has a patch to fix this on their side

the-phinet:
﻿
﻿

the-phinet:
Am I good to close this? It fixes a thing that has already been fixed in the one mod where it was an issue. Adding this would add another option for little good reason

River-Salmon:
I think it's good to close

RicardoLuis0:
> Am I good to close this? It fixes a thing that has already been fixed in the one mod where it was an issue. Adding this would add another option for little good reason

ah, if it's fixed in pirate doom then yeah

PR #144, opened 2025-10-18
Compile with C++20 standards (formerly C++23)

ucosty:
This update configures cmake to instruct the compiler to use the C++20
standard. Most of the code was already compliant, but some changes to
the range_map library were required to get it to compile on C++20 and
later.

Specifically, I removed the use of `allocator::rebind` which has been
deprecated, and instead use `std::allocator_traits`. I also implemented
a small number of code changes to make the compiler a little happier,
though there is still plenty of scope to clean the code up more.

I made this change because I wanted to debug an issue I had on startup, which
ultimately turned out to be something trivial, but in trying to use std::print I
realised that the code was being compiled as C++17. I fixed up the code to get it
to compile, and then realised I just forgot to copy one of the pk3 files into the uzdoom
directory. I figured the work shouldn't go to waste, so I may as well raise a PR.

I tested this out by playing some `myhouse.pk3` hoping it would exercise the engine
features enough to highlight any problems. As far as I can tell it played just fine.


kcat:
Quite the jump from C++17 to C++23. This would require rather new compilers to build, GCC 14/15 or Clang 19/20 etc (depending on what exactly gets used). Maybe C++20 would be a better more immediate target? That gets you `std::format`, which you can build a `std::print`-like function with, as well as many other features (ranges, concepts, coroutines, ~~modules~~, and more) without requiring as bleeding-edge of a compiler and toolchain.

But I don't know what this project's guidelines are for standard targets, or features (the aforementioned modules are starting to become usable, but don't expect them to work prior to GCC 15/Clang 17(?) or CMake 3.28 and they require Ninja or VS builds; Makefiles are still unsupported without handling dependencies and build commands manually). I personally don't have much of a complaint as I'm using recent GCC, Clang, and CMake versions, but it's not just me that this would affect.

RicardoLuis0:
yeah, i'd rather this update it to C++20, C++23 support is still quite partial in MSVC for example

RicardoLuis0 (approved):

ucosty:
I will take a look at this, at the least I'll figure out which minimum versions of GCC and Clang we need to build against.

RicardoLuis0:
according to CI just adding the missing typename keyword to the typedef should hopefully be enough
image

RicardoLuis0:
@ucosty any updates?

johncurley:
If the project is C++20, then it will be impossible to build on older MacOS builds for Intel. I already can't compile OpenAL-Soft on my machine.

kcat:
> If the project is C++20, then it will be impossible to build on older MacOS builds for Intel.

Which older versions? The CI succeeds on what seems to be macOS 15.4, and from what I'm reading there are still some Intel Macs supported on recent macOS releases.

RicardoLuis0:
> If the project is C++20, then it will be impossible to build on older MacOS builds for Intel. I already can't compile OpenAL-Soft on my machine.

if openal-soft won't build in those versions of macos then i'm afraid they won't be a valid target for uzdoom regardless - openal-soft is a dependency if you want sound at all (if you want to keep using uzdoom on those machines, i believe intel macs support linux?)

johncurley:
The brew .dynlib still links against UZDoom, it’s just not bundled with a static library. On 1 Nov 2025, at 7:35 PM, Ricardo Luís Vaz Silva ***@***.***> wrote:﻿RicardoLuis0 left a comment (UZDoom/UZDoom#144)

If the project is C++20, then it will be impossible to build on older MacOS builds for Intel. I already can't compile OpenAL-Soft on my machine.

if openal-soft won't build in those versions of macos then i'm afraid they won't be a valid target for uzdoom regardless - openal-soft is a dependency if you want sound at all

johncurley:
I’m looking at clang 14 support so by passing c++20 support as a compiler flag, enables partial support. It might be enough to get away with on macOS 12 for development. But really I need a Silicon machine particularly for modern support.

the-phinet:
Closing in favour of #502

PR #145, opened 2025-10-18
[FEATURE] Build script for linux

blockplace1232:
I wrote this script just for convenience sake and thought it would be a good idea to include in the main branch.

This has 0 conflicts so should be relatively straight forward merge.

### DIFF
--- a/buildlinux.sh
+++ b/buildlinux.sh
@@ -5,7 +5,7 @@ mkdir -pv ~/uzdoom_build
 cd ~/uzdoom_build &&
 git clone https://github.com/UZDoom/uzdoom.git &&
 mkdir -pv uzdoom/build
-c="$(lscpu -p | grep -v '#' | sort -u -t , -k 2,4 | wc -l)" ; [ "$c" -eq 0 ] && c=1
+c="$(nproc)" ; [ "$c" -eq 0 ] && c=1
 cd ~/uzdoom_build/uzdoom/build &&
 rm -f output_sdl/liboutput_sdl.so &&
 cmake .. -DCMAKE_BUILD_TYPE=Release $f &&


buildlinux.sh
```diff
@@ -0,0 +1,17 @@
...
+mkdir -pv ~/uzdoom_build
+cd ~/uzdoom_build &&
+git clone https://github.com/UZDoom/uzdoom.git &&
+mkdir -pv uzdoom/build
+a='' && [ "$(uname -m)" = x86_64 ] && a=64
+c="$(lscpu -p | grep -v '#' | sort -u -t , -k 2,4 | wc -l)" ; [ "$c" -eq 0 ] && c=1
+cd ~/uzdoom_build/uzdoom/build &&
+rm -f output_sdl/liboutput_sdl.so &&
+if [ -d ../fmodapi44464linux ]; then
+f="-DFMOD_LIBRARY=../fmodapi44464linux/api/lib/libfmodex${a}-4.44.64.so \
```
madame-rachelle: We no longer use the FMOD library. This will not do anything and probably will be better removed.

madame-rachelle (changes requested):

buildlinux.sh
```diff
@@ -0,0 +1,17 @@
+# based upon the GZDoom build guide avaliable at https://zdoom.org/wiki/Compile_GZDoom_on_Linux
+# this script is rather basic but gets the job done, executable files found in the "build" folder
+# also this script only works once you've installed all the dependencies since the install command varies from distro to distro
+mkdir -pv ~/uzdoom_build
+cd ~/uzdoom_build &&
+git clone https://github.com/UZDoom/uzdoom.git &&
+mkdir -pv uzdoom/build
+a='' && [ "$(uname -m)" = x86_64 ] && a=64
```
RicardoLuis0: also 32-bit builds are straight up _not_ supported

blockplace1232: I'm not at my pc right now, I'll remove 32-bit compilation once I get back as I need to make sure I haven't screwed something up

madame-rachelle (approved):

buildlinux.sh
```diff
@@ -0,0 +1,12 @@
+# based upon the GZDoom build guide avaliable at https://zdoom.org/wiki/Compile_GZDoom_on_Linux
+# this script is rather basic but gets the job done, executable files found in the "build" folder
+# also this script only works once you've installed all the dependencies since the install command varies from distro to distro
+mkdir -pv ~/uzdoom_build
+cd ~/uzdoom_build &&
+git clone https://github.com/UZDoom/uzdoom.git &&
+mkdir -pv uzdoom/build
+c="$(lscpu -p | grep -v '#' | sort -u -t , -k 2,4 | wc -l)" ; [ "$c" -eq 0 ] && c=1
```
RicardoLuis0: also this isn't being used, shouldn't it be passed as `--parallel $c` in cmake?

madame-rachelle: It is being used in the make line, here:
https://github.com/UZDoom/UZDoom/pull/145/files#diff-98187e1e979109bea615249da80d20c53cc32586164bb4a005a084e0810a3feeR12

RicardoLuis0: also why use `$(lscpu -p | grep -v '#' | sort -u -t , -k 2,4 | wc -l)` instead of simply `$(nproc)`?

blockplace1232:
i accidentally closed this woops

the-phinet:
Hey, thanks for working on this. After some discussion with the team, we've decided to close this. We really appreciate the effort you've put into this, but a script like this is not appropriate for this project.

Here are a breakdown of our reasoning:

- A build script exists in the form of CMakeLists.txt
- Gathering dependencies and installing a compiler are the majority of effort needed to build the project. A useful build script would need to at least gather dependencies, but that process is different for each distro.
- There is no place in the repository that a script could be placed that is both visible to beginners, and not in the way of more experienced developers.

Ultimately we feel the solution is to improve the build documentation. You can find the current draft [here](https://github.com/UZDoom/UZDoom/wiki/Compilation#linux)

PR #146, opened 2025-10-18
Add isnot operator

kevansevans:
Forgive me if this PR is extremely sloppy, I'm not exactly familiar with with how any of this works. BUT, this adds the ``isnot`` operator as sugar syntax, to work as an inverse as the ``is`` operator,

### DIFF
--- a/src/common/scripting/frontend/zcc-parse.lemon
+++ b/src/common/scripting/frontend/zcc-parse.lemon
@@ -1775,7 +1775,7 @@ expr(X) ::= expr(A) IS expr(B).			/* a is b */
 	BINARY_EXPR(A,B,PEX_Is);
 	X = expr2;
 }
-expr(X) ::= expr(A) ISNOT expr(B).			/* a is b */
+expr(X) ::= expr(A) BANG IS expr(B).	/* a !is b */
 {
 	BINARY_EXPR(A,B,PEX_Isnot);
 	X = expr2;
--- a/src/common/engine/sc_man_scanner.re
+++ b/src/common/engine/sc_man_scanner.re
@@ -210,7 +210,6 @@ std2:
 		'sealed'					{ RET(ParseVersion >= MakeVersion(4, 12, 0)? TK_Sealed : TK_Identifier); }
 
 		'is'						{ RET(ParseVersion >= MakeVersion(1, 0, 0)? TK_Is : TK_Identifier); }
-		'is!'						{ RET(ParseVersion >= MakeVersion(4, 15, 1)? TK_Isnot : TK_Identifier); }
 		'replaces'					{ RET(ParseVersion >= MakeVersion(1, 0, 0)? TK_Replaces : TK_Identifier); }
 		'states'					{ RET(TK_States); }
 		'meta'						{ RET(ParseVersion >= MakeVersion(1, 0, 0)? TK_Meta : TK_Identifier); }
--- a/src/common/engine/sc_man_tokens.h
+++ b/src/common/engine/sc_man_tokens.h
@@ -127,6 +127,7 @@ xx(TK_Include,				"'include'")
 xx(TK_Sealed,				"'sealed'")
 
 xx(TK_Is,					"'is'")
+xx(TK_Isnot,				"'!is'")
 xx(TK_Replaces,				"'replaces'")
 xx(TK_Vector2,				"'vector2'")
 xx(TK_Vector3,				"'vector3'")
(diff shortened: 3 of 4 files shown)


MajorCooke:
This brings to mind the concern for backwards compatibility and if someone has the variables `isnot`... Or is that even a concern?

kevansevans:
Would be a horribly unfortunate coincidence if there are any variables out there named that

MajorCooke:
I legit bring this up as a concern in the event the engine doesn't know how to handle such a situation. I have no issues with it otherwise.

kevansevans:
Would ``!is`` be a better keyword then?

RicardoLuis0:
why add something new for this when it can just be done as `!(<X> is <Y>)`? plus whitespace in tokens is a very big no from me (just make the `isnot` token require version 4,15,1)

RicardoLuis0:
also this does not at all need custom compilation, it can be simply done as a sugar for `!(<X> is <Y>)` with
```
case PEX_Isnot:
    return new FxUnaryNotBoolean(new FxTypeCheck(left, right));
```

src/common/scripting/frontend/zcc-parse.lemon
```diff
@@ -2428,10 +2428,4 @@ variable_list_with_init(X) ::= variable_list_with_init(A) COMMA var_init(B).
 {
 	AppendTreeNodeSibling(A, B);
 	X = A; /*X-overwrites-A*/
-}
-
-expr(X) ::= expr(A) ISNOT expr(B).			/* a is not b */
-{
-	BINARY_EXPR(A,B,PEX_Isnot);
-	X = expr2;
-}
```
RicardoLuis0: seems this got removed accidentally 😅

kevansevans: I'm trying I swear! I'll get to it tomorrow, it's past my bed time.

Gutawer:
If we're going to do this I'd like to point out that C# and Python use `is not`, VB and PowerShell use `isnot` and Kotlin uses `!is`

Pythons usage of it is self-consistent because its logical negation operator is `not`. C#'s is self-consistent because it exists in the context of a rich pattern-matching system where `not pattern` makes grammatical sense *without* the `is` (and crucially, `!` wouldn't have worked for them because it already has the boolean meaning). Kotlin's syntax would therefore be the most self-consistent within ZScript so it'd have my vote but parser troubles were ran into when implementing it.

Imo - I think `!(x is y)` is totally fine. If you see it like a function call, it's just like having to do `!(x.is(y))`. The only thing I feel at all strongly about though is that it should look like a grammatical modification of the `is` operator, so I don't like `isnot`

mc776:
> The only thing I feel at all strongly about though is that it should look like a grammatical modification of the is operator, so I don't like `isnot`

Also: "I snot"

RicardoLuis0:
does the parser not accept it if you do `::= expr(A) IS BANG expr(B).`? adding symbols as part of keywords isn't too great

RicardoLuis0:
> All of this reminds me of something... last time I asked a few years ago I was assured there was no functional difference between checking for "X is Y" and checking for "Y(X)" and I can only infer the same goes for "!(X is Y)" and "!Y(X)"*.
> 
> Is that (still) true?
> 
> *(which if I am correct about that interchangeability is the most compact and simple way to say that - just the class/pointer names plus 3 characters).

functionality-wise, true, yeah, but the compiler isn't smart enough to eliminate everything, so it's still marginally cheaper to use `is`

Gutawer:
`is!` is absolutely the worst option imo as `!` is already a unary operator - `x is! y` should parse as `x is (!y)` and then produce a type error, exactly how itd work if you did `x dot! y`, i really don't like the idea of `is` working differently to every other operator in the language

RicardoLuis0:
> `is!` is absolutely the worst option imo as `!` is already a unary operator - `x is! y` should parse as `x is (!y)` and then produce a type error, exactly how itd work if you did `x dot! y`, i really don't like the idea of `is` working differently to every other operator in the language
> 
> (or in some future version of zscript maybe we'll have operator overloading and it wouldn't even _necessarily_ be a type error)
> 
> also `!` honestly just "feels" like it should go before the thing its affecting - for example we write `!isSomething()` - it's ultimately an ascii-compatible version of ¬ (just like `&&` is one for ∧ and `||` is one for ∨) and shouldn't be read like the english word "not"

ugh LALR parser generators always confuse me with what would be accepted and what would not, but just the same, doing `::= expr(A) BANG IS expr(B).` should hopefully work

RicardoLuis0:
> The underlying reason for this reminds me of every time I need to write something like "if(abc!=x||abc!=y||abc!=z)" and I wish I could just write "if(abc!=x||y||z)" instead...

`if(abc!=x||y||z)` can't really be done unambiguously, but `if(abc!={x||y||z})` or `if(abc!=[x||y||z])` should be possible as syntax sugar

kevansevans:
That's odd, compiles and launches just fine for me. What exactly is the problem?

PR #147, opened 2025-10-19
fix models using stale animation data if no animation is playing

RicardoLuis0:


the-phinet (approved):

the-phinet:
Seems to work as intended 👍

PR #155, opened 2025-10-19
Add support for animated images loading

Cacodemon345:
This PR adds support for creating animated textures from animated images.


Cacodemon345:
I got word that video playback is already decoupled from the server in UZDoom.

PR closed while I investigate that feasibility of video textures.

PR #162, opened 2025-10-19
[Fix] audio engine not working because it expects "gzdoom.sf2" when it gets "uzdoom.sf2"

blockplace1232:
For some reason this file wasn't renamed and caused an issue where UZDoom couldn't find the soundfont when it was just under a different name.


blockplace1232:
(also i had to delete the file and re-add it to rename it because if i renamed it the file gets blanked)

RicardoLuis0:
it needs changes to CI as well

PR #188, opened 2025-10-21
Allow MoviePlayer instances to use different texture targets

Cacodemon345:
This PR allows MoviePlayer instances to use different texture targets instead of the default one.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1061,6 +1061,7 @@ set (PCH_SOURCES
 	common/cutscenes/playmve.cpp
 	common/cutscenes/movieplayer.cpp
 	common/cutscenes/screenjob.cpp
+	common/cutscenes/vidtexmanager.cpp
 	common/utility/engineerrors.cpp
 
 	common/utility/gitinfo.cpp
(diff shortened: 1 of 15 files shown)


Talon1024:
Is this for things like FMV talking heads like those seen in Wing Commander 3/4/Prophecy or Command & Conquer: Red Alert 2?

Cacodemon345:
[animtest.zip](https://github.com/user-attachments/files/23051960/animtest.zip)

Example. Load as a PK3. Uses the `Logo.ivf` example for showcase.

It will play back the video on the wall where the player faces initially.

Cacodemon345:
https://github.com/user-attachments/assets/045203c4-7c75-4a64-acec-21c5e1d8c19f

Cacodemon345:
[animtestacs.zip](https://github.com/user-attachments/files/23096674/animtestacs.zip)

Demonstration of the high level video manager API, called from ACS.

https://github.com/Cacodemon345/acc/tree/vidfunc

This is where the new ACS stuff exists.

jewalky:
Just to formalize by feedback from Discord (since channel moves, while PR doesn't)

1. There are issues with serialization of MoviePlayer to savegames when it's used in playsim context.
2. The video playback for textures does not pause if the game is paused.
3. The sound (if any) is played in UI, while the texture is in world space. (Also does not pause if the game is paused.)

And one new thing that I think I didn't try before:

4. What does the "lastframetime" parameter do on `VidMan`? I tried to pass 0 there, it works. I tried to pass 5000 there, it works. (And plays back the entire video). What time units are needed there?

PR #190, opened 2025-10-21
Restored Map Light Mode Option

### LINKED ISSUE #4: [Feature] Restore sector light options
Specifically what [this mod](https://forum.zdoom.org/viewtopic.php?t=78307) implements. I don't mind if it's still temporary and is reset upon restarting the engine, I just think it was a bad idea to remove it entirely.

MajorCooke:
This puts the Map Light Mode option back in Display -> Hardware Options.

I set this up to be cherry-pickable in the event persistence isn't desired. However I'm tired of relying on autoexec.cfg.

To everyone else who may want to reformat the map light mode, I understand. But if it's to be overhauled, I say we should have this anyway and it can be removed again at any other point in time, should an overhaul be worked on.

# Reminder, mapinfo still takes priority.

Fixes #4 .

### DIFF
--- a/src/g_level.cpp
+++ b/src/g_level.cpp
@@ -148,7 +148,7 @@ CUSTOM_CVAR(Bool, gl_notexturefill, false, CVAR_NOINITCALL)
 	}
 }
 
-CUSTOM_CVAR(Int, gl_maplightmode, -1, CVAR_NOINITCALL | CVAR_CHEAT) // this is just for testing. -1 means 'inactive'
+CUSTOM_CVAR(Int, gl_maplightmode, -1, CVAR_NOINITCALL | CVAR_ARCHIVE) // this is just for testing. -1 means 'inactive'
 {
 	if (self > 4 || self < -1) self = -1;
 }
--- a/wadsrc/static/menudef.txt
+++ b/wadsrc/static/menudef.txt
@@ -2663,6 +2663,16 @@ OptionValue "LightingModes"
 	2, "$OPTVAL_VANILLA"
 }
 
+OptionValue "MapLightModes"
+{
+	-1, "Inactive"
+	0, "$OPTVAL_STANDARD"
+	1, "$OPTVAL_BRIGHT"
+	2, "$OPTVAL_DOOM"
+	3, "$OPTVAL_DARK"
+	4, "$OPTVAL_LEGACY"
+}
+
 OptionValue "Precision"
 {
 	0, "$OPTVAL_SPEED"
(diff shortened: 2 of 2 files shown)


Boondorl:
I think this one will need to cook for a bit since the big problem I still have is that mappers are now forced to include their light mode with their maps if they were designing it for something specific, which can be easily forgotten if they're used to their own default. Basically, I view this as more of a workflow issue. The current defaults match vanilla so if someone has something else in mind, they already know they have to explicitly denote it, otherwise being compatible with the lighting model of other ports

MajorCooke:
Maybe so, but I don't see any harm in restoring it now and making adjustments to it later. 

Also I made sure it could be cherry picked in the event persistence isn't desired.

mc776:
I'm reading this thread in a room on a clear day at 4 PM with a window with broken blinds and the daylight is making a clear gradient on my screen from light blue to dark grey. It would take two men on a weekend to rearrange this room so this monitor does not face the window but my workflow remains remotely sane, while at the same time if we actually did that said monitor would be reflecting glare from the ceiling lights.

There is no way any mapper could have intended how the lighting in any map would look if I booted it up now. And I'm sure other people have worse setups for better excuses.

There were lots of comments in support of immediate restoration of this feature, while creators being so exacting with their lighting are demanding a consistency that simply cannot be enforced in real life. I think it's best to commit this sooner rather than later.

MajorCooke:
Exactly. Lets just get it in, and then we can change it later. Why wait?

Remember, I'm already making use of autoexec.cfg to **force** the cvar to 0 because I hate using the default of -1, it's horrendous and doesn't match *many* of the maps I've played with.

Boondorl:
"Why wait" is because the last time it got removed there was a total shit storm. Adding it back in just to remove it again would not be a good look, and I think there's an implicit understanding of this which is why people want it in now. I understand the desire for this, but I'd really like to pursue some other avenues first, and I will be doing so before 5.0 releases. If those fail, _then_ we can add light modes back in.

Speaking personally, I don't like being rushed out the door on this and it's making the topic toxic very quickly. We have plenty of time. Things will get done, I promise.

nashmuhandes:
On Discord, I mentioned a "meet in the middle" solution that I think would satisfy both creators and users:

We bring back the ambient light level slider. If you're an OG GZDoom user, you'll remember what this was.

Kex Doom also has this, it's simply called Extra Lighting.

Agreed on not rushing this.

src/g_level.cpp
```diff
@@ -148,7 +148,7 @@ CUSTOM_CVAR(Bool, gl_notexturefill, false, CVAR_NOINITCALL)
 	}
 }
 
-CUSTOM_CVAR(Int, gl_maplightmode, -1, CVAR_NOINITCALL | CVAR_CHEAT) // this is just for testing. -1 means 'inactive'
+CUSTOM_CVAR(Int, gl_maplightmode, -1, CVAR_NOINITCALL | CVAR_ARCHIVE) // this is just for testing. -1 means 'inactive'
```
RicardoLuis0, line 151: gl_maplightmode should **not** use CVAR_ARCHIVE

MajorCooke: Hence why I made it cherry pickable.

Not that it matters at this point, Nash and Boondorl don't want it back in so I might as well close this down.

mc776:
> Kex Doom also has this, it's simply called Extra Lighting.

Is it the same as Nugget's Extra Lighting? I find that basically does everything I want.
(and more - the default is zero and you can actually make things *darker*)

Boondorl:
Yep, I'm looking into adding this now. Just need to figure out how to add it to the sw renderer and it should be good to go

PR #197, opened 2025-10-22
Expose polyobjects to ZScript.

inkoalawetrust:
> [!NOTE]
> This is a reupload of [the PR I had initially made for GZDoom](https://github.com/ZDoom/gzdoom/pull/3095). However it also includes the changes that Jay had suggested, like sealing PolyAction and directly exposing FPolyObj's bbox array.

This PR aims to expose polyobjects to ZScript as read only data. Allowing mods to be able to get info such as if a polyobject is present, if it does crushing damage, is a portal, etc.

So far it consists of the following:
- Exposing the polyobjects[] array in LevelLocals (After @MrRaveYard fixed it with 51a069a to not include malformed polyobjects) so that mods can access all polyobjects on any given map.
- Exposing FLevelLocals::GetPolyobj(). For finding polyobjects by ID.
- Exposing the FPolyObj struct itself. With all of its' exposed data set as readonly. And also exposing FPolyObj::GetMirror() instead of MirrorNum itself.
- Making ZScript-side GetStartSpot() and GetCenterSpot() methods for PolyObj. To not need to expose FPolyVertex just for their own vector2 member variable.
- Adding a simple PolyObjPortalType enum to PolyObj for making bHasPortals more readable.
- Exporting ClosestPoint(), for, well, getting the closest point to the polyobject from the specified position. Which is a function that doesn't seem to be used anywhere else anyways.
- Exposing DPolyAction and all its' children. Which are what all the polyobject action specials like doors, rotators, and movers are under the hood (i.e specials like Polyobj_MoveToSpot). The DPolyAction-derived classes all use new (And pre-existing getters), none of the variables themselves are exposed (Since they are all protected and I decided to keep them that way), and PolyAction itself is sealed.

In addition, here is a simple example mod that reads a lot of this exposed data and prints it to the console from a custom weapon. Just spawn the "PolyPistol" class and fire it at a polyobject. It will print a ton of info, and also spawn a Doom candlestick at the nearest point to the polyobject your shot, relative to your position. Try it on something like Hexen's first map.
[PolyFinderPistol.zip](https://github.com/user-attachments/files/20402405/PolyFinderPistol.zip)

### DIFF
--- a/src/playsim/po_man.cpp
+++ b/src/playsim/po_man.cpp
@@ -40,6 +40,7 @@
 #include "g_levellocals.h"
 #include "actorinlines.h"
 #include "v_text.h"
+#include "vm.h"
 
 // TYPES -------------------------------------------------------------------
 
@@ -61,6 +62,8 @@ public:
 	void Construct(FPolyObj *polyNum);
 	void Serialize(FSerializer &arc);
 	void Tick ();
+	DVector2 GetSpeedV() const { return m_Speedv; }
+	DAngle GetAngle() const { return m_Angle; }
 protected:
 	DAngle m_Angle;
 	DVector2 m_Speedv;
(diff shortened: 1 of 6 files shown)


the-phinet:
﻿
﻿

PR #218, opened 2025-10-23
Fixed incorrect stat velocity reading

### LINKED ISSUE #151: [BUG] Stat velocity is inaccurate
### UZDoom version

4.15 pre-671

Doom 2

Windows 11

### Please describe your specific OS version

_No response_

### Relevant hardware info

_No response_

### Have you checked that no other similar issue already exists?

- [x] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

Using ccmd "stat velocity" shows a max SR 40 velocity of 19.34 instead of 21.34, which I got through vel.xy.length() - this is in accord with the value from https://doomwiki.org/wiki/Player#Player_characters ; other traversal speeds are lower as well

I know this is a ridiculously niche feature, but it'd be neat if it was exposed a bit better somehow and cleaned up - it definitely doesn't need two lines and a section that tells you what the current map is.

### Steps to reproduce the behaviour.

Explain how to reproduce
1. 
2.
3.

### Your configuration

```ini

```

### Provide a Log

_No response_

Boondorl:
Previously this read the player's velocity after friction, but now does so before. Also now takes 3D vs 2D movement into account. Also adds support for multiple players and will change the value based on which one is being viewed.

Fixes #151

### DIFF
--- a/src/g_levellocals.h
+++ b/src/g_levellocals.h
@@ -714,6 +714,7 @@ public:
 	int			total_monsters;
 	int			killed_monsters;
 
+	double		cur_velocity;
 	double      max_velocity;
 	double      avg_velocity;
 
--- a/src/g_levellocals.h
+++ b/src/g_levellocals.h
@@ -714,9 +714,34 @@ public:
 	int			total_monsters;
 	int			killed_monsters;
 
-	double		cur_velocity;
-	double      max_velocity;
-	double      avg_velocity;
+	struct VelocityMeasurer {
+		int total = 0;
+		double cur_velocity = 0.0;
+		double max_velocity = 0.0;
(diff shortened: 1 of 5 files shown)


RicardoLuis0:
> Previously this read the player's velocity after friction, but now does so before. Also now takes 3D vs 2D movement into account.
> 
> Fixes #151

might be good to store the velocity per-player, so that it can be shown correctly also when spectating (if spectating is added in the future that is - and for now just doing `level->cur_velocity[consoleplayer]` in stat velocity)

RicardoLuis0:
(same for max/average velocities)

TehRealSalt (approved):

PR #220, opened 2025-10-23
Extended A_TossGib

### LINKED ISSUE #186: [Feature] Expand A_TossGib to take a class name
### UZDoom version

GZDoom current devbuild (not particularly important for the suggestion)

Doom 2

Windows 10

### If Other OS, please describe

_No response_

### Relevant hardware info

_No response_

### Is your feature request related to a problem? Please describe.

Not really a problem as such, but a limitation.
Initially posted as a GZDoom request on the forum:
https://forum.zdoom.org/viewtopic.php?t=80990

### Describe the solution you'd like

A_TossGib currently takes no parameters and just auto-spawns the "Meat" actor or the "Junk" actor, depending on whether the calling actor bleeds or not. It would be useful if A_TossGib could be expanded to take the name of a class to spawn instead of either of these:

`A_TossGib("MyCoolGibs");`

I am aware that there is also A_SpawnDebris but it behaves slightly differently and the setup of the actors that it spawns is different too, as is how they behave. I also know that it can be faked by using A_SpawnProjectile. In both cases, I prefer the behaviour and implementation of A_TossGib.

Being able to specify a class in A_TossGib would make it easier to create drop-in replacement actors, or actors that inherit from others but which spawn different meat/junk.

In the thread, MartinHowe pointed out that the function is already scriptified and that a very simple addition (which I have tested) to the ZScript for the function would enable what I'm asking for.

It's quite easy to use this in a home-cooked actor, but it can't be used to extend the actor class (which can't be done).  If the feature was added to the actor class in uzdoom.pk3, it would make it available to any and all actors without having to add the code as an extension to them in their own ZScript.

Current A_TossGib ZScript
```
void A_TossGib()
	{
		class <Actor> gibtype;
		if (bNoBlood) gibtype = "Junk";
		else gibtype = "Meat";
		Actor gib = Spawn (gibtype, pos + (0,0,24), ALLOW_REPLACE);

		if (gib == null)
		{
			return;
		}

		gib.Angle = random[GibTosser]() * (360 / 256.);
		gib.VelFromAngle(random[GibTosser](0, 15));
		gib.Vel.Z = random[GibTosser](0, 15);
	}
```

Suggested change:
```
void A_TossGib(class <Actor> gibtype = null)
	{
		if (!gibtype)
		{
		    if (bNoBlood) gibtype = "Junk";
		    else gibtype = "Meat";
		}
		
		Actor gib = Spawn (gibtype, pos + (0,0,24), ALLOW_REPLACE);

		if (gib == null)
		{
			return;
		}

		gib.Angle = random[GibTosser]() * (360 / 256.);
		gib.VelFromAngle(random[GibTosser](0, 15));
		gib.Vel.Z = random[GibTosser](0, 15);
	}
```

### Describe alternatives you've considered

_No response_

### Add any other context or screenshots about the feature request here.

_No response_

Boondorl:
Added class name and z offset parameters. Now returns the gib that spawned.

Implements #186

### DIFF
--- a/wadsrc/static/zscript/actors/strife/strifefunctions.zs
+++ b/wadsrc/static/zscript/actors/strife/strifefunctions.zs
@@ -54,21 +54,24 @@ extend class Actor
 	//
 	//==========================================================================
 
-	void A_TossGib()
+	Actor A_TossGib(class<Actor> gibtype = null, double zOfs = 24.0)
 	{
-		class <Actor> gibtype;
-		if (bNoBlood) gibtype = "Junk";
-		else gibtype = "Meat";
-		Actor gib = Spawn (gibtype, pos + (0,0,24), ALLOW_REPLACE);
+		if (!gibtype)
+		{
+			if (bNoBlood) gibtype = "Junk";
+			else gibtype = "Meat";
+		}
+		Actor gib = Spawn (gibtype, pos.PlusZ(zOfs), ALLOW_REPLACE);
 
 		if (gib == null)
 		{
-			return;
+			return null;
 		}
 
 		gib.Angle = random[GibTosser]() * (360 / 256.);
 		gib.VelFromAngle(random[GibTosser](0, 15));
 		gib.Vel.Z = random[GibTosser](0, 15);
+		return gib;
 	}
 
 	//==========================================================================


RicardoLuis0:
will try to test and approve this tomorrow/during the weekend (just to make sure dehacked/etc won't break - i don't think it will, but it's good to test first), but if anyone can test it now feel free to approve it before

TehRealSalt (approved):

the-phinet:
Can you rebase this onto the current trunk?

PR #226, opened 2025-10-24
Accelerate YUV-to-RGB conversion in GPU

Cacodemon345:
This PR accelerates YUV-to-RGB conversion in the GPU instead of doing it on the CPU.

### DIFF
--- a/src/common/textures/animtexture.cpp
+++ b/src/common/textures/animtexture.cpp
@@ -52,27 +52,13 @@ void AnimTexture::SetFrameSize(int  format, int width, int height)
 	memset(Image.Data(), 0, Image.Size());
 }
 
-//TODO optimize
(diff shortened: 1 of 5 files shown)


RicardoLuis0:
as-is this wouldn't help much, i was planning to upload the texture including stride data/etc

RicardoLuis0:
that way any and all cpu loops are removed

Cacodemon345:
I don't have enough knowledge to add the 3 grayscale textures that'd be needed to eliminate this altogether (since it'd simplify YUV sampling here a lot) so it's probably going to require some more thinking.

RicardoLuis0:
not 3 images, just passing the strides to the shader and calculating x/y separately there for y/u/v

PR #228, opened 2025-10-24
Add translucent canvas textures support

Cacodemon345:
Adds support for translucent canvas textures.

[map01trans.zip](https://github.com/user-attachments/files/23151354/map01trans.zip)

Example file.

image

### DIFF
--- a/src/common/2d/v_draw.cpp
+++ b/src/common/2d/v_draw.cpp
@@ -1715,7 +1715,7 @@ DEFINE_ACTION_FUNCTION(FCanvas, Clear)
 
 void DoDim(F2DDrawer *drawer, PalEntry color, float amount, int x1, int y1, int w, int h, FRenderStyle *style)
 {
-	if (amount <= 0)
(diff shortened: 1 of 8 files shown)


Boondorl:
~~I think I'd move this check into a `IsTranslucentHardwareCanvas()` function, that way this check doesn't need to be written in full for every API since it's agnostic from how they work. Other than that it seems to check out~~

This feels like a weird can of worms due to that weird buffer creation function in the translucency checker, maybe we'll just handle that later.

RicardoLuis0 (dismissed):
code looks good, but don't merge until nash has had the time to test it in-game

nashmuhandes:
Tested and seems to be good, my only suggestion is to:

1) Replace the example file to prevent future users from copy/pasting the `STYLE_Source` usage in scripts
2) Squash commit?

Otherwise, works as advertised (finally!)

Cacodemon345:
Example file updated.

nashmuhandes (approved):
All good to me!

PR #231, opened 2025-10-25
Removed net mode selection and set packet-server as the default

Boondorl:
With the final engine desync being fixed in packet-server mode, peer-to-peer mode has little reason to be used and will be unable to support certain features in the future. As such, packet-server will now always be how net games are handled for better stability and a more expected experience alongside being easier to maintain overall.

Also, VisualStudio didn't like setting FString to `'\0'` so I changed it to an empty string literal.


RicardoLuis0 (approved):

nashmuhandes:
Tested with an actual game, 6 players, both unmodded and with light mods, seems to be working with no problems

nashmuhandes (approved):

PR #241, opened 2025-10-25
Update v_video assumed cache line size from 32 -> 64 bytes

River-Salmon:
That's a far more common configuration with modern computers.

### DIFF
--- a/src/common/utility/x86.h
+++ b/src/common/utility/x86.h
@@ -4,6 +4,11 @@
 #include "basics.h"
 #include "zstring.h"
 
+#ifdef __cpp_lib_hardware_interference_size
+#include "version"
+#include <new>
+#endif
+
 struct CPUInfo	// 92 bytes
 {
 	union
@@ -223,8 +228,11 @@ struct CPUInfo	// 92 bytes
 		};
 		uint32_t AMD_DataL1Info;
 	};
-
-	static inline constexpr uint8_t AssumedDefaultCacheLineSizeBytes = 64;
+#ifdef __cpp_lib_hardware_interference_size
+	static inline constexpr size_t AssumedDefaultCacheLineSizeBytes = std::hardware_destructive_interference_size;
+#else
+	static inline constexpr size_t AssumedDefaultCacheLineSizeBytes = 64;
+#endif
 };
 
 
--- a/src/common/utility/x86.h
+++ b/src/common/utility/x86.h
@@ -223,6 +223,8 @@ struct CPUInfo	// 92 bytes
 		};
 		uint32_t AMD_DataL1Info;
 	};
+
+	static inline constexpr uint8_t AssumedDefaultCacheLineSizeBytes = 64;
 };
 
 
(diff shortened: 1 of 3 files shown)


RicardoLuis0:
seems that that assumption is in multiple other places as well https://github.com/UZDoom/UZDoom/blob/master/src/common/utility/x86.cpp#L45
https://github.com/UZDoom/UZDoom/blob/master/src/common/utility/x86.cpp#L75

RicardoLuis0:
this bit of code should probably be re-evaluated as well, it refereces very old CPUs and i doubt the arhictecture of anything at all modern is in any way similar
https://github.com/UZDoom/UZDoom/blob/master/src/common/rendering/v_video.cpp#L276-L288

River-Salmon:
> this bit of code should probably be re-evaluated as well, it refereces very old CPUs and i doubt the arhictecture of anything at all modern is in any way similar https://github.com/UZDoom/UZDoom/blob/master/src/common/rendering/v_video.cpp#L276-L288

Might be worth removing the "optimize pitch" path entirely since it's based around 20+ year old cpus

River-Salmon:
I've updated the remaining cases that I found

src/common/utility/x86.h
```diff
@@ -223,6 +223,8 @@ struct CPUInfo	// 92 bytes
 		};
 		uint32_t AMD_DataL1Info;
 	};
+
+	static inline constexpr uint8_t AssumedDefaultCacheLineSizeBytes = 64;
```
kcat: C++ has [`std::hardware_destructive_interference_size`](https://en.cppreference.com/w/cpp/thread/hardware_destructive_interference_size.html) defined for the target system, which is essentially the L1 cache line size and may be better than hardcoding a specific value. Some compilers/STLs don't have it (should be checked by the `__cpp_lib_hardware_interference_size` feature test macro and clang/libc++ version, as some libc++ versions erroneously define the feature test macro despite not having the variable) so can still use 64 or whatever as a fallback if `std::hardware_destructive_interference_size` isn't available. May also need to build with `-Wno-interference-size` since some compilers warn that the value is compile-time despite really being defined by the runtime CPU.

RicardoLuis0: that's a constant though, this default is just a fallback for if grabbing it from CPUID fails - it should be something generic for any CPU, not for the one being compiled for

River-Salmon: Using destructive interference size could be good future work. But yeah, this is the fallback value in case there was no other detected size

kcat: The compiler knows the general target its compiling for, and can set the value as appropriate. ARM vs x86-64 vs RISC-V vs ... If `-march=...` or `-mtune=...` might influence it, then shouldn't it? That's the point of tuning optimizations for the specified CPU. The CPUID checks are done first to get the runtime value if possible, yes, but I think `std::hardware_destructive_interference_size` is a better second-try fallback since the compiler has some idea of what it's compiling for, before going with a hand-picked number like `64` across the board.

River-Salmon: Updated to use the destructive interference size

the-phinet (approved):

PR #242, opened 2025-10-25
Build translation files from standard translation database

the-phinet:
Build translation files from git repo managed by weblate, containing gnu gettext files


RicardoLuis0:
seems it fails on MacOS due to missing gawk

Boondorl (dismissed):

TehRealSalt (dismissed):
I've only noticed one regression -- `OPTVAL_NONE` doesn't appear to have been carried over. I noticed this on Options -> Player Setup -> Team. This is probably just an issue for UZDoom/Translation though, the actual code here looks good :+1:

the-phinet:
> I've only noticed one regression -- `OPTVAL_NONE` doesn't appear to have been carried over. I noticed this on Options -> Player Setup -> Team. This is probably just an issue for UZDoom/Translation though, the actual code here looks good 👍

Interesting. Since I used python to build the initial language files, I wonder if I had a type issue where the "None" string was being treated as the `None` value in python and being skipped. This has been fixed now

TehRealSalt (approved):

Boondorl (approved):

RicardoLuis0 (approved):

PR #243, opened 2025-10-25
Global storage

TehRealSalt:
Cherry-picked from Selaco. Implements a canon way for mods to save variables between multiple sessions, instead of abusing console variables for it.

Related to #5, but not exactly a replacement for it. (Implementing that could still be useful for cross-port compatibility.)

# Lots to do

- [ ] Rework for multiplayer support. These are a straight-forward key-value pairs, it should be easy to do exactly the same as console variables
- [ ] Use separated storage, limit one per mod
  - [ ] Add a place to reserve your file's name in advance (`MAPINFO/GameInfo`?), and to signal the game to load it into memory
  - [ ] Read-only access to other mods' storage
- [ ] Implement a size limit
  - [ ] Size limit cvar
  - [ ] Warn the user in the console once when this limit is breached
  - [ ] Determine how to handle altering the storage after the limit is reached. (Probably keep the changes in memory, in case the user wants to change their limit afterwards.)

### DIFF
--- a/src/m_misc.cpp
+++ b/src/m_misc.cpp
@@ -118,6 +118,25 @@ DEFINE_ACTION_FUNCTION(_Globals, GetInt)
 }
 
 
+DEFINE_ACTION_FUNCTION(_Globals, GetKeys)
+{
+	PARAM_PROLOGUE;
+	PARAM_POINTER(out, TArray<FString>);
+
+	// Copy keys to out
+	TMapIterator<FString, FString> it(globalStorage);
+	TMap<FString, FString>::Pair* pair;
+	int outSize = out->Size();
+
+	while (it.NextPair(pair))
+	{
+		out->Push(pair->Key);
+	}
+
+	ACTION_RETURN_INT(out->Size() - outSize);
+}
+
+
 DEFINE_ACTION_FUNCTION(_Globals, Set)
 {
 	PARAM_PROLOGUE;
--- a/src/m_misc.cpp
+++ b/src/m_misc.cpp
@@ -33,6 +33,7 @@
 #include <stdlib.h>
 #include <errno.h>
 #include <time.h>
+#include <fstream>
 
 #include "r_defs.h"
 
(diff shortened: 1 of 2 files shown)


Boondorl:
I'm not entirely sure how I feel about this one. In particular, I'm on the fence if we should be encouraging mods to rely on something like this. Globals are often best handled through singletons or, in the case of controlling the engine, cvars, and this feels a bit too C-ish for my taste. Technically it is gated behind an API, but the accessibility all but encourages people to throw random things in there which can only lead to really bad paradigms that have impossible to track down bugs (at least we'll have the debugger going forward).

Some things I would personally like to see if we stick with this:
-Names for keys. I've never liked strings for this kind of thing and Names are more efficient to hash and more consistent with how existing named fields work in ZScript
-A proper variant type (likely using `std::variant`) for storing values. String conversions can be expensive and lossy for floats
-Default value arguments for getters if a key doesn't exist. This is just super useful in general

I'll have to see how I feel about this once we start chipping away at it, but I think it's going to be contentious in its current form. Fine for a standalone title but probably not fitting for upstream

RicardoLuis0:
> String conversations can be expensive and lossy for floats

not saying variants aren't needed, but floats can be stringified to hex floats, without any loss
https://godbolt.org/z/f7d7aj4r6

Boondorl:
Another possible knock on effect that I'm worried about is people resorting to globals instead of cvars which can hurt accessibility overall. Some things that would normally be controllable become entirely locked behind scripting. For some features this is appropriate, but for others it's just pointless gatekeeping. We can chalk that up to mods not designing things well, but we have a chance to enforce better design paradigms and loosening up on that is fully on us if we add something that ends up being a pain more often than not.

TehRealSalt:
Nah I actually agree, those are pretty valid concerns.

I think if we did roll with this, we definitely need a built-in command to change them manually. That would at least go a little ways into helping the accessibility concern, but I also concede that wouldn't have auto-complete like cvars can either.

The main reason I wanted this over just continuing to use console variables is simply to keep them away from your settings. Every mod that uses them for achievements or unlockables will get them wiped by "reset to default" in the menus, when ideally it'd be another option to reset global progress. It also works better with keeping your saves on cloud storage. (Sorry, shoulda clarified that in the description itself)

I know other people want this feature for things like using arbitrary keys, but I mainly just want the separation of concerns. How would you feel about a cvar flag to save it in a separate file? I was worried about it making CVARINFO definitions even more clumsy, but the system has so many advantages that I haven't been convinced it would be a terrible approach

MrRaveYard:
What if I load two mods each trying to write/read their own storage?

TehRealSalt:
> What if I load two mods each trying to write/read their own storage?

That is the issue that most of the to-do list in the main post is about, yes

Boondorl:
I've been thinking about this one a bit over the past week, and from the sounds of it, I think what we're really describing here is dynamic cvars. It might not be a bad idea to recontextualize it like that instead of as just raw memory storage

TehRealSalt:
I'm still interested in this, but I will probably start a new branch for it when it comes time :v:

PR #246, opened 2025-10-26
Tweaked defaults

Boondorl:
-Map colors now default to game type (added new Default option)
-Master volume set to 0.5 to prevent blowing out system audio easily
-Menu mouse mode set to On (non-touchscreen mode); makes the mouse's position more readable
-Save sorting order set to time-based
-Render quality set to Quality to patch holes by default
-AF set to 16x since it's cheap and looks way better
-ENDOOM screen set to off to make exiting more streamlined
-Status bar and fullscreen HUDs set to scale to fullscreen for better readability
-Simple options menu turned off (this will be replaced later)
-Rocket trails set to off to better match vanilla Doom
-Enhanced night vision mode set to off
-Sprite clipping set to Forced-Perspective to better match vanilla Doom
-Particle style set to square since it better matches Doom's aesthetic
-Fuzz type set to software to better match vanilla Doom

Screenshot_Doom_20251025_225427

### DIFF
--- a/src/am_map.cpp
+++ b/src/am_map.cpp
@@ -171,7 +171,7 @@ CVAR(Bool, am_showitems, false, CVAR_ARCHIVE);
 CVAR(Bool, am_showtime, true, CVAR_ARCHIVE);
 CVAR(Bool, am_showtotaltime, false, CVAR_ARCHIVE);
 CVAR(Bool, am_showlevelname, true, CVAR_ARCHIVE);
-CVAR(Int, am_colorset, 0, CVAR_ARCHIVE);
+CVAR(Int, am_colorset, -1, CVAR_ARCHIVE);
 CVAR(Bool, am_customcolors, true, CVAR_ARCHIVE);
 CVAR(Int, am_map_secrets, 1, CVAR_ARCHIVE);
 CVAR(Int, am_drawmapback, 1, CVAR_ARCHIVE);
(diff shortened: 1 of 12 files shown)


mc776:
> AF set to 16x since it's cheap

TIL! I always thought that could be a big performance hit...

Boondorl:
Nah, on modern hardware especially AF is basically free. I'm not entirely sure why 8x was the default since it actually had some pretty gnarly artifacting as early as that opening shot in Doom 2. Was probably just set a really long time ago and no one ever got around to updating the defaults.

axredneck:
> Render quality set to Quality to patch holes by default

Last time i tried it it introduced major slowdown (at least in Action Doom 2 mod). It was several years ago though.

TehRealSalt (approved):

Boondorl:
> > Render quality set to Quality to patch holes by default
> 
> Last time i tried it it introduced major slowdown (at least in Action Doom 2 mod). It was several years ago though.

It could be a problem on older hardware, but from all the surveys Graf has taken, very few people are using hardware old enough to where it will be noticeable. My hardware has never been the best but I've always played with it on and never noticed any performance issues over the past decade or so. We have plans to redo the menu relatively soon which will hopefully make all of this easier to deal with rather than needing to dig through a mess of 30 submenus to find it (plus we'd like to look into adding graphics options to the launcher itself). For now we felt it was best to leave it on so people aren't confused as to why there are seams showing when they boot up the game for the first time because it's a very difficult option to track down.

River-Salmon:
In theory, > 8x AF can be a burden on the scheduling system of the GPU and cause not-so-great parallelism, however, I think you'd be hard pressed to find a scenario where this matters on anything running on UZDoom. There just isn't enough GPU work for the machine to do. It might be worth revisiting if/when the cpu side of the rendering is no longer the main bottleneck for big maps.

mc776:
Any thoughts about making ZDoom's random-nondetection invisibility powerup effect off by default? Nevermind that it makes the gameplay worse, the discrepancy can throw people off pretty hard when they're not expecting it coming in from other ports.

the-phinet (approved):

Boondorl:
> Any thoughts about making ZDoom's random-nondetection invisibility powerup effect off by default? Nevermind that it makes the gameplay worse, the discrepancy can throw people off pretty hard when they're not expecting it coming in from other ports.

I was thinking about this while playing through Doom 1 to test things. The main issue is the compat flag is just a bit so I can't add any flexibility to it. I'd love to have a Default option to use the game's preference, but it'd also be nice to have a way to add it without needing to use GHOST to always activate it. What I might do is add a ShadowCompat flag to Actors that'll signal to the engine it should use the game's default behavior instead. That or just add a compat flag for ZDoom's behavior so we'd have a compat flag for a compat flag lol

mc776:
Do any other ports use this flag? What if we just ditched it in favour of a cvar or something?

PR #259, opened 2025-10-26
Define SHADOW sneak behavior based on game

Boondorl:
The current SHADOW behavior always allows sneaking since this was brought over from Raven's version of it (while also adding a GHOST flag to emulate its fuller behaviors). This has very odd effects in Doom, so it now only applies this behavior to Raven games specifically. A compat flag, compat_zdoomshadow, was added to bring back the old behavior of allowing it in any game.

### DIFF
--- a/src/d_main.cpp
+++ b/src/d_main.cpp
@@ -844,7 +844,7 @@ CUSTOM_CVAR(Int, compatmode, 0, CVAR_ARCHIVE|CVAR_NOINITCALL)
 
 	case 4: // Old ZDoom compat mode
 		v = COMPATF_SOUNDTARGET | COMPATF_LIGHT;
-		w = COMPATF2_MULTIEXIT | COMPATF2_TELEPORT | COMPATF2_PUSHWINDOW | COMPATF2_CHECKSWITCHRANGE | COMPATF2_NOMBF21;
+		w = COMPATF2_MULTIEXIT | COMPATF2_TELEPORT | COMPATF2_PUSHWINDOW | COMPATF2_CHECKSWITCHRANGE | COMPATF2_NOMBF21 | COMPATF2_ZDOOMSHADOW;
 		break;
 
 	case 5: // MBF compat mode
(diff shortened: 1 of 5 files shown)


Boondorl:
TCs and standalones are actually my biggest concern with a change like this, but looking at how sneaking was implemented, I doubt most people even knew about it and the rest boiled down to sight being purely random. I don't know if this behavior specifically was ever intentionally designed around, but we'll have to see. For that reason it's probably best to keep this "internal" for now.

mc776:
I'd be surprised if anything actually relied on the ZDoom behaviour, given how unrewardingly random it is - there's no real gameplay to it, nothing seems to affect your chances of being seen.

Would it help to post something on the ZDF and DW forums asking people to submit any stealth-based (non-PvP) mods?

Boondorl:
Part of the issue is the weird hardcoded sneaking requirements Raven used. You have to be greater than 128 xy units away and moving less than 5 units/tic. That is extremely slow as only crouch walking will let you consistently move slower than that, otherwise needing to tap move. The other condition is just randomness, so yeah, I don't think it would ever be thought of as intentional. I could possibly see a standalone existing that used sneaking, but I imagine the number is so small that the compat flag will easily handle it.

mc776:
TIL such a thing existed! (And in case anyone's wondering, HD's running (not sprinting) at all default settings with no equipment barely breaks 4 units/tic. I can't see anything that could be accidentally triggering this behaviour but the way some monsters take forever to react to the player sometimes I'm starting to wonder...)

Is this consistent for *all* Raven games?

Boondorl:
> TIL such a thing existed! (And in case anyone's wondering, HD's running (not sprinting) at all default settings with no equipment barely breaks 4 units/tic. I can't see anything that could be accidentally triggering this behaviour but the way some monsters take forever to react to the player sometimes I'm starting to wonder...)
> 
> Is this consistent for _all_ Raven games?

It's how Raven programmed it for Heretic, yes. ZDoom simply allowed all forms of SHADOW to use it because originally it was baked into how Raven updated the SHADOW flag. This kinda also confirms that sometimes it's even undesired behavior lol

TehRealSalt (approved):

Boondorl:
If the auto compat thing gets in I'll probably redo this with that since imo it captures a wider use case

the-phinet:
﻿
﻿

PR #267, opened 2025-10-27
Remove translucent renderstyle from Doom's bullet puff

nashmuhandes:
This one might be contentious so I'd like to know what peoples' thoughts are on this.

My argument for it is that the translucent renderstyle greatly reduces the visual impact from the bullet puffs, and in dark rooms, they're straight up almost unseeable.

Here are some video comparisons comparing the translucent style versus the classic, solid style.

https://github.com/user-attachments/assets/215d50de-4ea4-4de3-a439-f01723ae36f3

https://github.com/user-attachments/assets/74de5822-9f17-4a68-8e5a-54687ccc1ff7

### DIFF
--- a/wadsrc/static/zscript/actors/doom/doommisc.zs
+++ b/wadsrc/static/zscript/actors/doom/doommisc.zs
@@ -61,9 +61,6 @@ class BulletPuff : Actor
 		+NOGRAVITY
 		+ALLOWPARTICLES
 		+RANDOMIZE
-		+ZDOOMTRANS
-		RenderStyle "Translucent";
-		Alpha 0.5;
 		VSpeed 1;
 		Mass 5;
 	}


River-Salmon:
Any way it can be made into a user option?

nashmuhandes:
A user option, exposed in the menu, was something I've also thought of, but wasn't sure if it was the better way forward.

If it's preferred, I can change the PR to expose a user option instead.

Boondorl:
Out of curiosity, does anyone know why was it set to translucent? This feels like something the engine should've been handling rather than defaulting to it if it's some technical thing like dehacked. I imagine the argument now is that too many things inherit from it and rely on this behavior, but I'm wondering why it was added.

River-Salmon:
So mods can still control whether to make their puffs transparent, right? If this is just the vanilla puffs then yeah I say just return them to being solid

nashmuhandes:
> Out of curiosity, does anyone know why was it set to translucent? This feels like something the engine should've been handling rather than defaulting to it if it's some technical thing like dehacked. I imagine the argument now is that too many things inherit from it and rely on this behavior, but I'm wondering why it was added.

Perhaps @rheit could provide some insight as to the history of the translucent effect?

inkoalawetrust:
@River-Salmon Yeah, there's no reason why this would prevent mods from having transculent puffs, just make a custom mod puff that turns those flags and properties back on, there's even several methods to make the stock puff look like that (Like event handlers or modifying them post-spawn as they are created) if you really want to do.

mc776:
Aesthetically I prefer the solid puffs but can understand why this particular thing (which for most of its lifetime is primarily a cloud of smoke and dust) would be translucent.

rheit:
> Perhaps @rheit could provide some insight as to the history of the translucent effect?

Sure! Doom didn't have transparency effects, so when they were added, there was the question of what to add them to to enhance the experience. What would id likely have made transparent had they had that capability? The bullet puff was on that list because it's not a solid object.

Though really, I probably just copied DOSDoom and/or Doom Legacy, which had done transparency first so had already made their own lists of objects. It's been almost 28 years now, and that's a long time to remember anything.

rheit:
> I imagine the argument now is that too many things inherit from it and rely on this behavior

Also, this. The time to unconditionally remove the effect was before DECORATE was a thing. You're pretty much guaranteed to be breaking mods by changing a core game actor like this now.

mc776:
I can imagine some old DECORATE mod being like "SpookyFadingGhost : BulletPuff" making an advanced zombieman version of the mummy ghost animation in Heretic...

nashmuhandes:
Well, after a bit of re-thinking about this, yeaaaa, this is probably too late to change. Closing.

PR #268, opened 2025-10-27
allow defcvars from pwads

madame-rachelle:
The blocking code is `#if 0`'d for now, but it can be removed completely.

### DIFF
--- a/src/d_defcvars.cpp
+++ b/src/d_defcvars.cpp
@@ -45,6 +45,7 @@ void D_GrabCVarDefaults()
 
 	while ((lump = fileSystem.FindLump("DEFCVARS", &lastlump)) != -1)
 	{
+#if 0 // leaving the lines in place for now (for gzdoom sync) - can remove if needed
 		// don't parse from wads
 		if (lastlump > fileSystem.GetLastEntry(fileSystem.GetMaxIwadNum()))
 		{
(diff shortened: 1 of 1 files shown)


Boondorl:
I think something like this will require a lot of deliberating on the best way to virtualize it so mods aren't a settings nightmare. I think the IWAD defaults should still take priority where applicable. One option here would be to add a flag for cvars that denotes they actually are accessible through pwad defcvars, that way performance/accessibility cvars aren't targeted.

andOlga:
I strongly agree that this at least needs to have a way to store settings per PWAD. With IWADs this is handled by the `AutoName` param in `IWADINFO`, which is more or less required for standalone games so this doesn't cause any problems there, but with PWADs, most of them don't have any identifiers and we cannot expect that to change. An ini section based on the file path, hash, or combination thereof would therefore be needed if we want mods to have such strong influences over the settings.

I also do realize that there is already a blacklist of settings that can't be changed via DEFCVARS, but I can't help but wonder if it needs to be extended to include more stuff... Or even the *majority* of stuff. I don't want a simple mod changing my automap settings, for example. Hell, I would probably want a toggle to turn this off altogether.

Would it perhaps make more sense to expose more settings to MAPINFO, the way we already do for `compat_` CVARs and Jump/Crouch? Or otherwise have some limited mechanism of exposing only specifically "blessed" variables that are okay for mods to change, rather than give them the same list as we give to IWADs?

Boondorl:
I think for now I'll close this PR since we're going to be taking a different approach from the discussions we've had, but I can re-open it later if needed

PR #274, opened 2025-10-27
L3 Prefetch in FBlockThingsIterator::Next

River-Salmon:
Sometimes shows performance improvements in large slaughtermaps. This is in theory a fair prefetch assumption to make, since the NextActor is basically always of interest during this iteration.

### DIFF
--- a/src/common/utility/basics.h
+++ b/src/common/utility/basics.h
@@ -4,7 +4,10 @@
 #include <stddef.h>
 #include <stdint.h>
 #include <type_traits>
+
+#if defined(_M_X64) || defined(__x86_64__) || defined(__amd64__) || defined(__amd64) || defined(__x86_64__) || defined(__x86_64)
 #include <xmmintrin.h>
+#endif
 
 #define MAXWIDTH 12000
 #define MAXHEIGHT 5000
--- a/src/common/utility/basics.h
+++ b/src/common/utility/basics.h
@@ -4,6 +4,7 @@
 #include <stddef.h>
 #include <stdint.h>
 #include <type_traits>
+#include <xmmintrin.h>
 
 #define MAXWIDTH 12000
 #define MAXHEIGHT 5000
@@ -79,8 +80,7 @@ T clamp(T val, T minval, T maxval)
 
 static inline void PrefetchL3(const void* Address)
 {
-#ifdef _M_X64
-    static constexpr int L3CacheHint = 2;
-    _mm_prefetch(static_cast<const char*>(Address), L3CacheHint);
+#if defined(_M_X64) || defined(__x86_64__) || defined(__amd64__) || defined(__amd64) || defined(__x86_64__) || defined(__x86_64)
+    _mm_prefetch(static_cast<const char*>(Address), _MM_HINT_T1);
 #endif
 }
--- a/src/common/utility/basics.h
+++ b/src/common/utility/basics.h
@@ -1,8 +1,9 @@
 #pragma once
 
+#include <algorithm>
 #include <stddef.h>
 #include <stdint.h>
-#include <algorithm>
+#include <type_traits>
 
 #define MAXWIDTH 12000
 #define MAXHEIGHT 5000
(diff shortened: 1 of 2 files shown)


MajorCooke:
This doesn't change Next()'s current return values, does it? As in the first time it's called, the actual first actor is returned, right? If it does, great. 

If not, that could be a problem with backwards compatibility.

RicardoLuis0:
> This doesn't change Next()'s current return values, does it? As in the first time it's called, the actual first actor is returned, right? If it does, great.
> 
> If not, that could be a problem with backwards compatibility.

yeah, it just tells the CPU that "memory at X location will be needed soon, so make sure it's available"

RicardoLuis0 (dismissed):

TehRealSalt (dismissed):

src/common/utility/basics.h
```diff
@@ -75,3 +76,11 @@ T clamp(T val, T minval, T maxval)
 {
     return std::max<T>(std::min<T>(val, maxval), minval);
 }
+
+static inline void PrefetchL3(const void* Address)
+{
+#ifdef _M_X64
+    static constexpr int L3CacheHint = 2;
+    _mm_prefetch(static_cast<const char*>(Address), L3CacheHint);
```
kcat: Make sure to include the appropriate header (`xmmintrin.h`), or else this can fail to compile depending on what other headers are included alongside this one. You could also check the `__SSE__` macro along with `_M_X64` for GCC/Clang to pick up on support for SSE too (`_M_X64` is MSVC-only to detect a 64-bit target). That header should also define `_MM_HINT_T1` as the proper hint name to use, rather than the magic number `2`.

River-Salmon: Good catch. We might want a standardized way of doing platform target detection in this project, something like Boost.Predef though that might be overkill

River-Salmon: Updated to use more of the known platform macros. I don't do a separate check for SSE since that's required by the x64 specification anyway.

RicardoLuis0 (approved):

rheit:
I'd be interested in seeing any before/after performance numbers you collected with this change.

RicardoLuis0:
> I'd be interested in seeing any before/after performance numbers you collected with this change.

image

kcat:
Which CPU is that on? I see here from a few years ago:
https://stackoverflow.com/questions/65604355/understanding-mm-prefetch
it mentions "AMD processors alias PREFETCH1 and PREFETCH2 to PREFETCH0". Since this tries to load into L3 cache, it may load into all levels of cache and cause different performance effects compared to other CPUs that may be more precise.

Also, how does it affect "simpler" maps? I see it mentioned here:
https://zzqcn.github.io/perf/intel_opt_manual/7.html
that "For example specifying a NULL pointer (0L) as address for a prefetch can cause long delays," which will inevitably happen at the end of a linked list. Depending on how that long delay is, its effect can be more apparent when there's fewer items in the list.

Boondorl:
That's true, but maybe not relevant as smaller maps don't struggle to get high numbers anyway. I'll run some testing on my own machine to see how it fairs, but realistically a change like this is heavily targeted towards helping with slaughtermaps where thousands of Actors are active at once (this was actually how it was found; Salmon was looking at the performance times on Nuts and noticed the blockmap was the second most expensive thing behind the sprite rendering). It's possible a chunk of FPS could be lost here, but going from 900 FPS down to 600 FPS isn't really that big of a concern compared to helping out in the worst case scenarios like Sunder.

Edit: Tried Nuts on my Ryzen 5 3600, didn't notice any difference. I'm guessing it must be something else on my machine slowing down performance, but it didn't have a positive nor negative impact from what I can tell.

kcat:
> That's true, but maybe not relevant as smaller maps don't struggle to get high numbers anyway.

Right. Since it allows nuts.wad to stay just above 60, I don't imagine it could be worse than that on simpler maps. If it causes a drop from 900 fps to 800, that's no big deal. But if it drops from 900 to like 90 or 100, that would be more concerning. Unfortunately that page doesn't clarify "long delays", and since "long" is a relative term that could mean a number of things, I think it wouldn't hurt to check.

PR #283, opened 2025-10-28
Trace fix and minor refactor

dileepvr:
Engine internal was unkludged. `FPathTraverse::Next()` now actually advances to next intercept. Somehow it was made to mostly work, so I had to remove all workarounds and make it work again.

- Fixed trace behavior when crossing flats without crossing lines when passing through sector portals. 
- Properly populate `Results->ffloor` when hitting 3D-floor flat. 
- Crossing liquid top surfaces only populates `Results->Crossing[3D]Water` when intercept is within sector.
- Also fixed OoB usage of trace  on 3D floors (which is not exposed to zscript).

`Crossing3DWater` only populates with the first top surface of liquid that is crossed. I assume that this is intended behavior.

[edit]
Added a `TRACE_3DLiquidCallback` flag to run callback function when a 3DFloor liquid-surface crossing occurs.

`HitType` will be `TRACE_HitFloor` for top crossing, `TRACE_HitCeiling` for bottom crossing, and `TRACE_HitWall` for side-wall crossing. `Crossed3DWater` and `Crossed3DWaterPos` will get populated. ~~It is left to the modder to calculate if the 3D-liquid volume is being entered or exited.~~ Added a `bool entering3DLiquid` field to tell whether entering or exiting 3D-liquid volume when callback is triggered from `TRACE_3DLiquidCallback` flag.
The return from callback will be ignored and assumed to be `TRACE_Continue`, just like for `TRACE_ReportPortals`.
The callback is not triggered if the 3D-liquid volume crossing surface is flush with map geometry. So starting inside a swimming pool and pointing a trace at the floor will not be reported as a 3D-liquid crossing (unless the pool bottom surface is magically floating in mid-air above the actual sector floor). Same with swimming pool walls.
[/edit]

~~Keeping this a draft until after 4.13.3 releases.~~
~~UZD can't afford new playsim bugs in a debut release version.~~

Test map uses m8f's code for laser sight as a base. Look for color changes in particle beam and some console messages: 
[tracer_visualizer_v5.zip](https://github.com/user-attachments/files/31335754/tracer_visualizer_v5.zip)
Test map without new flags (`TRACE_3DLiquidCallback`) or features (checking the `entering3DLiquid` field) for comparing against current trunk (two implementations of tracer visualization):
[tracer_visualizer_v5_no_new_flags.zip](https://github.com/user-attachments/files/31335767/tracer_visualizer_v5_no_new_flags.zip)
[tracer_visualizer_v2.zip](https://github.com/user-attachments/files/31335768/tracer_visualizer_v2.zip)

### DIFF
--- a/src/playsim/p_maputl.cpp
+++ b/src/playsim/p_maputl.cpp
@@ -1399,20 +1399,19 @@ void FPathTraverse::AddThingIntercepts (int bx, int by, FBlockThingsIterator &it
 intercept_t *FPathTraverse::Next()
 {
 	intercept_t *in = NULL;
-
 	double dist = FLT_MAX;
-	for (unsigned scanpos = intercept_index; scanpos < intercepts.Size (); scanpos++)
+
+	if (intercept_count > 0 && intercept_index < intercept_count)
 	{
-		intercept_t *scan = &intercepts[scanpos];
-		if (scan->frac < dist && !scan->done)
-		{
-			dist = scan->frac;
-			in = scan;
-		}
+		in = &intercepts[intercept_index++];
+		dist = in->frac;
 	}
-	
-	if (dist > 1. || in == NULL) return NULL;	// checked everything in range			
-	in->done = true;
+	else
+	{
+		return NULL;
+	}
+	if (dist > 1. || in == NULL) return NULL;	// checked everything in range
+	// in->done = true;
 	return in;
 }
 
@@ -1630,6 +1629,17 @@ void FPathTraverse::init(double x1, double y1, double x2, double y2, int flags, 
 			break;
 		}
 	}
+
+	struct
+	{
+		bool operator()(intercept_t a, intercept_t b) const { return a.frac < b.frac; }
+	} lesserfrac;
+	intercept_count = intercepts.Size();
+	if (intercept_count > intercept_index)
+	{
+		std::sort(intercepts.begin(), intercepts.end(), lesserfrac); // [DVR] Can start from [intercept_index] instead of begin()??
+	}
+	intercepts.end()->done = true;
 }
 
 //===========================================================================
@@ -1656,7 +1666,12 @@ int FPathTraverse::PortalRelocate(intercept_t *in, int flags, DVector3 *optpos)
 		P_TranslatePortalZ(in->d.line, optpos->Z);
 	}
 	line_t *saved = in->d.line;	// this gets overwritten by the init call.
-	intercepts.Resize(intercept_index);
+	if (intercept_index > 0) intercepts.Resize(--intercept_index);
+	if (flags & PT_DELTA)
+	{
+		endx -= hitx;
+		endy -= hity;
+	}
 	init(hitx, hity, endx, endy, flags, in->frac + EQUAL_EPSILON);
 	return saved->getPortal()->mType == PORTT_LINKED? 1:-1;
 }
@@ -1665,9 +1680,9 @@ void FPathTraverse::PortalRelocate(const DVector2 &displacement, int flags, doub
 {
 	double hitx = trace.x + displacement.X;
 	double hity = trace.y + displacement.Y;
-	double endx = hitx + trace.dx;
-	double endy = hity + trace.dy;
-	intercepts.Resize(intercept_index);
+	double endx = (flags & PT_DELTA ? trace.dx : hitx + trace.dx);
+	double endy = (flags & PT_DELTA ? trace.dy : hity + trace.dy);
+	if(intercept_index > 0) intercepts.Resize(--intercept_index);
 	init(hitx, hity, endx, endy, flags, hitfrac);
 }
 
@@ -1679,7 +1694,8 @@ void FPathTraverse::PortalRelocate(const DVector2 &displacement, int flags, doub
 
 FPathTraverse::~FPathTraverse()
 {
-	intercepts.Resize(intercept_index);
+	// intercepts.Resize(intercept_index);
+	intercepts.Clear();
 }
 
 
--- a/src/playsim/p_maputl.cpp
+++ b/src/playsim/p_maputl.cpp
@@ -1601,14 +1601,11 @@ void FPathTraverse::init(double x1, double y1, double x2, double y2, int flags, 
 		}
 	}
 
-	struct
-	{
-		bool operator()(intercept_t a, intercept_t b) const { return a.frac < b.frac; }
-	} lesserfrac;
 	intercept_count = intercepts.Size();
 	if (intercept_count > intercept_index)
 	{
-		std::sort(intercepts.begin(), intercepts.end(), lesserfrac); // [DVR] Can start from [intercept_index] instead of begin()??
+		std::sort(intercepts.begin(), intercepts.end(), // [DVR] Can start from [intercept_index] instead of begin()??
+				  [](const intercept_t &a, const intercept_t &b) { return a.frac < b.frac; });
 	}
 	intercepts.end()->done = true;
 }
--- a/src/playsim/p_maputl.cpp
+++ b/src/playsim/p_maputl.cpp
@@ -1630,14 +1630,11 @@ void FPathTraverse::init(double x1, double y1, double x2, double y2, int flags, 
 		}
 	}
 
-	struct
-	{
-		bool operator()(intercept_t a, intercept_t b) const { return a.frac < b.frac; }
-	} lesserfrac;
 	intercept_count = intercepts.Size();
 	if (intercept_count > intercept_index)
 	{
-		std::sort(intercepts.begin(), intercepts.end(), lesserfrac); // [DVR] Can start from [intercept_index] instead of begin()??
+		std::sort(intercepts.begin(), intercepts.end(), // [DVR] Can start from [intercept_index] instead of begin()??
+				  [](const intercept_t &a, const intercept_t &b) { return a.frac < b.frac; });
 	}
 	intercepts.end()->done = true;
 }
--- a/src/playsim/p_maputl.cpp
+++ b/src/playsim/p_maputl.cpp
@@ -1370,20 +1370,19 @@ void FPathTraverse::AddThingIntercepts (int bx, int by, FBlockThingsIterator &it
 intercept_t *FPathTraverse::Next()
 {
 	intercept_t *in = NULL;
-
 	double dist = FLT_MAX;
-	for (unsigned scanpos = intercept_index; scanpos < intercepts.Size (); scanpos++)
-	{
-		intercept_t *scan = &intercepts[scanpos];
-		if (scan->frac < dist && !scan->done)
-		{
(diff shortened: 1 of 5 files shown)


TehRealSalt:
> Keeping this a draft until after 4.13.3 releases. UZD can't afford new playsim bugs in a debut release version.

If it's not draft for any other reasons, then fwiw we're just cherry-picking on top of the previous GZDoom release for 4.13.3, so there shouldn't be any risk of this being in it accidentally.

dileepvr:
Okay. I'll remove it from draft then. I don't think I have much to add right now.
Slightly updated test pk3: 
[tracer_visualizer_v2.zip](https://github.com/user-attachments/files/23199403/tracer_visualizer_v2.zip)

MajorCooke:
Does this also address the issue with aiming straight up/down? Just curious.

MajorCooke:
I can't thank you enough for doing this btw, I was having trouble with traces doing things like not traversing through portals for some ungodly reason. Hopefully this means I can remove all the hacks around having to check through all portals to make sure it offsets properly...

RicardoLuis0:
uncludging this engine one function at a time 🙏

src/playsim/p_maputl.cpp
```diff
@@ -1630,6 +1629,17 @@ void FPathTraverse::init(double x1, double y1, double x2, double y2, int flags,
...
 	}
+
+	struct
+	{
+		bool operator()(intercept_t a, intercept_t b) const { return a.frac < b.frac; }
+	} lesserfrac;
+	intercept_count = intercepts.Size();
+	if (intercept_count > intercept_index)
+	{
+		std::sort(intercepts.begin(), intercepts.end(), lesserfrac); // [DVR] Can start from [intercept_index] instead of begin()??
```
kcat: A lambda would be cleaner:

```c++
static constexpr auto lesserfrac = [](const intercept_t &a, const intercept_t &b) { return a.frac < b.frac; };
std::sort(intercepts.begin(), intercepts.end(), lesserfrac); // [DVR] Can start from [intercept_index] instead of begin()??
```

Or with C++20 ranges, projections make it even more cleaner:

```c++
std::ranges::sort(intercepts, std::less{}, &intercept_t::frac);  // [DVR] Can start from [intercept_index] instead of begin()??
// or 
std::ranges::sort(intercepts | std::views::drop(intercept_index), std::less{}, &intercept_t::frac);
```

MajorCooke: > Or with C++20 ranges, projections make it even more cleaner:

The project doesn't support C++ 20 yet, IIRC.

kcat: Not yet, but potentially soon: #144

dileepvr: I am not skilled enough to make the call on this.
Cleaner as in aesthetically? Or does it help the compiler better?

kcat: It's functionally the same, but the "more correct" way to do something like this. When you define a lambda, what really happens is the compiler generates a unique struct with member variables matching what's specified in the capture list (what's between the square brackets, `[]`, which is nothing in this case), and a call operator (`operator()`) taking the specified parameters. So

```c++
static constexpr auto lesserfrac = [](const intercept_t &a, const intercept_t &b) { return a.frac < b.frac; };
```

is equivalent to

```c++
struct some_internally_generated_name {
    auto operator()(const intercept_t &a, const intercept_t &b) { return a.frac < b.frac; }
};
static constexpr auto lesserfrac = some_internally_generated_name();
```

(a capture-less lambda like this also auto-generates a conversion operator to a C-style function pointer that calls the same method, though that's not really relevant here). The lamda is less overall typing, no risk of accidentally using the same struct names for different callables, and is clearer to the reader what's going on. The lambda can also be specified at the call site, like this:

```c++
std::sort(intercepts.begin(), intercepts.end(),
    [](const intercept_t &a, const intercept_t &b) { return a.frac < b.frac; });
```

Whether or not that's better than having the separate `lesserfrac` object that's passed to the algorithm is personal preference (especially for a simple one like this, the compiler will have no issue at all inlining the function and optimizing the extraneous objects away). Some may like avoiding extraneous explicit variables, others may prefer breaking up longer multi-line statements and giving names to objects, but the lambda form is clearer either way compared to an explicit struct.

TehRealSalt (dismissed):
This removes the need for some hacks I've put in some of my own mods, very cool :+1:

dileepvr:
Thanks. Shall I commit that lambda function?

Boondorl:
Yeah, I'd use a lambda here personally. Just a lot more readable

src/playsim/p_maputl.cpp
dileepvr: Okay done!

MajorCooke:
> Crossing3DWater only triggers when top surface of the liquid is intercepted. I assume that this is intended behavior.

...No wonder why it was failing to work properly whenever I was shooting through water floors! Can this be altered to affect sides and bottom too please?

RicardoLuis0:
> > Crossing3DWater only triggers when top surface of the liquid is intercepted. I assume that this is intended behavior.
> 
> ...No wonder why it was failing to work properly whenever I was shooting through water floors! Can this be altered to affect sides and bottom too please?

maybe with a flag? i can see reasons for wanting to only check 3d water top (ex. splashes)

MajorCooke:
Fine by me, better than nothing.

dileepvr:
Let's think about how that would work. Right now, there is no callback triggered for crossing water. You read that field when some other event triggers a callback, or after the trace hits end of distance. There is only one field for `Crossed3DWater` (and one `CrossedWater` field for Boom-style deep water, but let's ignore that, since there the top surface is the only possible crossing?).

Using 3D floors for water, you can have floating blocks of liquid with side walls and bottom surface exposed. You can make aquariums with windows, upside-down pools on the ceiling, and have the trace pass through multiple liquid 3Dfloors on the same or different sectors. All novel uses that are not covered. Right now, only the first top surface crossed gets stored (presumably, for splash purposes).

What should populate in `TraceResults`?
So a new `TRACE_Report3DWater` flag should:

1. trigger a callback, populate `Crossed3DWater` and `Crossed3DWaterPos`, 
2. populate `HitType` with `TRACE_HitFloor` (for top), `TRACE_HitCeiling` (for bottom), or `TRACE_HitWall` (for side wall). 
3. It should also populate `ffloor` and the rest of the fields. 
4. And finally, if some map element (wall, floor, ceiling) is flush with those 3Dfloor faces, it triggers a separate callback afterwards for that surface?

MajorCooke:
I think it should also still follow the rules of TRACE_HitFloor/Wall/Ceiling, but indeed populate the Crossed3D stuff. If it doesn't hit a 3D floor at all, then the 3D stuff is nulled out.

dileepvr:
Added a `TRACE_3DLiquidCallback` flag to run callback function when a 3DFloor liquid-surface crossing occurs. 

`HitType` will be `TRACE_HitFloor` for top crossing, `TRACE_HitCeiling` for bottom crossing, and `TRACE_HitWall` for side-wall crossing. `Crossed3DWater` and `Crossed3DWaterPos` will get populated. ~~It is left to the modder to calculate if the 3D-liquid volume is being entered or exited.~~ Added a `bool entering3DLiquid` field to tell whether entering or exiting 3D-liquid volume when callback is triggered from `TRACE_3DLiquidCallback` flag.
The return from callback will be ignored and assumed to be `TRACE_Continue`, just like for `TRACE_ReportPortals`.
The callback is not triggered if the 3D-liquid volume crossing surface is flush with map geometry. So sitting inside a swimming pool and pointing a trace at the floor will not be reported as a 3D-liquid crossing (unless the pool bottom surface is magically floating in mid-air above the actual sector floor).

The above text will be added to the first comment for summary.

Behavior reverts to old style if `TRACE_3DLiquidCallback` is not present, where only the first top-surface crossing is used to populate the `Crossed3DWater` and `Crossed3DWaterPos` fields and the callback function is not triggered.

Test pk3 with this new feature (comment out `TRACE_3DLiquidCallback` flag for old behavior): 
[tracer_visualizer_v3.zip](https://github.com/user-attachments/files/23588290/tracer_visualizer_v3.zip)

MajorCooke:
This makes me wonder when hitting a 3D floor, if it also reports the inverse (floor for ceiling, ceiling for floor)?

dileepvr:
Well, if you want to use the control sector's flats, then it will seem like the inverse. But if you want to use `ffloor` flats, then the `topplane` is the `secplane` for a floor (with air above), while the `bottomplane` is the `secplane` for a ceiling (with air below).

the-phinet:
﻿
﻿

dileepvr:
I was working through some tests and I began to wonder, is passing the `TRACE_ReporPortals` flag supposed to **NOT** obey the return value of the `TraceCallback` function that gets called? Because that seems to be the default. It always treats it like `TRACE_Skip` is returned.

RicardoLuis0:
> I was working through some tests and I began to wonder, is passing the `TRACE_ReporPortals` flag supposed to **NOT** obey the return value of the `TraceCallback` function that gets called? Because that seems to be the default. It always treats it like `TRACE_Skip` is returned.

it might need to ignore TRACE_Continue if traversing the portal irreversibly changes the state of the tracer (i don't know if it does), but it should be safe to allow TRACE_Stop/TRACE_Abort, though, might need to be behind a flag since it does change behavior

dileepvr:
Okay. I'll leave the behavior alone for now. I can make that a separate PR if someone asks for it. I've updated the example pk3/zip in the first comment to show a usage of `TRACE_ReportPortals`.

src/playsim/p_maputl.cpp
```diff
@@ -1370,20 +1370,19 @@ void FPathTraverse::AddThingIntercepts (int bx, int by, FBlockThingsIterator &it
...
+		dist = in->frac;
+	}
+	else
+	{
+		return NULL;
 	}
-
 	if (dist > 1. || in == NULL) return NULL;	// checked everything in range
-	in->done = true;
+	// in->done = true;
```
RicardoLuis0: just one last thing, this should be cleaned up and it'll be good to merge

src/playsim/p_maputl.cpp
```diff
@@ -1650,7 +1662,8 @@ void FPathTraverse::PortalRelocate(const DVector2 &displacement, int flags, doub
 
 FPathTraverse::~FPathTraverse()
 {
-	intercepts.Resize(intercept_index);
+	// intercepts.Resize(intercept_index);
```
RicardoLuis0: and this too

RicardoLuis0 (changes requested):

src/playsim/p_maputl.cpp
dileepvr: Done :)

RicardoLuis0 (approved):

PR #287, opened 2025-10-29
Fix null access in DoDim

TehRealSalt:
`style` can be null here, and the equivalence operators for FRenderStyle doesn't handle null.

I'm not sure if this is the correct way to fix this, I'm pretty unfamiliar with this code. Could possibly be addressed by changing how the operators work, but this seems like the only place that this situation can happen anyways.

### DIFF
--- a/src/common/2d/v_draw.cpp
+++ b/src/common/2d/v_draw.cpp
@@ -1715,7 +1715,7 @@ DEFINE_ACTION_FUNCTION(FCanvas, Clear)
 
 void DoDim(F2DDrawer *drawer, PalEntry color, float amount, int x1, int y1, int w, int h, FRenderStyle *style)
 {
-	if (amount < 0 || (amount == 0.f && *style != LegacyRenderStyles[STYLE_Source]))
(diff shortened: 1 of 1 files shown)


src/common/2d/v_draw.cpp
```diff
@@ -1715,7 +1715,7 @@ DEFINE_ACTION_FUNCTION(FCanvas, Clear)
 
 void DoDim(F2DDrawer *drawer, PalEntry color, float amount, int x1, int y1, int w, int h, FRenderStyle *style)
 {
-	if (amount < 0 || (amount == 0.f && *style != LegacyRenderStyles[STYLE_Source]))
+	if (amount < 0 || (amount == 0.f && style && *style != LegacyRenderStyles[STYLE_Source]))
```
RicardoLuis0: probably best to have `(!style || *style != LegacyRenderStyles[STYLE_Source])` instead of `style && *style != LegacyRenderStyles[STYLE_Source]`

TehRealSalt: Good point, amended in 83af3f0f9586e8fffd4ed71114b356cf6da92630

Boondorl (approved):

PR #288, opened 2025-10-29
Made BobWeapon functions server-side

Boondorl:
Sit down around the campfire and get your marshmellows out because this one requires a novel to explain what the issue is and what's being fixed. Once upon a time, P_BobWeapon was located in A_WeaponReady, specifically its bob handling. This was how the original Doom did it, but it presented a huge issue: the Chainsaw. While the functionality worked perfectly fine for weapons that called the function every tick, the Chainsaw only called it every 4 ticks, giving it the iconic choppy swaying. On a lesser note this also meant it didn't have interpolation, so a decision was made to move it to the renderer.

This worked fine and fixed both the above issues because the tic fraction could be passed, but then the large ZScript export rolled around for 2.3. It was decided it would be exported as a function on the player itself as a virtual, allowing it to be overridden. This was fine until 2.4 when the scoping system was added. This set a separation of logic between the ui and play scopes, limiting how these two accessed each other in order to make designing for multiplayer much simpler. One problem, however, is that BobWeapon was never changed to ui scope and was kept as play scoped (perhaps for backwards compatibility?). This makes it the only play scoped function in the engine to be called from the renderer, breaking this principle.

To make things worse, it also breaks how the rollback works. The rollback code relies on snapshotting the player's state after the server logic is finished but before the renderer is ran. This fundamentally means that it could not work with this solution as BobWeapon is always called after this snapshot: any changes made to fields in it will never be saved. For the default logic that wasn't a big deal, but mods like Lithium use custom fields to further tune the bobbing. Future plans for the rollback include ignoring ui scoped fields when backing up, but this means BobWeapon is fundamentally impossible to solve in its current state since the fields it modifies must necessarily be play scoped.

With the addition of client-side Actors, I wanted to keep their behavior consistent between singleplayer and multiplayer, so I enabled the rollback in singleplayer as well. That unfortunately made this problem start occurring in singleplayer games. After a lot of deliberating, I decided the best solution here was to move BobWeapon (and its other play scoped parts) over to the server logic. To help prevent desyncs, these are now ran for every player who calls P_PlayerThink when doing the server loop, modifying it at the end to best try and preserve how the renderer would call it afterwards. The tic fraction passed is now always 1 and instead the engine itself will handle interpolating from previous values. This manages to work as a 2-for-1 as now the functionality is synchronized _and_ it's applied before the rollback, allowing it to properly iterate.

Currently it relies on a bit of a gross render hack to determine which function should be called, but I couldn't see a better way of handling this. This will require a lot of testing and @RicardoLuis0 has future plans to allow the function to be returned to the renderer by safely making the function ui scoped, but this PR will always be necessary for backwards compatibility with mods that exist with the current play scoped version.

### DIFF
--- a/src/playsim/p_pspr.cpp
+++ b/src/playsim/p_pspr.cpp
@@ -640,58 +640,101 @@ void P_BringUpWeapon (player_t *player)
 //
 // [XA] Added new bob styles and exposed bob properties. Thanks, Ryan Cordell!
 // [SP] Added new user option for bob speed
+// 
+// Previously this was called from the renderer but accidentally left in the
+// play scope, making it possible to modify the world from it and breaking
+// the prediction. It's now been moved to the player thinking so it can be
+// called across all players to help prevent desyncs, but requires a bit of a
+// gross render hack to determine which one to call. Unfortunately there isn't
+// a good way to fix this, so for now this will have to do.
 //
 //============================================================================
 
-void P_BobWeapon (player_t *player, float *x, float *y, double ticfrac)
+EPSPBobType BobType = PSPB_None;
+FWeaponBobInfo BobInfo = {}, PrevBobInfo = {};
+
+void P_BobWeapon(player_t* player)
 {
+	const bool isConsolePlayer = player == &players[consoleplayer];
 	IFVIRTUALPTRNAME(player->mo, NAME_PlayerPawn, BobWeapon)
 	{
-		DVector2 result = CallVM<DVector2>(func, player->mo, ticfrac);
-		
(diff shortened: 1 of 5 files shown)


Boondorl:
@Macil If you're willing to test this with PB that would be great. I think this should fix both the sway problem and desyncs (but no guarantee on that last one)

Macil:
I just tried it and it fixes the weapon sway issue in PB.

TehRealSalt (approved):
Didn't see any regressions running with my own mods w/ BobWeapon replacements in multiplayer, :+1:

PR #293, opened 2025-10-29
Updated vanilla transparency handling

Boondorl:
Previously there were two auto options decided on whether or not there was dehacked or DECORATE present. This has been reworked so that only dehacked Actors and Actors in the core ZScript lump are now considered for it as these are the only safe options. All other Actors should default to transparency since this is likely what they were designed for given it was the existing default and so many Actors in the pk3 use it.

### DIFF
--- a/src/d_main.cpp
+++ b/src/d_main.cpp
@@ -1124,7 +1124,6 @@ void D_Display ()
 	cycles.Reset();
 	cycles.Clock();
 
-	r_UseVanillaTransparency = UseVanillaTransparency(); // [SP] Cache UseVanillaTransparency() call
 	r_renderercaps = GetCaps(); // [SP] Get the current capabilities of the renderer
 
 	if (players[consoleplayer].camera == NULL)
@@ -3764,9 +3763,6 @@ static int D_InitGame(const FIWADInfo* iwad_info, std::vector<std::string>& allw
 			return 0;
 		}
 	}
-
-	// [SP] Force vanilla transparency auto-detection to re-detect our game lumps now
-	UpdateVanillaTransparency();
 
 	// [RH] Lock any cvars that should be locked now that we're
 	// about to begin the game.
--- a/src/gamedata/d_dehacked.cpp
+++ b/src/gamedata/d_dehacked.cpp
@@ -67,6 +67,7 @@
 #include "vmbuilder.h"
 
 extern TArray<PalEntry> TranslationColors;
+extern TMap<FName, bool> AutoTrans;
 
 void JitDumpLog(FILE *file, VMScriptFunction *func);
 
@@ -141,6 +142,7 @@ static PClassActor* FindInfoName(int index, bool mustexist = false)
 			cls = static_cast<PClassActor*>(RUNTIME_CLASS(AActor)->CreateDerivedClass(name.GetChars(), (unsigned)sizeof(AActor)));
 			NewClassType(cls, -1);	// This needs a VM type to work as intended.
 			cls->InitializeDefaults();
+			AutoTrans[cls->TypeName] = true;
 			PClassActor::AllActorClasses.Push(cls);
 		}
 		if (cls)
@@ -3806,6 +3808,7 @@ void FinishDehPatch ()
 			if (newlycreated)
 			{
 				subclass->InitializeDefaults();
+				AutoTrans[subclass->TypeName] = true;
 				PClassActor::AllActorClasses.Push(subclass);
 			}
 		} 
--- a/src/r_data/r_vanillatrans.cpp
+++ b/src/r_data/r_vanillatrans.cpp
@@ -45,7 +45,8 @@
 #endif
 
 bool r_UseVanillaTransparency;
-CVAR (Int, r_vanillatrans, 0, CVAR_ARCHIVE)
+CVAR(Int, r_vanillatrans, 0, CVAR_ARCHIVE)
+TMap<FName, bool> AutoTrans = {};
 
 namespace
 {
--- a/src/r_data/r_vanillatrans.h
+++ b/src/r_data/r_vanillatrans.h
@@ -36,3 +36,5 @@
 void UpdateVanillaTransparency();
 bool UseVanillaTransparency();
 extern bool r_UseVanillaTransparency;
+extern TMap<FName, bool> AutoTrans;
+EXTERN_CVAR(Int, r_vanillatrans)
--- a/src/rendering/hwrenderer/scene/hw_sprites.cpp
+++ b/src/rendering/hwrenderer/scene/hw_sprites.cpp
@@ -1331,11 +1331,11 @@ void HWSprite::Process(HWDrawInfo *di, AActor* thing, sector_t * sector, area_t 
 	{
 		trans = 1.f;
 	}
-	if (r_UseVanillaTransparency)
+	if (r_vanillatrans && (thing->renderflags & RF_ZDOOMTRANS))
 	{
 		// [SP] "canonical transparency" - with the flip of a CVar, disable transparency for Doom objects,
 		//   and disable 'additive' translucency for certain objects from other games.
-		if (thing->renderflags & RF_ZDOOMTRANS)
+		if (r_vanillatrans == 1 || AutoTrans.CheckKey(thing->GetClass()->TypeName) != nullptr)
 		{
 			trans = 1.f;
 			RenderStyle.BlendOp = STYLEOP_Add;
--- a/src/rendering/swrenderer/things/r_sprite.cpp
+++ b/src/rendering/swrenderer/things/r_sprite.cpp
@@ -167,9 +167,9 @@ namespace swrenderer
 		if (thing->flags5 & MF5_BRIGHT)
 			vis->renderflags |= RF_FULLBRIGHT;
 		vis->RenderStyle = thing->RenderStyle;
-		if (r_UseVanillaTransparency)
+		if (r_vanillatrans && (thing->renderflags & RF_ZDOOMTRANS))
 		{
-			if (thing->renderflags & RF_ZDOOMTRANS)
+			if (r_vanillatrans == 1 || AutoTrans.CheckKey(thing->GetClass()->TypeName) != nullptr)
 				vis->RenderStyle = LegacyRenderStyles[STYLE_Normal];
 		}
 		vis->FillColor = thing->fillcolor;
--- a/src/scripting/thingdef.cpp
+++ b/src/scripting/thingdef.cpp
@@ -417,11 +417,14 @@ void ParseAllDecorate();
 void SynthesizeFlagFields();
 void SetDoomCompileEnvironment();
 
+extern TMap<FName, bool> AutoTrans;
+
 void ParseScripts()
 {
 	int lump, lastlump = 0;
 	FScriptPosition::ResetErrorCounter();
 
+	bool firstLump = true;
 	while ((lump = fileSystem.FindLump("ZSCRIPT", &lastlump)) != -1)
 	{
 		ZCCParseState state;
@@ -442,6 +445,18 @@ void ParseScripts()
 			Printf(TEXTCOLOR_ORANGE "%d warnings while compiling %s\n", FScriptPosition::WarnCounter, fileSystem.GetFileFullPath(lump).c_str());
 		}
 
+		if (firstLump)
+		{
+			// Only the core Actors defined in the engine should ever auto to non-transparent as too many things have likely
+			// relied on the forced transparency for their aesthetic. Dehacked Actors should also auto to it, but that's
+			// handled elsewhere when those classes are created.
+			for (auto cls : PClass::AllClasses)
+			{
+				if (cls->IsDescendantOf(NAME_Actor) && (GetDefaultByType(cls)->renderflags & RF_ZDOOMTRANS))
+					AutoTrans[cls->TypeName] = true;
+			}
+		}
+		firstLump = false;
 	}
 }
 
--- a/wadsrc/static/menudef.txt
+++ b/wadsrc/static/menudef.txt
@@ -949,8 +949,7 @@ OptionValue VanillaTrans
 {
 	0.0, "$OPTVAL_VTFZDOOM"
 	1.0, "$OPTVAL_VTFVANILLA"
-	2.0, "$OPTVAL_VTAZDOOM"
-	3.0, "$OPTVAL_VTAVANILLA"
+	2.0, "$OPTVAL_AUTO"
 }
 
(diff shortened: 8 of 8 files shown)


kcat:
> Previously there were two auto options decided on whether or not there was dehacked or DECORATE present.

IIRC, one was to replicate what normal ZDoom applied transparency to (since like many source ports, it would add effects to the base actors when the effects became possible but wasn't (easily) moddable, which some people liked, and it stuck), and the other was for the vanilla game, both while allowing mods to work as intended. I don't think these should be combined since the distinction between what ZDoom made transparent and what Vanilla made transparent was purposely separated. At least not without some significant community discussion.

kcat:
> This is part of a further effort to bring the initial experience more in-line with Doom's original vision out of the box

I think this is also something to bring into question. ZDoom hasn't been about replicating Doom's "original vision". It's always attempted to make an enhanced experience, both out-of-the-box and with mods. Like Marisa said in another PR:

> Doom didn't have transparency effects, so when they were added, there was the question of what to add them to to enhance the experience. What would id likely have made transparent had they had that capability?

Undoing these as a default I think does a disservice to the goal of the ZDoom family of ports, and risks entrenching the engine in a purist mindset where the default experience is just pure vanilla, when there are already other engines for that. Having to turn the enhancements on (especially when they're separate options spread all over various menus) to get "the real (U)ZDoom experience" isn't great and won't help the engine stand apart.

mc776:
While I agree in principle, I think this particular case could go either way - a lot of source ports already do translucent stuff by default (I just remarked on another issue how I'd immediately turned off translucency in another sourceport when I saw it), but few of them do things like dynamic lights, brightmaps or constant freelook right out of the box. And I suspect most people using GZDoom in recent years have already come in with some mod or other in mind...

(That said, additive fireballs is still in my view one of the best base enhancements that no other port seems to ever get right.)

RicardoLuis0:
i'd personally like for fireballs/explosions/etc to be additive (and fullbright) by default, not opaque, but that's my own personal opinion (but a lot of the other stuff, ex. puffs, i would still prefer as opaque)

Boondorl:
> > This is part of a further effort to bring the initial experience more in-line with Doom's original vision out of the box
> 
> I think this is also something to bring into question. ZDoom hasn't been about replicating Doom's "original vision". It's always attempted to make an enhanced experience, both out-of-the-box and with mods. Like Marisa said in another PR:
>
> > Doom didn't have transparency effects, so when they were added, there was the question of what to add them to to enhance the experience. What would id likely have made transparent had they had that capability?
> 
> Undoing these as a default I think does a disservice to the goal of the ZDoom family of ports, and risks entrenching the engine in a purist mindset where the default experience is just pure vanilla, when there are already other engines for that. Having to turn the enhancements on (especially when they're separate options spread all over various menus) to get "the real (U)ZDoom experience" isn't great and won't help the engine stand apart.

The problem here is you're arguing with the original intent of the artist and this could easily be fired right back in that direction about the original game. I don't think it's the ports domain to be dictating what the game "should" look like, and features being opt-out isn't a great philosophy, especially when you look at our current menus. I know there's a whole mountain of other things to bring up here regarding that, but those are things I'd like to carefully fix over time by considering what they add to the experience. When people boot up Doom, they do so to play Doom. We have the most powerful modding ecosystem in the entire port scene, I really don't think random graphical effects being on by default are adding a ton here nowadays. After having played with it on for a bit, I agree with Nash's take that transparency really removes a lot of the impact things have. That being said, I'm not too determined to change it to auto, I'd be fine with changing it back to translucent.

> >  Previously there were two auto options decided on whether or not there was dehacked or DECORATE present.
>
> IIRC, one was to replicate what normal ZDoom applied transparency to (since like many source ports, it would add effects to the base actors when the effects became possible but wasn't (easily) moddable, which some people liked, and it stuck), and the other was for the vanilla game, both while allowing mods to work as intended. I don't think these should be combined since the distinction between what ZDoom made transparent and what Vanilla made transparent was purposely separated. At least not without some significant community discussion.

The change to auto is because the heuristic for it was not very good as having any ZScript mods loaded overrided everything else. I think the original intent was to enable it with any dehacked (which can't support transparency for Actors) and disable it when any mods were loaded via DECORATE, but to me this doesn't come across as a very realistic solution. Some mods have nothing to do with these Actors and the result is that 99% of the time it'll just fall back on whatever preference you set anyway except in niche cases, in which you'd be better off just forcing your preference for consistency. Another problem is the ZDOOMTRANS flag is inherited, likely without people realizing this. Since ZDoom defaulted to transparency, I can't imagine a ton of replacements were made with this in mind but did use the translucent/additive render styles. The new auto option limits this purely to vanilla and dehacked Actors since that seemed to have been its original intent from my own reading, but it's hard to tell what it was trying to do with how it was checking lumps.

kcat:
> The problem here is you're arguing with the original intent of the artist

Not really. There's an implicit acknowledgement that this is deviating from the original game, as we're considering what they would have done if things were different (if they had the ability and time to add these things in, which they didn't, so it's inherently conjecture), with the goal of providing an enhanced experience.

> I don't think it's the ports domain to be dictating what the game "should" look like

I disagree. The port's domain is to be as the port wants to be. While there can be reasons to follow what the original game did in some aspects, it's not some sacrosanct state of being that ought to be adhered to by default. It's all personal preference how any one person wants the game to be and how to look, and ports like ZDoom and its derivatives (and even other non-ZDoom-based ports) have opted for a more enhanced out-of-the-box experience compared to the vanilla game. That's what many people came to and like ZDoom for, and reverting that over the idea that the default experience must be as intended by vanilla is counter to that goal. Especially with all the (G)ZDoom specific mods that are built on the expectation that the engine is more modern by default, reverting that will also do a disservice to those mods.

If it's not the port's call to be different from vanilla by default, we should also change the default renderer to the original Carmack software renderer. But despite the hardware renderer being significantly different with true-color rendering, smooth shading and accurate translucency, and texture mipmapping/anisotropic filtering/etc, I doubt that idea would ever be considered. So if some things are allowed to be different, there should be a better argument than "the original intent of the artist" or "what the game should look like" for why one thing shouldn't be changed when another thing can. And yes, this can sometimes come down to "because I don't like it"; Graf did it, Marisa did it, anyone who's ever led a project where these kinds of subjective decisions must be made will likely do it (hopefully with reasons backing that preference, but it still is ultimately preference).

> When people boot up Doom, they do so to play Doom.

If people want to play Doom, they will boot up Doom. If they want to play [G|U]ZDoom, they boot up [G|U]ZDoom. If people want to play vanilla Doom in [G|U]ZDoom, the options are there to make it possible (for the most part). People don't boot up [G|U]ZDoom only ever to play completely vanilla Doom. They'd boot up Doom.exe, or Chocolate Doom, or any of the official re-releases intended to provide a more vanilla-accurate experience. The only reason to use ZDoom (or any other port) is if you want something to be different, where each port can find its niche to fill in what's different, which should be made apparent to the user.

mc776:
Seconding kcat here... my go-tos are UZ, Chocolate and Nugget because each of them offers something distinct (Chocolate being the only one truly comprehensively faithful to Doom) while I mostly boot up DSDA, Crispy and Helion (all of which offer primarily straight-up visual upgrades) just for testing stuff on different ports and otherwise aren't my standbys.

I'm not opposed to changes from what ZDoom has done, and most of the changes I'm not opposed to would probably end up making UZ more in line with other ports *by default*, but for me the concern with those changes is compatibility with content made for vanilla or other ports.

EDIT: Anyway would it be better to hive all this off into a discussion or something? I'm having trouble seeing how this relates to the specific technical issues of this PR at this point.

Boondorl:
Regardless, the auto behavior change should probably still be added. The logic for how the auto works now is:

* If any ZScript mod is loaded, always use the preference
* If any DEHACKED is loaded, always use vanilla
* If any DECORATE is loaded, always use translucent
* Otherwise, use the preference

To me this feels functionally broken? If you have any non-engine ZScript loaded for any reason, the auto option genuinely does nothing. I'm confused as to what the goal was and I can only assume this new version was what it was going for. The new auto only changes vanilla and dehacked actors to opaque since any kind of modded actor is likely too risky to change given the above stated issue with ZDOOMTRANS. With this heuristic having two options doesn't really make sense because you should just be forcing a preference if you desire one since 99% of the time that's likely what UZDoom was doing anyway.

Boondorl:
I'm just gonna default this back to ZDoom translucency because frankly I care more about the auto change than I do the default.

kcat:
> Regardless, the auto behavior change should probably still be added. The logic for how the auto works now is:
> 
>     * If any ZScript mod is loaded, always use the preference
>     * If any DEHACKED is loaded, always use vanilla
>     * If any DECORATE is loaded, always use translucent
>     * Otherwise, use the preference
> 
> To me this feels functionally broken?

Hmm. It'd be nice if we could ask Graf what his intention was (I'm pretty sure this option is from GZDoom; if I'm wrong and it was from ZDoom, maybe Marisa can chime in on what's intended and the idea behind it).

> If you have any non-engine ZScript loaded for any reason, the auto option genuinely does nothing. I'm confused as to what the goal was and I can only assume this new version was what it was going for.

If I had to guess, perhaps the idea is:

* if there's a mod using ZScript, then its a more recent (G)ZDoom mod where the two auto options have been available, so its best to follow what the user prefers, otherwise:
* if there's a DECORATE lump, it's possibly from the time period where there weren't separate auto-vanilla and auto-zdoom options (so either ZDoom style translucency was expected or the user would have forced one or the other), otherwise:
* if there's a DEHACKED lump, it's likely a mod for either vanilla or is generally more port-agnostic, which would expect more vanilla compatibility. Otherwise,
* no way to know, use what the user prefers.

It's not a perfect solution, as there's no way to truly know what any given mod expects (you can have a mod that doesn't use any of this, or an older ZDoom mod that only uses DEHACKED but expects ZDoom translucency, or a more recent ZDoom mod that only uses DECORATE and doesn't require objects to use ZDoom translucency, or a vanilla-based mod expecting vanilla translucency used at the same time as a ZDoom mod with ZScript that doesn't touch any affected entity). Since a mod only needs to change the graphics of an object for the (lack of) translucency to become a problem, there isn't an easy way to know whether a given mod or graphic really needs ZDoom- or vanilla-style translucent objects as a base if they don't explicitly specify it in the actor definition. But it lets users select their preference for object translucency being like vanilla or ZDoom, without always forcing it (with the option to force it still being there, for cases where it's necessary).

Boondorl:
All the comments for it were labeled with [SP] and I'm not quite sure who that is.

In my case what I'd really like is an option to set vanilla things only to be opaque without messing with mods which, in all honestly, likely all rely on being translucent. I doubt a ton of people knew what this option was or how it even worked. To me it always feels safest to assume translucent, but if someone wants a different look, that should be what the forced options are for. It's just funny to me that you could load up an old map with a dehacked file to change its name and suddenly everything is forced to opaque. It's a very hard option to read what it's doing.

nashmuhandes:
> All the comments for it were labeled with [SP] and I'm not quite sure who that is.

[SP] would be @madame-rachelle

madame-rachelle:
The original intent of the auto-translucency algorithm was to try and automatically choose a transparency style based on a mod's apparent age and format, using the presence of certain lump types such as Dehacked, Decorate, or ZScript as heuristics - it was an attempt to preserve visual consistency across mods of different eras, though the method was ultimately unreliable and intended as a workaround for the inconsistent transparency in older ZDoom mods, due to the rather short-term thinking that was involved with the decision to arbitrarily make certain actors transparent in ZDoom, while the original Vanilla Doom did not have such things.

(p.s. I don't mind if it is axed entirely, I am not sure it ever actually served its purpose)

Boondorl:
Which reminds me that I still see people using DECORATE to make mods to this day lol

mc776:
Every day I see an ancient gross DECORATE hack in a new project for something that could be done in 3 lines of ZScript I die a little. (At least nobody's using A_TakeInventory loops to count variables anymore... right?....)

I'm in favour of axing since I think it will make things easier for both the users to understand and the coders to maintain.

Boondorl:
Thinking about this more, I think there is potential to add two modes here:
* Auto (Translucent) - Try and keep things translucent unless a dehacked Actor that overrides something with ZDOOMTRANS is detected. In that case, switch all vanilla Actors to opaque to try and keep parity between them, but leave mods as translucent.
* Auto (Vanilla) - Try and keep things opaque, unless a mod with an Actor that has ZDOOMTRANS is detected, in which case set it all to translucent to keep visual parity.

I think the current naming conventions don't make a ton of sense with these heuristics, but I _do_ think they capture the spirit of what these auto modes were going for. I think this current mode (always have vanilla actors as opaque) could be added as its own mode (editing the LANGUAGE files might be awkward atm since I don't know the status on our new system yet, but it should be doable with a simple edit).

MajorCooke:
Yes, that would be perfect. Also I hope Auto (Vanilla) sets the renderstyles to normal.

kcat:
> I think this current mode (always have vanilla actors as opaque) could be added as its own mode

Isn't that what "Vanilla (Forced)" is? To always assume vanilla-style translucency[1], unless explicitly modded otherwise? Or does that force it even when the modded actor definition says to be different?

[1] Which would be nothing in Doom1/2, though Heretic and Hexen do have some things as translucent in vanilla.

mc776:
Can we say "Opaque" instead of "Vanilla"? We're getting to a point where I see a lot of people who say the latter term and end up turning out to have meant some limit-removing sourceport that does the thing in question slightly differently.

kcat:
> Can we say "Opaque" instead of "Vanilla"?

"Opaque" would be incorrect when talking about the non-Doom games, which do have translucency in their vanilla engines. If I'm playing Heretic and see the option set to force or prefer "opaque" rendering, I'd worry that it's messing with the creatures that are supposed to be translucent.

Boondorl:
> > I think this current mode (always have vanilla actors as opaque) could be added as its own mode
> 
> Isn't that what "Vanilla (Forced)" is? To always assume vanilla-style translucency[1], unless explicitly modded otherwise? Or does that force it even when the modded actor definition says to be different?
> 
> [1] Which would be nothing in Doom1/2, though Heretic and Hexen do have some things as translucent in vanilla.

In this case it always keeps the vanilla and dehacked Actors opaque while leaving modded Actors as transparent aka a mix and match system. The current system uses an "all or nothing" approach where it decides what everything marked with ZDOOMTRANS should be based on what type of mods are loaded. The big issue here is that this flag is inherited meaning a lot of people likely have it on their own Actors without even realizing it. Realistically this never should've been inheritable so people would always be forced to make the explicit choice.

the-phinet (approved):

nashmuhandes (approved):

PR #314, opened 2025-10-31
Default addrocketexplosion to true, as this is more consistent with e…

nashmuhandes:
…very other explosion actors' renderstyles. Also expose it to the menu.

This option has been completely hidden and undocumented for god-knows-how-long (decades?) and my argument for it to be enabled is that stylistically, it being enabled makes the rocket explosions' aesthetics consistent with every other explosion actor defined in uzdoom.pk3 who all have the additive render style applied.

~~Personally, I'd remove this CVar completely and move all the renderstyle scripting to the Actor itself, but as has been recently discussed, it's too late to change stuff significantly like this...~~

Image comparisons: `addrocketexplosion` off versus on.

Screenshot_Doom_20251031_172946

Screenshot_Doom_20251031_173014


mc776:
YES PLEASE FINALLY

~~also for the record agreeing with the struckthrough portion of the OP, in case this is revisited again in a few years~~

RicardoLuis0 (approved):

PR #332, opened 2025-11-02
allow explicit object type casts (for Floor/etc)

### LINKED ISSUE #199: Making exposed stock sector effects use getters instead of directly exposing the variables.
## Context
A few months ago, I made a PR for GZDoom that exposed some of the vanilla sector thinkers like doors and platforms to ZScript, this was merged by Rachael (After I first made it read only like Jay told me).

However, the way they are exposed as you can see [here](https://github.com/UZDoom/UZDoom/blob/ba2b8ae0b41f2054ee99c4ef7110d93e9860ce03/wadsrc/static/zscript/doombase.zs#L737-L985) is mostly by directly exposing the variables to ZScript as read only. This was good enough for merging the PR. But then about a week ago, [the PR was reverted in GZDoom](https://github.com/ZDoom/gzdoom/commit/0ad433f1b88cc6839e5db485b25dbc8009e241aa), with the reason being that it makes updating them later impossible, and that it'd be better to fully scriptify them instead, and _then_ expose them.

## The question
So the question is basically what to do with it, initially I didn't think of changing anything, but then @nashmuhandes suggested that I should redo it to instead use getter functions, which I do think would be a good idea. But @Boondorl disagrees. (Plus I was told to make this an issue ticket instead of just bringing it up on the Discord)

Personally I think it could be a good idea make a new PR to replace the directly exposed variables with getters. But on the other hand I'm not sure if the thinkers that were directly exposed will ever even need to be updated again to begin with, like I don't see it super likely that DDoor will ever get some big update that might be made harder by its' variables being exposed. So with that in mind, I suppose it might not be that big of a deal if it's kept as is (Otherwise I don't think Rachael would've merged it).

## Compatibility
Another potential point against changing the sector thinkers to use getters is that the current setup has actually [made its' way in a GZDoom release](https://forum.zdoom.org/viewtopic.php?t=80447), so changing it now might break any mods that already are reading that data. However, I don't think _this_ is a very good reason against changing it because:

1. This is a relatively obscure feature. I find it unlikely there's yet that many mods that even read the sector thinker variables directly. Plus, since this is a fairly esoteric feature that is most likely mainly used by people who are very deep in G/UZDoom modding, I believe it would make it more likely that whichever few mods use it (Like mine), would be fine with just pushing a new update that uses getters.
2. The exposed thinkers have only been in one recent incremental release. The other reason is that the current version of the PR has only been in one small GZDoom release so far, that has only been out for about ~6 months. So combined with its' relative obscurity, I think it makes less likely that too many mods have begun using it yet. Plus the short period the current iteration of the PR has been in a release for (And the aforementioned likelyhood of the people using it being very immersed in modding), makes me think that whichever few mods already use it, are still in active development by their creators to simply update to match the breaking change.
3. More of a minor point, but from what I can tell, a lot of people already expect some level of breaking changes with UZDoom as is. So if the sector thinkers being exposed like that IS actually a problem, then this is probably the best timeframe to make this breaking change (With all the other reasons above). And AFAIK the first UZDoom release is only a week or two away. If it is a problem after all and it slips into the first release of the new fork, then I think _that_ would be a point where making such a breaking change could be a problem, though even then, I think the relative obscurity of reading sector thinker data and the still small time frame work in favor of making such a change.

## Floor (Thinker) VS Floor (Builtin)
Besides whether or not the exposed sector thinker data should be refactored before the first release or not. There's definitely another thing I've observed that is **_absolutely_** a problem, which is that you CANNOT do something like a pointer cast with the Floor thinker, because the ZScript parser confuses it with the floor() math builtin. Which means that currently you cannot check much on the Floor thinker.

For example, this code:

```
		SectorEffect FloorD = CurSector.FloorData;
		If (FloorD Is "Floor" && Floor(FloorD).m_Direction == Ceiling.DirWait)
			Return True;
```
Will return a "Numeric value expected for parameter" error if you try to run it. However, I'm not sure how it could be fixed, maybe Floor could be exposed under a different name instead, like DFloor, to stop this parser error?

RicardoLuis0:
partially fixes #199

needs to be done as `(<TYPE>)(...)` instead of `(TYPE)(...)` due to parsing conflicts though, so:
```
SectorEffect FloorD = CurSector.FloorData;
if(FloorD is 'Floor' && (<Floor>)(FloorD).m_Direction == Ceiling.DirWait)
{
    return true;
}
```

### DIFF
--- a/src/common/scripting/frontend/zcc_exprlist.h
+++ b/src/common/scripting/frontend/zcc_exprlist.h
@@ -75,4 +75,6 @@ xx(Trinary,			'?')
 
 xx(Cast,			TK_Coerce)
 
+xx(ExplicitCast,	'(')
+
 #undef xx
--- a/src/common/scripting/frontend/ast.cpp
+++ b/src/common/scripting/frontend/ast.cpp
@@ -648,6 +648,16 @@ static void PrintExprClassCast(FLispString &out, const ZCC_TreeNode *node)
 	out.Close();
 }
 
+static void PrintExprExplicitCast(FLispString &out, const ZCC_TreeNode *node)
+{
+	ZCC_ExplicitCast *enode = (ZCC_ExplicitCast *)node;
+	assert(enode->Operation == PEX_ExplicitCast);
+	out.Open("expr-explicit-cast");
+	out.AddName(enode->ClassName);
+	PrintNodes(out, enode->Parameters, false);
+	out.Close();
+}
+
 static void PrintExprFunctionPtrCast(FLispString &out, const ZCC_TreeNode *node)
 {
 	ZCC_FunctionPtrCast *enode = (ZCC_FunctionPtrCast *)node;
@@ -1100,6 +1110,7 @@ static const NodePrinterFunc TreeNodePrinter[] =
 	PrintTwoArgIterationStmt,
 	PrintThreeArgIterationStmt,
 	PrintTypedIterationStmt,
+	PrintExprExplicitCast,
 };
 
 FString ZCC_PrintAST(const ZCC_TreeNode *root)
--- a/src/common/scripting/frontend/zcc-parse.lemon
+++ b/src/common/scripting/frontend/zcc-parse.lemon
@@ -1545,6 +1545,7 @@ primary(X) ::= primary(A) LPAREN func_expr_list(B) RPAREN. [DOT]		// Function ca
 	expr->Parameters = B;
 	X = expr;
 }
+
 primary(X) ::= LPAREN CLASS LT IDENTIFIER(A) GT RPAREN LPAREN func_expr_list(B) RPAREN. [DOT]		// class type cast
 {
 	NEW_AST_NODE(ClassCast, expr, A);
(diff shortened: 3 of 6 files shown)


src/common/scripting/frontend/zcc_exprlist.h
```diff
@@ -75,4 +75,6 @@ xx(Trinary,			'?')
 
 xx(Cast,			TK_Coerce)
 
+xx(ExplicitCast,	'(')
```
Boondorl, line 78: Should this be a named token?

RicardoLuis0: nah, that's just for error messages with expressions, duplicate things aren't a problem

RicardoLuis0: it isn't used for parsing

RicardoLuis0: or was it for binary expressions? one of the two

RicardoLuis0: either way, it wouldn't impact these casts

MrRaveYard:
Alternative? #379

I don't like the syntax here tbh.
How about something else entirely?

This:
```cpp
var as 'type'
```

```cpp
void Test()
{
    let x = self as 'PlayerPawn';   // sleek and modern, easy on the eyes
    let x = (<PlayerPawn>)(self);   // C++ template hell vibes
    let x = PlayerPawn(self);       // good-old function style cast

    let floor = self;
    let f = floor as 'floor';  // Seems clear enough
}
```

I'd mandate the `' '` regardless of the fact we'll likely never want to support some sort of dark magical way to cast into a type specified by another variable.

Boondorl:
Yep, we landed on `x as Y` too a few days ago when talking about it

inkoalawetrust:
Any progress on this? The primary point of #199 was already settled as "keep it as is" so this is this only important bit about the exposed sector thinkers left.

RicardoLuis0:
still waiting on this change https://github.com/UZDoom/UZDoom/pull/379#issuecomment-3503239060

i'll come back to this PR later to change it to `as` (and to support any type, like name/string/int/etc)

the-phinet:
﻿
﻿

PR #342, opened 2025-11-03
allow switching to modern typeface alternatives with -moderntypeface arg

River-Salmon:
I've generated bitmap fonts from more legible fonts and added a commandline argument to enable them instead of the typical display fonts. 
BigUpper, BigFont, SmallFont are using lexend. The HUD font is using MomoTrustDisplay. 
I've included the relevant OFL license file in the wadsrc. 
Currently only implemented for Doom. 
image
image


andOlga:
- Why are you including a jizzilion copies of the same "undefined" glyph here? Why is it present even in places where actual characters should be? The entire cyrillic range is just this box, for example. This will break languages.

- Why is this a CLI argument and not a CVAR? This should go to the accessibility menu, in my opinion.

RicardoLuis0:
> * Why is this a CLI argument and not a CVAR? This should go to the accessibility menu, in my opinion.

i'm sure it _will_ be added as a cvar, it's just a draft for now

River-Salmon:
This is probably being shelved in favor of more robust font rendering support after the update to c++20.

PR #344, opened 2025-11-03
Haiku fixes for UZDoom

Peppersawce:
This patch adds a couple of fixes for UZDoom on Haiku:

1) Replaces 'xdg-open' with 'open' on Haiku, this makes the "Browse game config/screenshots/saved games" options work there.

2) Adds Haiku-specific paths in "gameconfigfile.cpp"

For context, this is a part of the patches I'm currently applying to GZDoom when building it on Haiku. As such, they've been tested multiple times, work as intended, don't break anything and don't affect any other OS.

Even with these patches applied building stock UZDoom will currently fail on Haiku due to ZWidget. Syncing your ZWidget to upsteam should solve that.

### DIFF
--- a/src/common/engine/i_specialpaths.h
+++ b/src/common/engine/i_specialpaths.h
@@ -4,7 +4,11 @@
 
 #if defined(__unix__) || defined(__HAIKU__)
 FString GetUserFile (const char *path);
+const char * GetConfigPath();
+const char * GetCachePath();
+const char * GetDataPath();
 #endif
+
 FString M_GetAppDataPath(bool create);
 FString M_GetCachePath(bool create);
 FString M_GetAutoexecPath();
@@ -15,7 +19,6 @@ FString M_GetDocumentsPath();
 FString M_GetDemoPath();
 
 FString M_GetNormalizedPath(const char* path);
-
 
 #ifdef __APPLE__
 FString M_GetMacAppSupportPath(const bool create = true);
--- a/src/common/platform/posix/unix/i_specialpaths.cpp
+++ b/src/common/platform/posix/unix/i_specialpaths.cpp
@@ -123,7 +123,7 @@ FString GetUserFile (const char *file)
 
 FString M_GetAppDataPath(bool create)
 {
-	static FString path = FStringf("%s/" GAMENAMELOWERCASE, GetDataPath());
+	static FString path = FStringf("%s/games/" GAMENAMELOWERCASE, GetDataPath());
 	path = NicePath(path.GetChars());
 
 	if (create)
@@ -191,8 +191,7 @@ FString M_GetConfigPath(bool for_reading)
 
 FString M_GetDocumentsPath()
 {
-	static FString path = FStringf("%s/" GAMENAMELOWERCASE "/", GetDataPath());
-	return NicePath(path.GetChars());
+	return M_GetAppDataPath(false) + "/";
 }
 
 
--- a/src/gameconfigfile.cpp
+++ b/src/gameconfigfile.cpp
@@ -36,21 +36,23 @@
 
 #include <stdio.h>
 
-#include "gameconfigfile.h"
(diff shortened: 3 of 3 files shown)


libraries/ZWidget/CMakeLists.txt
RicardoLuis0, line 1: files in libraries/ZWidget (and libraries/ZMusic for that matter) should not be changed directly, they're updated from upstream

Peppersawce: On this note, are you planning on merging upstream ZWidget before next release? That would fix it as well.
Either way I'll remove that part of the patch.

EDIT: Done

Peppersawce:
I had a look at #353 and noticed it also changed paths so I took some inspiration out of it :)

the-phinet:
Does haiku follow xdg? If not, I can carve out an exception for it

Peppersawce:
On Haiku the XDG paths are declared as:
```
XDG_DATA_HOME = /boot/home/config/non-packaged/data
XDG_CONFIG_HOME = /boot/home/config/settings
XDG_CACHE_HOME = /boot/home/config/cache
```
In fact the standard unix paths declared around end up working fine out of the box for most stuff, gamedata and soundfonts being the exception.
Atm all the paths added with this patch are useful on the system but maybe some could be rewritten to use XDG_DATA_HOME and made more generic?

The "doomdata" paths are currently used to share game data between source ports by some packages in the repo ie. Freedoom or by the user putting the wads in "non-packaged"

Now, if you're talking about "xdg-open" since names are confusing and I might've misundestood, Haiku doesn't have "xdg-" commands as far as I know. "xdg-open" is the only instance I found so far where it mattered.

the-phinet:
Looking at your reply and your code, I realized that I entirely forgot to connect the paths defined in gameconfigfile to the ones in i_specialpaths.

I was talking about the XDG_* vars specifically. It's interesting that Haiku uses them but not xdg-open.

---

Looking at [chocolate doom](https://depot.haiku-os.org/#!/pkg/chocolate_doom/haikuports/haikuports_x86_64/3/1/1/-/1/x86_64?bcguid=bc2-OGVY) and [prboom+](https://depot.haiku-os.org/#!/pkg/prboom_plus/haikuports/haikuports_x86_64/2/6/66/-/5/x86_64?bcguid=bc2-CSFA), they both will look for wads in `$XDG_DATA_HOME/doom`, so I don't think adding `doomdata` is needed.

Once #353 gets merged, the is pr could likely be simplified by changing the DEFGETPATH fallback paths in `i_specialpaths.cpp` and SHARE_DIR in `i_system.h`

Peppersawce:
Iirc the XDG_PATHS are specified on Haiku for ease of porting and not much else.

> Looking at [chocolate doom](https://depot.haiku-os.org/#!/pkg/chocolate_doom/haikuports/haikuports_x86_64/3/1/1/-/1/x86_64?bcguid=bc2-OGVY) and [prboom+](https://depot.haiku-os.org/#!/pkg/prboom_plus/haikuports/haikuports_x86_64/2/6/66/-/5/x86_64?bcguid=bc2-CSFA), they both will look for wads in $XDG_DATA_HOME/doom, so I don't think adding doomdata is needed.

That's true for now, but I'm looking to retire some of those paths  for newer ports, makes them easier to mantain long-term.

Of course, nothing stops any extra paths the project doesn't like from being implemented as a patch from our side.
The ones with GAMENAMELOWERCASE are the important ones either way.

> Once https://github.com/UZDoom/UZDoom/pull/353 gets merged, the is pr could likely be simplified by changing the DEFGETPATH fallback paths in i_specialpaths.cpp and SHARE_DIR in i_system.h

I could drop the path changes from the patch and submit an updated patch once the new system is in place. Let me know if that's more ideal.

Peppersawce:
Accidentally closed this PR, will open a new one shortly

PR #349, opened 2025-11-04
Default gender to "neutral"

andOlga:
Hardly any reason to ship with outdated assumptions from the nineties, let's use the thing that works for everyone.


mc776:
The sight of this PR, compared to the feeling I get (regardless of whatever I'm identifying as at the time) whenever I reconfigure a fresh install of any affected Doom port and see *that*, feels like when they finally start cleaning up the Latin Quarter in earnest in From Up On Poppy Hill.

andOlga:
> The sight of this PR, compared to the feeling I get (regardless of whatever I'm identifying as at the time) whenever I reconfigure a fresh install of any affected Doom port and see _that_, feels like when they finally start cleaning up the Latin Quarter in earnest in From Up On Poppy Hill.

I have absolutely no idea what any of this means. 😅

RicardoLuis0 (approved):

mc776:
> I have absolutely no idea what any of this means. 😅

Dilapidated old boy's club sort of environment finally gets the love it needs to make it welcoming for everyone!

zhs2:
Should have called this PR "Oops! All Cyborgs!"

andOlga:
> Should have called this PR "Oops! All Cyborgs!"

Object gender (using it/its pronouns, what Quake refers to as "cyborg") is *not* what this PR defaults to. This defaults to the generic human they/them.

PR #353, opened 2025-11-04
Update Linux path handling to follow XDG spec

the-phinet:
The following are now respected and used:
- XDG_DATA_HOME (~/.local/share by default)
- XDG_CONFIG_HOME (~/.config by default)
- XDG_CACHE_HOME (~/.cache by default)

Default install location of engine data has been updated to not conflict with other ports.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,7 +132,7 @@ function( add_pk3 PK3_NAME PK3_DIR )
 	if( WIN32 )
 		set( INSTALL_PK3_PATH . CACHE STRING "Directory where zdoom.pk3 will be placed during install." )
 	else()
-		set( INSTALL_PK3_PATH share/games/doom CACHE STRING "Directory where zdoom.pk3 will be placed during install." )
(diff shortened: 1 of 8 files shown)


TehRealSalt (dismissed):
I've noticed that setting `XDG_CONFIG_HOME` can cause a "Failed to create [the path] directory: No such file or directory" error if it wasn't created in advance, but it will still create the directory in the process. This feels a bit counter-intuitive, since it will then run fine on a second attempt *because* it created the directory for you.

I probably won't block over this (could be argued that this is an existing issue, since it looks like the directory error code is mostly the same), but still would be nice to fix imo. Otherwise lgtm! :+1:

Boondorl:
Can we get another double check on this? Need this for 4.14.3

TehRealSalt:
> Can we get another double check on this? Need this for 4.14.3

On it now (got side-tracked with that crazy assert business last night lol)

TehRealSalt (approved):
Not having the folder made ahead of time works now :+1:

PR #368, opened 2025-11-05
Some Haiku fixes for UZDoom

Peppersawce:
Follow-up to #344, includes:

- Changing xdg-open to open to fix the "Browse (folder)" menu options
- Patches to paths on Haiku

With this patch "dataDir" will point at non-packaged data, "SHARE_DIR" at system-wide install data and the 3rd install option, packaged local data, is covered by the additions in "gameconfigfile.cpp".

The block in gameconfigfile.cpp is also meant to avoid adding "/usr" or ".local" entries in the config file as those paths are invalid on the system.

### DIFF
--- a/src/common/platform/posix/i_system.h
+++ b/src/common/platform/posix/i_system.h
@@ -17,7 +17,11 @@ struct WadStuff;
 struct FStartupSelectionInfo;
 
 #ifndef SHARE_DIR
+#ifdef __HAIKU__
+#define SHARE_DIR "/boot/system/data"
+#else
 #define SHARE_DIR "/usr/local/share"
+#endif
 #endif
 
 void CalculateCPUSpeed(void);
--- a/src/common/platform/posix/sdl/i_system.cpp
+++ b/src/common/platform/posix/sdl/i_system.cpp
@@ -424,7 +424,11 @@ void I_OpenShellFolder(const char* infolder)
 	{
 		if (longsavemessages)
 			Printf("Opening folder: %s\n", infolder);
-		std::system("xdg-open .");
+		#ifdef __HAIKU__
+			std::system("open .");
+		#else
+			std::system("xdg-open .");
+		#endif
 		chdir(curdir);
 	}
 	else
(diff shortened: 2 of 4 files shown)


src/common/platform/posix/i_system.h
```diff
@@ -17,8 +17,12 @@ struct WadStuff;
 struct FStartupSelectionInfo;
 
 #ifndef SHARE_DIR
+#ifdef __HAIKU__
+#define SHARE_DIR "$HOME/config/data"
```
the-phinet: Would this not be something like `/boot/system/data` ?

`/usr/local/share` is for system-wide stuff on Linux

the-phinet: If that's the case, the `shareDirChanged` stuff can probably be re-jiggered to have less haiku-specific blocks

```cpp
#if defined(__HAIKU__) || ( defined(__unix__) && !defined(__APPLE__) )
#   ifdef __HAIKU__
#       define DEFAULT_SHARE_DIR "/boot/system/data"
#   else
#       define DEFAULT_SHARE_DIR "/usr/local/share"
#   endif
    bool shareDirChanged = 0 != strcmp(SHARE_DIR, DEFAULT_SHARE_DIR);
    FString dataDir = GetDataPath();
#endif
```

```cpp
if (shareDirChanged)
{
    SetValueForKey ("Path", DEFAULT_SHARE_DIR "/games/" GAMENAMELOWERCASE, true);
    SetValueForKey ("Path", DEFAULT_SHARE_DIR "/games/doom", true);
    SetValueForKey ("Path", DEFAULT_SHARE_DIR "/doom", true);
}
```

```cpp
#ifdef DEFAULT_SHARE_DIR
#   undef DEFAULT_SHARE_DIR
#endif
```

Peppersawce: > Would this not be something like /boot/system/data ?

That would work as well. Will implement these suggestions shortly

Peppersawce: Ok, changes made. I was unsure where to place

```
#ifdef DEFAULT_SHARE_DIR
#   undef DEFAULT_SHARE_DIR
#endif
```

So I put it at the end of `FGameConfigFile::FGameConfigFile ()`, let me know if that's the right place for it.

the-phinet (approved):

TehRealSalt (approved):

PR #379, opened 2025-11-06
Resolve ZScript Floor(x) call ambiguity by determining if 'x' resolves to an object.

MrRaveYard:
```cpp
void FreeYourselfFromGrossHackMentality()
{
	let f = Floor(self); // cast
	int x = Floor(pos.x); // flop
}
```

```
Disassembly @ 0000022F02BDC0B0:

00000000: 57010000 dyncast a1,a0,0000022CE6D8CB10       ;1,0,0
00000004: 19000000 ldp     f0,a0,88                     ;0,0,0

00000008: bf000007 flop    f0,f0,7 [floor]              ;0,0,7
0000000c: 54000004 cast    d0,f0                        ;0,0,4

00000010: 65808000 ret                                  ;128,128,0
```

Memory leaks? Hopefully not.
Should be fine since there is a guard against duplicate `Resolve` calls in the codegen.

### DIFF
--- a/src/common/scripting/backend/codegen.cpp
+++ b/src/common/scripting/backend/codegen.cpp
@@ -8572,13 +8572,30 @@ FxExpression *FxFunctionCall::Resolve(FCompileContext& ctx)
 		}
 	}
 
-	for (size_t i = 0; i < countof(FxFlops); ++i)
+	// [RaveYard] resolve expression ahead to differentiate between `floor(object)` and `floor(numeric)`
+	bool singleParamIsObject = false;
+	if (ArgList.size() == 1)
 	{
-		if (MethodName == FxFlops[i].Name)
+		ArgList[0] = ArgList[0]->Resolve(ctx);
+		if (ArgList[0] == nullptr)
 		{
-			FxExpression *x = new FxFlopFunctionCall(i, ArgList, ScriptPosition);
 			delete this;
-			return x->Resolve(ctx);
+			return nullptr;
+		}
+
+		singleParamIsObject = ArgList[0]->IsObject();
+	}
+
+	if (!singleParamIsObject)
+	{
+		for (size_t i = 0; i < countof(FxFlops); ++i)
+		{
+			if (MethodName == FxFlops[i].Name)
+			{
+				FxExpression *x = new FxFlopFunctionCall(i, ArgList, ScriptPosition);
+				delete this;
+				return x->Resolve(ctx);
+			}
 		}
 	}
 


RicardoLuis0:
hmm, that's a bit too targetted of a fix compared to what i was trying to do myself, it helps with class-name conflicts with flops, but not really with class-name conflicts with anything else

MrRaveYard:
@RicardoLuis0 

I think the compiler should be able to understand context/nuance bit more better, if possible. 

Actually, I would be fine with withdrawing this PR in favor of the new casting syntax as I suggested in here: #332
```cpp
var as 'type'
```

Regardless, as for the compiler itself, it gets more confused than it should be (in some cases such as the `Floor`), or here:
```cpp
class Cat
{
    Object Cat() { return self; }

    void Meow()
    {
       Cat a; // "Error: Unable to resolve Cat as a type" (seems like a bug tbh)
    }
}
```

Here's an example of debatable ambiguity I can think of:
```cpp
class Cat
{
	Cat Cat(Cat cat) // The compilers seems to get this part right actually
	{
		return Cat(cat); // the compiler decided it is a member function call not a cast
	}

	void Meow()
	{
		let x = Cat(self); // same
	}
}
```
Imo it should compile into a cast, because you can/should be able to use `super.` (or `Cat.` for static) to use the function.
Edit: Now that I think about it, outside of this very specific example: any unexpected mod or change in engine's ZScript (introducing new classes) would cause issues with existing code if it suddenty went from function calling to casting.

Any specific cases you had on mind?

RicardoLuis0:
```
Cat Cat(Cat cat) // The compilers seems to get this part right actually
{
    return Cat(cat); // the compiler decided it is a member function call not a cast
}
```
for something like this instead of an explicit call, can't really choose either option, it'd need to just be an error for ambiguous call, so you'd either use an explicit cast or `self.Cat(cat)`

RicardoLuis0:
> Actually, I would be fine with withdrawing this PR in favor of the new casting syntax as I suggested in here

i don't mind both the function-style casts being fixed and explicit casts being added (as long as the fixed function-style casts work for any function, instead of only for flops)

MrRaveYard:
> > Actually, I would be fine with withdrawing this PR in favor of the new casting syntax as I suggested in here
> 
> i don't mind both the function-style casts being fixed and explicit casts being added (as long as the fixed function-style casts work for any function, instead of only for flops)

Unlike any other cases, the flops are omnipresent. I'd rather have the `X as 'Y'` as a solution, even more than this actually than allowing every `function(expression)` to be ambiguous until you resolve what `expression` evaluates to. Unless, we want to support overloading in general that is.

So yes, I'm fine with closing this if the `X as 'Y'` gets implemented, since it resolves all the related ambiguity. Otherwise this would be a somewhat inconsistent behavior in regards to other functions (only flops could be "overloaded").

RicardoLuis0:
i do plan to add overloading in the future yes, but nothing soon

RicardoLuis0 (approved):

RicardoLuis0:
but i guess the fixing for all function types can wait until then

PR #380, opened 2025-11-06
PRINT_NOCONSOLE

MajorCooke:
Prevents messages from being added to the open console's display. Very useful for functions like [ProcessNotify](https://zdoom.org/wiki/ProcessNotify) so I can disable item pickup message spam whenever I need to debug and print other information to the console.

### DIFF
--- a/src/common/console/c_console.cpp
+++ b/src/common/console/c_console.cpp
@@ -451,7 +451,7 @@ int PrintString (int iprintlevel, const char *outline)
 				}
 			}
 		}
-		if (Logfile != nullptr && !(iprintlevel & (PRINT_NOLOG|PRINT_NOCONSOLE)))
+		if (Logfile != nullptr && !(iprintlevel & PRINT_NOLOG))
 		{
 			WriteLineToLog(Logfile, outline);
 		}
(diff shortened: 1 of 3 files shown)


src/common/console/c_console.cpp
```diff
@@ -451,7 +451,7 @@ int PrintString (int iprintlevel, const char *outline)
 				}
 			}
 		}
-		if (Logfile != nullptr && !(iprintlevel & PRINT_NOLOG))
+		if (Logfile != nullptr && !(iprintlevel & (PRINT_NOLOG|PRINT_NOCONSOLE)))
```
Boondorl: I don't think this should be implied since the console and log are different things. Ideally it should be a new combination flag or the modder should pass both if they want it.

MajorCooke: Fixed.

mc776:
~~I've got a few things that fake pickup messages using A_Log. Is there some way to make such messages follow this behaviour?~~ EDIT: Making a note to replace those with ProcessNotify.

MajorCooke:
You should be able to with `PrintfEx` in `ProcessNotify` indeed.

Boondorl (approved):

PR #386, opened 2025-11-07
Fixes for menuing

Boondorl:
* Properly account for holding mouse buttons down when checking its dormant state
* Made the dormant check timer longer so it's more lenient
* Fixed music not being properly initialized, forcing the master volume to need to be reset to get it to proper audio levels
* Set music volume default to 1 to account for the master volume being fixed


RicardoLuis0 (approved):

RicardoLuis0:
looks good but should probably wait to merge after someone else can take a look at the mouse changes, 'cause i'm tired and might've missed something lol

PR #394, opened 2025-11-08
Polish new cursor GFX a bit

XaserAcheron:
I dig the new cursor graphics, but I felt they could use a bit of pixel polish. Figured I'd try my hand at it (which is almost a damn pun thanks to the Heretic/Hexen cursors), and here they be!

Hope y'all like 'em -- either way, the new cursors are an improvement over the old ones for sure.


SeventhSentinel:
Personally, I think the Hexen and Doom ones are a little better, but I don't care for the other ones. The Chex spork's handle being shaped like that kinda removes some depth from it compared to Nash's IMO, and the smoothening of the other ones comes at the cost of some detail.

Boondorl:
We'll see what @nashmuhandes thinks when he's back around

XaserAcheron:
I think I'll put the thicker handle back on the Chex spork, but I'm not sure what to do about the others -- a lot of the "detailed" mentioned is pixel aliasing, and I wouldn't know what to put back in without specifics.

Only thing I can think of would be adding bit more 'texture' (i.e. some slight roughness to the solid-color-ish parts) to the Strife and Heretic ones, since the leathery surface is a lot smoother on the updated sprites.

the-phinet:
My preferences:

|Cursor|Nash|Xaser|Why?|
|--|--|--|--|
|Doom||✅|Better depth, less aliasing|
|Chex|✅||Better depth|
|Heretic||✅|Better texture, less aliasing|
|Hexen| |✅|Better texture|
|Strife|✅| |Better depth, better colours|

XaserAcheron:
Just pushed up some updates to the Chex and Strife cursors -- the thicker spork handle is back, and I redid the outer border of the Strife cursor to match the old one's slightly-orangeish gold tinge.

the-phinet:
My updated preferences:

|Cursor|Nash|Xaser|Why?|
|--|--|--|--|
|Doom| |✅|Less aliasing|
|Chex| |✅|Less aliasing|
|Heretic| |✅|Better texture, less aliasing|
|Hexen| |✅|Better texture|
|Strife|🟰|🟰|Extremely similar. No opinion|

nashmuhandes (approved):

XaserAcheron:
Woohoo! Thanks, y'all. :D

PR #399, opened 2025-11-08
Add support for $DOOMWADPATH environment variable in default config

### LINKED ISSUE #340: DOOMWADPATH environment variable doesn't work with more than one entry
### UZDoom Version

UZDoom g4.15pre-735-g8a7bd0be0 - 2025-10-14 22:00:26 -0400 - SDL version

### Game

Doom

### Operating System

Linux x86_64

### OS Version

OpenSuSE Tumbleweed

### Hardware

Irrelevant

### Duplicate Confirmation

### Bug Description

I currently have two paths set in this variable: one for my pwads directory, and one for anything I've made myself. I've tested and it only properly loads the wad when there is only one path in the variable. On other ports, they load wads just fine regardless of which directory they're in, and in fact I'm pretty sure it works with an arbitrary (and reasonable) number of directories.

I also use DOOMWADDIR just to point to the iwads, but that variable AFAIK is only meant to have one path in it, hence the existence of DOOMWADPATH.

Issue persists on Windows, so not an OS problem. On Windows I have a couple more paths that I forgot to add when I installed OpenSuSE, and it's the same situation -- works in other ports. Yes, I put $DOOMWADPATH as a search path in the .ini.

### Replication

_No response_

### Configuration

```ini

```

### Log

_No response_

XaserAcheron:
Closes #340

It dawned on me when generating a brand-new UZDoom config that the engine doesn't support [$DOOMWADPATH](https://doomwiki.org/wiki/Environment_variables) (basically $DOOMWADDIR that lets you specify multiple folders via semicolon separator). Past-me had just hacked around this in GZD's config by adding a bunch of lines manually to my config file, which is a bit tedious, so I wrote a PR this time instead. :P

This PR also adds a new `PathList` config keyword to the `FileSearch.Directories` section, since I didn't want to muck with the behavior of `Path` (for backwards-compat) and the semantics are too different from `RecursivePath` to use that one. Though if y'all would rather just update how `Path` works (which currently just fails if a semicolon is present), I'll gladly rework it.

### DIFF
--- a/src/common/utility/findfile.cpp
+++ b/src/common/utility/findfile.cpp
@@ -304,6 +304,24 @@ const char* BaseFileSearch(const char* file, const char* ext, bool lookfirstinpr
 					}
 				}
 			}
+			else if (stricmp(key, "PathList") == 0)
+			{
+				FString dir;
+				TArray<FString> dirlist = ExpandEnvVars(value).Split(';', FString::TOK_SKIPEMPTY);
+
+				for (FString& dirname : dirlist)
+				{
+					dir = NicePath(dirname.GetChars());
+					if (dir.IsNotEmpty())
+					{
+						BFSwad.Format("%s%s%s", dir.GetChars(), dir.Back() == '/' ? "" : "/", file);
+						if (DirEntryExists(BFSwad.GetChars()))
+						{
+							return BFSwad.GetChars();
+						}
+					}
+				}
+			}
 		}
 	}
 
--- a/src/common/utility/findfile.cpp
+++ b/src/common/utility/findfile.cpp
@@ -51,6 +51,12 @@ CUSTOM_CVARD(Int, i_exit_on_not_found, REQUIRE_DEFAULT, CVAR_ARCHIVE|CVAR_GLOBAL
 	if (self != masked) self = masked;
 };
 
+#ifdef _WIN32
+static constexpr char PATH_SEPARATOR = ';';
+#else
+static constexpr char PATH_SEPARATOR = ':';
+#endif
+
 //==========================================================================
 //
 // D_AddFile
@@ -307,7 +313,7 @@ const char* BaseFileSearch(const char* file, const char* ext, bool lookfirstinpr
 			else if (stricmp(key, "PathList") == 0)
 			{
 				FString dir;
-				TArray<FString> dirlist = ExpandEnvVars(value).Split(';', FString::TOK_SKIPEMPTY);
+				TArray<FString> dirlist = ExpandEnvVars(value).Split(PATH_SEPARATOR, FString::TOK_SKIPEMPTY);
 
 				for (FString& dirname : dirlist)
 				{
--- a/src/gameconfigfile.cpp
+++ b/src/gameconfigfile.cpp
@@ -245,7 +245,7 @@ FGameConfigFile::FGameConfigFile ()
 	{
 		SetSection ("FileSearch.Directories", true);
 		SetValueForKey ("Path", "$DOOMWADDIR", true);
-
+		SetValueForKey ("PathList", "$DOOMWADPATH", true);
 		for (unsigned int i = 0; i < DefaultSearchPaths.Size(); i++)
 		{
 			SetValueForKey ("Path", DefaultSearchPaths[i].GetChars(), true);
--- a/src/gameconfigfile.cpp
+++ b/src/gameconfigfile.cpp
@@ -233,7 +233,7 @@ FGameConfigFile::FGameConfigFile ()
 		SetSection ("IWADSearch.Directories", true);
 		SetValueForKey ("Path", ".", true);
 		SetValueForKey ("Path", "$DOOMWADDIR", true);
-
+		SetValueForKey ("PathList", "$DOOMWADPATH", true);
 		for (unsigned int i = 0; i < DefaultSearchPaths.Size(); i++)
 		{
 			SetValueForKey ("Path", DefaultSearchPaths[i].GetChars(), true);


electricbrass:
Don't forget that the path list separator on unix-like systems is `:` and not `;` like on Windows.

XaserAcheron:
Oops, I missed a spot. :P -- Thanks there!

The solution's a bit low-tech -- I didn't find any more robust way of checking platform other than the various `_WIN32`/`__APPLE__`/etc. defines, so I just tossed in one of those. Lemme know if there's something nicer or more preferable here.

the-phinet (dismissed):

TehRealSalt (changes requested):
Love it & code looks good to me! One small thing though, I would recommend adding it to the `FileSearch` directories as well, not just for IWADs.

While the Doom Wiki's wording does imply that it's only used for IWADs, it would be more consistent with how ZDoom uses DOOMWADDIR. I also double-checked Chocolate Doom, and it also uses it for both -iwad and -file.

XaserAcheron:
Oh welp -- I thought I'd already did that; fixed now.

TehRealSalt (dismissed):

the-phinet (approved):

God-damnit-all:
Would you please add a delimiter that works for all OSes, maybe `|`? I'd like my configurations to be cross-platform compatible wherever possible without having to check OS.

the-phinet:
> Would you please add a delimiter that works for all OSes, maybe `|`? I'd like my configurations to be cross-platform compatible wherever possible without having to check OS.

There is no character that works as a universal path delimiter. Every non-null character is valid in most Linux filesystem paths.

What we could do is support `:` and `;` simultaneously, and try to be clever about scanning paths. That would introduce complexity and a number of edge cases, but I think that could be a good solution.

---

Actually, why do you want this to be changed at all? Environmental variables are generally not shared between different operating system. I could see common dotfiles for Unix systems, but I've never come across a windows user doing that

God-damnit-all:
> Actually, why do you want this to be changed at all? Environmental variables are generally not shared between different operating system. I could see common dotfiles for Unix systems, but I've never come across a windows user doing that

Using the same configurations for cross-platform launchers or scripts. PowerShell is a good example of a script that can be used on all OSes.

> What we could do is support : and ; simultaneously, and try to be clever about scanning paths. That would introduce complexity and a number of edge cases, but I think that could be a good solution.

As long as you bear in mind that `:` is normally used after drive letters, yeah. If you are clever about it that'd be good.

XaserAcheron:
How do other ports handle delimiters across different platforms? This is essentially a cross-port feature, so if UZDoom extends the behavior somewhat, it needs to at least do it in a way that doesn't break existing behavior on one particular OS.

electricbrass:
> How do other ports handle delimiters across different platforms? This is essentially a cross-port feature, so if UZDoom extends the behavior somewhat, it needs to at least do it in a way that doesn't break existing behavior on one particular OS.

AFAIK all ports handle this with platform specific separators, either using ; on Windows and : on Unix-like platforms, or they use ; only, but only because of lacking support for any OS but Windows.

This variable is supposed to act as a normal OS path list, which will always use these separators.

electricbrass:
> Using the same configurations for cross-platform launchers or scripts. PowerShell is a good example of a script that can be used on all OSes.

Launchers that I've seen usually aren't trying to set this environment variable at all, but if they are, I would consider it a bug in the launcher if its not constructing the path list correctly for whatever OS it's currently running on, and storing its own configuration it some other format.

For something like a PowerShell script you can also construct the string from a list in the script with something like this:
```powershell
$PathList = "path/to/folder1", "path/to/folder2", "path/to/folder3"
$DOOMWADPATH = [string]::Join([IO.Path]::PathSeparator, $PathList)
```

And even then, trying to have your files laid out so that absolute paths are the same between Windows and other OSes seems way more hassle than its worth.

God-damnit-all:
> > Using the same configurations for cross-platform launchers or scripts. PowerShell is a good example of a script that can be used on all OSes.
> 
> Launchers that I've seen usually aren't trying to set this environment variable at all, but if they are, I would consider it a bug in the launcher if its not constructing the path list correctly for whatever OS it's currently running on, and storing its own configuration it some other format.
> 
> For something like a PowerShell script you can also construct the string from a list in the script with something like this:
> 
> ```powershell
> $PathList = "path/to/folder1", "path/to/folder2", "path/to/folder3"
> $DOOMWADPATH = [string]::Join([IO.Path]::PathSeparator, $PathList)
> ```

Python or nushell then? Not that it would hurt to be able to use a more conventional line on PowerShell.

> And even then, trying to have your files laid out so that absolute paths are the same between Windows and other OSes seems way more hassle than its worth.

Unless `DOOMWADPATH` works differently from `DOOMWADDIR`, you can use relative paths in the variable.

God-damnit-all:
It's also worth mentioning that forward-slashes for paths works on Windows for these as well, so it's not like platform-agnostic handling would be new here.

electricbrass:
> Python or nushell then?

Not sure about how it's done in nushell but it's even simpler in Python. Pass a list to `os.pathsep.join()`. There's an equivalent in basically every cross platform language.

> It's also worth mentioning that forward-slashes for paths works on Windows for these as well, so it's not like platform-agnostic handling would be new here.

That's only because Windows itself supports both types of slashes.

> Unless `DOOMWADPATH` works differently from `DOOMWADDIR`, you can use relative paths in the variable.

I'm actually curious about this. Do ports generally agree about where it's relative to? CWD is what I would've assumed, which is why I mentioned absolute paths, to avoid inconsistent behavior.

God-damnit-all:
> I'm actually curious about this. Do ports generally agree about where it's relative to? CWD is what I would've assumed, which is why I mentioned absolute paths, to avoid inconsistent behavior.

It's based on CWD, which is what I want to happen. The script sets the CWD to the proper place.

God-damnit-all:
> Not sure about how it's done in nushell but it's even simpler in Python. Pass a list to `os.pathsep.join()`. There's an equivalent in basically every cross platform language.

I'm aware there's something you can do to make it happen. I just like consistency.

If it's too big a deal then alright.

PR #404, opened 2025-11-08
Distance culling

dileepvr:
Distance culling with Fixed distance (with or without fog), and dynamic distance with a target fps.
Works with both hardware and software. But fog only works in hardware.
Uses the ThickFog functionality under the hood. Sky gets blanketed by fog to hide silhouette of map geo popping in/out. This intensity is controllable from the `gl_cull_skyfog` CVar. Usage of skymist in maps is respected.
ThickFog boundary moves slowly in dynamic distance culling mode.
Fog color is a CVar `gl_cullcolor`, but uses sector's `FadeColor` when it is non-zero.
Fog in this mode persists even in fullbright situations like invulnerability or nightvision powerups.

Test with large line-count maps like Frozen Time bridge scene.


MajorCooke:
Between this and [blockrender](https://github.com/UZDoom/UZDoom/pull/405) for side defs, I expect mods will make great use of both to reclaim some FPS on much needed titles. =D

MajorCooke:
Oh, what about brightmaps? Are those accounted for as well?

I would expect the fog envelopes those as well for the sake of brevity. Might look a little funny if if brightmapped textures just popped in out of nowhere.

kcat:
Can you make images to show what this looks like? Particularly the sky fog, since I can't envision how you could overly fog on the sky to hide random geometry pop-in/out, without hiding the sky with fog completely.

dileepvr:
Yes. I have to blanket the sky completely with the color of the fog. So no sky when you use fog to hide the culling.
Here is a demo: https://youtu.be/vzRh_EW-hsM

RicardoLuis0:
> Yes. I have to blanket the sky completely with the color of the fog. So no sky when you use fog to hide the culling. Here is a demo: https://youtu.be/vzRh_EW-hsM

can't you enable skymist if that isn't already enabled when culling?

MajorCooke:
> Can you make images to show what this looks like? Particularly the sky fog, since I can't envision how you could overly fog on the sky to hide random geometry pop-in/out, without hiding the sky with fog completely.

He added a feature that allows sky fog on just the horizon. You should still be able to look up and see the fog above. I forgot what it was, he added it for me.

dileepvr:
Okay. I heard. Skies are important art and are a lot of work.
I added `gl_cull_skyfog` CVAR to make the culling skyfog user controllable, and allowed skymist settings to not get overridden. I also changed the default `skymist1.png` that comes with uzdoom.pk3 to make it work well with this.
Use the console command `skymisttoggle` to test this.
Note that for tall-enough structures, the pop-in will be visible when looking up if you have any kind of gradient in skymist. So if a mapper threw in a light skymist into their mapinfo, it might not hide the popin ever well.
Your new engine default `skymist1` lump (probably only visible in dark mode, and it adopts the fog/fade color used in the map and overlays onto the sky):
skymist1

TwelveEyes:
Might be worth implementing some sort of [PID](https://en.wikipedia.org/wiki/Proportional%E2%80%93integral%E2%80%93derivative_controller) in target FPS mode so the fog doesn't flicker (fog increased until target FPS hit, fog decreased, target fps lost, repeat) so much at the edge.

kcat:
I also wonder if it would be possible to have it respond to rendering complexity (how long it takes frames to render) instead of purely FPS. If the engine is being overwhelmed with something unrelated to what's being drawn, or at least isn't affected by what would be culled, while the rendering itself is plenty fast, the culling distance would pointlessly reduce to a minimum without improving the frame rate much if at all.

dileepvr:
PID is a great idea. I'll just do a weighted sum of the last few frame times and make the delta in culling distance proportional to that. It seems to eliminate the oscillations.
Re: Engine losing time on non-rendering things, I suppose I could literally count the number of subsectors processed. But since multiple frames get rendered every tic, I don't see why frame time wouldn't capture the issue (unless we drop below 70 fps). I can register time at the start of the frame being drawn and move the culling distance update to the end of drawing the frame. That ought to address it.
Do people have test map suggestions? Sunder? Monument?

TwelveEyes:
> Do people have test map suggestions? Sunder? Monument?

Bastion of Chaos is quite performance intensive. PUSS X map 32 is another frame destroyer. You could also try [this](https://www.dropbox.com/scl/fi/ykpc58se8zjp2d41u5jcz/sunlust28_16times.zip?rlkey=qeu9ymqutsllg0o8oznbnvxcd&e=2&st=kxvikzep&dl=0) monstrosity (it's Sunlust map 28 copy pasted 16 times and stitched together) created to stress test Helion.

dileepvr:
> You could also try [this](https://www.dropbox.com/scl/fi/ykpc58se8zjp2d41u5jcz/sunlust28_16times.zip?rlkey=qeu9ymqutsllg0o8oznbnvxcd&e=2&st=kxvikzep&dl=0) monstrosity (it's Sunlust map 28 copy pasted 16 times and stitched together) created to stress test Helion.

I removed the oscillations by adding three separate error regimes and adding a weighted sum of frame-time-errors as the error signal. I never realized how beautiful Sunder gets in map 15 and 16. The linked sunlist28_16times.wad file results in a Fatal error when I try to run it though. Don't know if it is a me problem.

TwelveEyes:
> The linked sunlist28_16times.wad file results in a Fatal error when I try to run it though. Don't know if it is a me problem.

I expected it to chug, not to hard crash. I guess the map is just too much for the engine to handle.

dileepvr:
I can add some basic check for large-renderradius actors, in case someone puts all their tree models in the level in a single actor with a 32676 `RenderRadius`. Shall I stick a credits file in `uzdoom.pk3` for the skymist pngs? CC BY-SA 4.0 just like the logo?

the-phinet:
﻿
﻿

dileepvr:
Was I not supposed to do that? I thought I'd rebase, since it has gotten old.

the-phinet:
lol, you're fine. I assume this PR was created before the bot was added. You can safely disregard it. It's funny how late it showed up to the party

RicardoLuis0 (approved):

Issue #213, opened 2025-10-23
[BUG] Missile-flags with = bug


RonDallas:
### UZDoom version

Last UZDoom build from git

### Which game are you running with UZDoom?

Doom 2

### What Operating System are you using?

Windows 11

### Have you checked that no other similar issue already exists?

- [x] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

The code causes VM abort on every gz/uzd version after 4.11.3 causes "Attempting to pass non-value" VM abort if both missile-flags is in use like in a code below:

    override void Tick()
    {
        Super.Tick();
        bISMONSTER = bMISSILEMORE = bMISSILEEVENMORE = true;
    }

### Your configuration

```ini

```

SanyaWaffles:
are the asterisk part of the code?

RonDallas:
> are the asterisk part of the code?

Nope. I think editing mess up a bit, i create issue from phone.

River-Salmon:
Looks like these flags have been intentionally marked Deprecated in the code starting with version 4.13

Enjay001:
Yep, they have been superseded by the actor property MissileChanceMult ( https://zdoom.org/wiki/Actor_properties#MissileChanceMult )

jekyllgrim:
While the flags are deprecated, the current behavior breaks reverse compatibility. For clarity, this happens specifically if *both* bMissilleMore and bMissileEventMore are set like this; using only one of them, or setting each separately doesn't cause the issue. But again, this is a matter of reverse compatibility rather than good practices.

QuickBanjo:
See attached minimal reproduction that uses OP's code in a replacement for DoomImp.

Able to reproduce in 4.14.3, g4.13.0a. Unable to reproduce in g4.12.2, g4.12.0.

Bisected crash to 8354c4a5c360460cbcc53e141da32ec2684601e4. Note this gives a different error message, `Unknown identifier 'bMISSILEMORE'`. I presume that the property was not exposed to ZScript, which was fixed in the next commit, 3798fd815f7f1c227ec6b9ca522e08fe4f4d0b32. The latter is the first with error `Attempted to pass a non-value`.

Maybe one of the above commits is the culprit. 

[...]

PR #405, opened 2025-11-08
Add "blockrendering" UDMF sidedef property

### LINKED ISSUE #355: [Feature] "Block Rendering" sidedef flag
### Game

All

### Problem Description

In Supplice, there's a particular map that's tanking performance 'cause it's trying to draw the entire universe at once :P -- it's a space station with windows along the edges, and there are some huge vantage points on the exterior of the station that let you view the entire exterior _and_ interior all at once, since everything's visible through the windows.

The solution I'd like to employ is making the windows opaque only from the outside, i.e. have them be one-way-glass, but the only way I know of to achieve this in-engine is to add some physical sector blockers and wire them up to the correct spots. This is a pain to do in the map itself 'cause of how the 'airlocks' are scripted, and it's not very co-op proof in the best case anyway. ;)

### Feature Description

It'd be awesome to have a "block rendering" _sidedef_ (not linedef) flag which tells the renderer to stop drawing anything behind the line (i.e. treat it as if it was 1-sided), only when viewing it from that particular side. This would make it dead easy to set up "one-way" visbility, which would be a massively useful tool for mappers to improve performance in complex maps.

### Additional Media

Here's a quick test wad -- it's basically a mockup of the scenario in Supplice where we hope to use this feature:

[translucent-vis-test.zip](https://github.com/user-attachments/files/23346706/translucent-vis-test.zip)

I left some comments on the linedefs where the new flag would go -- the desire here is that the windows of the building are opaque from the outside _and_ instruct the renderer to skip drawing anything behind it (i.e. from the player start point, the trees on the other side of the building should not draw), while still allowing you to see through the windows whenever you enter the building.

Hopefully this all makes sense, but lemme know if you could use any more examples or clarifications.

XaserAcheron:
Closes #355 

This adds a new "blockrendering" UDMF sidedef property which instructs the renderer to clip the world behind the line when viewing from that side -- it's exactly the same thing the engine did with a 1-sided line, just now you can set it on sidedefs explicitly.

The big use case here is for mappers to be able to set it on lines to create one-way glass or stop rendering faraway/offscreen scenery from one direction for performance's sake (see more musings on that topic in the linked issue).

The feature itself is technically complete in the first commit, but commits 2 and 3 do a small performance optimization by replacing the 1-sided-line check _completely_ with the flag check (i.e. collapse the two checks into one), instead opting to assign the flag to 1-sided lines in the map loader. Older savegames are patched to work too, so this shouldn't cause any interruption of "service". :P

NOTE: at time of writing this is only implemented for the hardware renderer -- I can take a peek at wiring this up to the software renderer too, but that may take some fiddling since the intended behavior is to draw the sky (which the hardware renderer does), rather than HOM ;).

### DIFF
--- a/src/rendering/hwrenderer/scene/hw_bsp.cpp
+++ b/src/rendering/hwrenderer/scene/hw_bsp.cpp
@@ -331,9 +331,11 @@ void HWDrawInfo::AddLine (seg_t *seg, bool portalclip)
 
 	uint8_t ispoly = uint8_t(seg->sidedef->Flags & WALLF_POLYOBJ);
 
-	// [XA] TODO: replace backsector check with flag check outright;
-	// handle the 1s line case by setting the flag in the map loader
-	if (!seg->backsector || (seg->sidedef->Flags & WALLF_BLOCKRENDERING))
+	// [XA] NOTE: this used to be a check for !seg->backsector, but
+	// MapLoader::UpdateSidedefFlags now automagically applies the
+	// WALLF_BLOCKRENDERING flag to any sidedefs belonging to a 1s
+	// line, allowing these two checks to be collapsed into one.
+	if (seg->sidedef->Flags & WALLF_BLOCKRENDERING)
 	{
 		if(!doOob)
 			if (!(seg->sidedef->Flags & WALLF_DITHERTRANS_MID)) clipper.SafeAddClipRange(startAngle, endAngle);
--- a/src/rendering/hwrenderer/scene/hw_bsp.cpp
+++ b/src/rendering/hwrenderer/scene/hw_bsp.cpp
@@ -331,7 +331,9 @@ void HWDrawInfo::AddLine (seg_t *seg, bool portalclip)
 
 	uint8_t ispoly = uint8_t(seg->sidedef->Flags & WALLF_POLYOBJ);
 
-	if (!seg->backsector)
+	// [XA] TODO: replace backsector check with flag check outright;
+	// handle the 1s line case by setting the flag in the map loader
+	if (!seg->backsector || (seg->sidedef->Flags & WALLF_BLOCKRENDERING))
 	{
 		if(!doOob)
 			if (!(seg->sidedef->Flags & WALLF_DITHERTRANS_MID)) clipper.SafeAddClipRange(startAngle, endAngle);
--- a/src/rendering/hwrenderer/scene/hw_bsp.cpp
+++ b/src/rendering/hwrenderer/scene/hw_bsp.cpp
@@ -331,11 +331,11 @@ void HWDrawInfo::AddLine (seg_t *seg, bool portalclip)
 
 	uint8_t ispoly = uint8_t(seg->sidedef->Flags & WALLF_POLYOBJ);
 
-	// [XA] NOTE: this used to be a check for !seg->backsector, but
-	// MapLoader::UpdateSidedefFlags now automagically applies the
-	// WALLF_BLOCKRENDERING flag to any sidedefs belonging to a 1s
-	// line, allowing these two checks to be collapsed into one.
-	if (seg->sidedef->Flags & WALLF_BLOCKRENDERING)
+	// [XA] NOTE: ideally it'd be nice to collapse these checks into one,
+	// but it's possible to add & remove WALLF_BLOCKRENDERING via zscript
+	// so auto-setting it on 1s lines may result in a crash if someone
+	// later tries to remove the flag from a line.
+	if (!seg->backsector || (seg->sidedef->Flags & WALLF_BLOCKRENDERING))
 	{
 		if(!doOob)
 			if (!(seg->sidedef->Flags & WALLF_DITHERTRANS_MID)) clipper.SafeAddClipRange(startAngle, endAngle);
--- a/src/g_level.cpp
+++ b/src/g_level.cpp
@@ -2565,6 +2565,27 @@ AActor* FLevelLocals::SelectActorFromTID(int tid, size_t index, bool clientSide,
 }
 
 //==========================================================================
+// UpdateSidedefRenderFlags
+//
+// [XA] The new WALLF_BLOCKRENDERING flag just exposes the behavior the
+// engine uses internally to cull geometry behind 1-sided lines; rather
+// than have the renderer check for both 1s lines _and_ the flag, just
+// internally set the flag on 1s lines to collapse the checks into one.
+//==========================================================================
+
+void FLevelLocals::UpdateSidedefRenderFlags()
+{
+	for (unsigned int i = 0; i < sides.Size(); ++i)
+	{
+		side_t& side = sides[i];
+		if (side.linedef && !side.linedef->backsector)
+		{
+			side.Flags |= WALLF_BLOCKRENDERING;
+		}
+	}
+}
+
+//==========================================================================
 // IsPointInMap
 //
 // Checks to see if a point is inside the void or not.
--- a/src/g_level.cpp
+++ b/src/g_level.cpp
@@ -2565,27 +2565,6 @@ AActor* FLevelLocals::SelectActorFromTID(int tid, size_t index, bool clientSide,
 }
 
 //==========================================================================
-// UpdateSidedefRenderFlags
-//
-// [XA] The new WALLF_BLOCKRENDERING flag just exposes the behavior the
-// engine uses internally to cull geometry behind 1-sided lines; rather
-// than have the renderer check for both 1s lines _and_ the flag, just
-// internally set the flag on 1s lines to collapse the checks into one.
-//==========================================================================
-
-void FLevelLocals::UpdateSidedefRenderFlags()
-{
-	for (unsigned int i = 0; i < sides.Size(); ++i)
-	{
-		side_t& side = sides[i];
-		if (side.linedef && !side.linedef->backsector)
-		{
-			side.Flags |= WALLF_BLOCKRENDERING;
-		}
-	}
-}
-
-//==========================================================================
 // IsPointInMap
 //
 // Checks to see if a point is inside the void or not.
--- a/src/g_levellocals.h
+++ b/src/g_levellocals.h
@@ -150,6 +150,7 @@ struct FLevelLocals
 	void ApplyCompatibility();
 	void ApplyCompatibility2();
 	AActor* SelectActorFromTID(int tid, size_t index, bool clientSide, AActor* defactor);
+	void UpdateSidedefRenderFlags();
 
 	void Init();
 
--- a/src/g_levellocals.h
+++ b/src/g_levellocals.h
@@ -150,7 +150,6 @@ struct FLevelLocals
 	void ApplyCompatibility();
 	void ApplyCompatibility2();
 	AActor* SelectActorFromTID(int tid, size_t index, bool clientSide, AActor* defactor);
-	void UpdateSidedefRenderFlags();
 
 	void Init();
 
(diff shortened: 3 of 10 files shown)


BouncyTEM:
Fantastic.

MajorCooke:
**Yes!** Now, this can be changed on the fly, right? If so, this'll be _incredibly_ useful in conjunction with portals.

kcat:
With the hardware renderer, what's done to block any geometry behind it from appearing? There are cases where you can inadvertently see "behind" one-sided lines (bad nodes, etc), where you can end up seeing random bits of map geometry. The hardware renderer isn't exact about culling geometry from behind one-sided lines, since it can sometimes be faster to draw extra geometry that's going to be overdrawn by an opaque texture (or blocked from drawing by the z-buffer) rather than pixel-perfect clip it before drawing.

I can see this being somewhat useful in situations where a line's mid texture completely covers up the area you could see through otherwise, but when using translucent, transparent, or missing mid textures, I don't think there's any guarantee about what you'll see in its place.

XaserAcheron:
It's definitely "swim at your own risk" territory if someone wants to try and use it on a non-opaque line in a spot where the player can easily see beyond. This is designed as a performance enhancement tool for use in spots where you _know_ the player isn't going to see beyond it (but is still causing stuff to render "invisibly" 'cause of how the Doom renderer works).

The flag is exposed to ZScript, so it's changeable on the fly that way; could be possible to do a quick ACS shim function too. Could definitely see places where folks would want to toggle it as you enter different parts of the map.

[EDIT] I'll post some real-world example cases from Supplice when I get another related-ish PR (per-sidedef translucency) finished up.

kcat:
> It's definitely "swim at your own risk" territory if someone wants to try and use it on a non-opaque line in a spot where the player can easily see beyond. This is designed as a performance enhancement tool for use in spots where you _know_ the player isn't going to see beyond it (but is still causing stuff to render "invisibly" 'cause of how the Doom renderer works).

I was just a bit concerned since you said:

> I can take a peek at wiring this up to the software renderer too, but that may take some fiddling since the intended behavior is to draw the sky (which the hardware renderer does), rather than HOM ;).

when that's not the guaranteed behavior of the hardware renderer. And if it's only intended to be used where the view is blocked, there shouldn't be a risk of HOM with the software renderer. If anything, having the software renderer create HOM while the hardware renderer can show errant map geometry (where the sky may not even show) would be a good way to help deter misuse, from mappers expecting certain behavior with blocking lines being made see-through that only happens to give what they want but can change in the future or with different settings.

XaserAcheron:
Sorry, I posted that earlier line before stress-testing it -- you're correct, the behavior can't be relied on to always draw plain sky, nor should it.

XaserAcheron:
Some before-and-after shots of the new property in action on Supplice E2M3:

Old: you can peer straight through the exterior windows of the station, which renders a shitton of stuff and looks pretty cluttery.
Screenshot_Supplice_20251109_010423

New: windows blacked out from the exterior (using the new sidedef alpha from lucky number #420) and set to block rendering. Instant 22fps gain.
Screenshot_Supplice_20251109_010445

There are plenty more places where I'm itching to use this; this was just the big obvious use case that inspired this idea in the first place.

RicardoLuis0:
would it be possible similar to the distance cull PR to make some lines block rendering only at a certain distance configurable in the linedef (and to gradually fade to black/whatever else it currently does in some transitionary distance)? that'd let windows smoothly become visible vs blocking depending on the distance from the player (plus possibly defining a texture for it to use when it's fully culled, that way some approximation of what's behind the window can be used)

RicardoLuis0:
if this is too complex to ask feel free to ignore it or leave it for a later PR though, it's just something that i think would make it much more seamless for maps to use

XaserAcheron:
This sounds like something for a different PR, TBH -- different use cases and a lot of work on the implementation in different areas.

RicardoLuis0:
> This sounds like something for a different PR, TBH -- different use cases and a lot of work on the implementation in different areas.

yeah, just something to make windows in particular nicer, to let them still be 2-way up close, but 1-way in the distance

dileepvr:
> The hardware renderer isn't exact about culling geometry from behind one-sided lines, since it can sometimes be faster to draw extra geometry that's going to be overdrawn by an opaque texture (or blocked from drawing by the z-buffer) rather than pixel-perfect clip it before drawing.

Is this why the automap shows lines that you haven't seen yet? Can we mitigate that by holding off on marking a subsector as DRAWN or SEEN unless they are visible through the clipper for, say, ten consecutive frames?

TehRealSalt (dismissed):

XaserAcheron:
Apologies for the bump, but is there anything here that's needed in order to get this merged? It's been approved, but there hasn't been any movement on this in a month. We're wanting to make use of this in Supplice soon.

RicardoLuis0:
just needs fixing the merge conflicts i guess

RicardoLuis0:
> Apologies for the bump, but is there anything here that's needed in order to get this merged? It's been approved, but there hasn't been any movement on this in a month. We're wanting to make use of this in Supplice soon.

also for supplice if you're going with a trunk build i'd suggest reverting this https://github.com/UZDoom/UZDoom/commit/605d627e24d34a6e5d1782dae3f6e2c4bd29ba7b, we're planning to revert it for the 5.0 release branch (not on trunk though) since it does still have possibly undiscovered crashes

the-phinet:
﻿
﻿

flippygaming:
Would this be something that can be turned on and off in ACS during the level?

dileepvr:
Not as written no. Xaser is auto-setting single-sided lines with the same flag and using that as the sole check for clipping.

MajorCooke:
It can be set via ZScript though, so if nothing else is done, then ScriptCall can be used. Though it would be helpful to implement directly onto ACS itself.

src/rendering/hwrenderer/scene/hw_bsp.cpp
dileepvr, line 1: Majorcooke's comment brought up the possibility of toggling this flag from zscript even for one-sided lines. Maybe we keep the `!seg->backsector` check along with the flag check, take the small performance hit (should be negligible, immeasurable even),  and simplify the rest of the commit.

XaserAcheron:
Hey, I know y'all are busy, but could we get some additional eyes (read: community volunteers) to do a sweep on issues like this which are 90-something-complete done? The merge conflict that's left is super-trivial to fix and there's really no reason for it to have sat there waiting on me for months and months. I do what I can, but I've got a ton of stuff on my plate and could really use a hand with things like this.

XaserAcheron:
Ugh, sorry for my previous comment -- I'm venting some frustrations that are not at all y'all's fault (I've got some grumbles about the general lethargy of parts of the Doom community, but the UZDoom team is very much the direct opposite :P -- y'all are killing it, keep that train goin'). Got this one fixed; should be fine if CI doesn't complain.

That said, it's probably still worth finding someone to make a sweep through the currently-open PRs (ideally some new blood so it doesn't overextend one of the current devs, since y'all are busy as heck). There are probably some others open that only need trivial fixes like this one, and waiting on the original contributors to come back sucks since we're wishy washy goofballs.

the-phinet:
No need to apologize, I can very much empathize with the above. Having 30+ PRs open is stressing us all out a bit.

Basically the thing is that 5.0 is essentially feature-frozen, and we're just trying to finish the last coupe things needed to get it out the door. With issues like this one I don't know enough if it's a 'feature' or a 'fix', so I'm deferring to the others to make that call.

Some things are trivial fixes and I'm super willing to fix them, but I'm not sure what the optics of me modifying PRs without explicit permission are -- especially for the old ones where our contributor agreement isn't agreed to. For simple things that have been sitting for a while with merge conflicts, what is the consensus about someone from uzd going in and fixing them? What about PRs with requested changes?

RicardoLuis0:
> Hey, I know y'all are busy, but could we get some additional eyes (read: community volunteers) to do a sweep on issues like this which are 90-something-complete done? The merge conflict that's left is super-trivial to fix and there's really no reason for it to have sat there waiting on me for months and months. I do what I can, but I've got a ton of stuff on my plate and could really use a hand with things like this.

ah, didn't realize you were fine with it (also to aynone else PR-ing, if you're fine with us fixing the conflicts/etc in your PR just say so and we will)

XaserAcheron:
Personally speaking, I think the only time I'd be weirded out by a PR being modified is if it either changes the intent or if it's done very shortly after creation -- for this one, the way to fix the merge-conflict is super obvious and it's been months. To me it's totally reasonable to fix it up, since if I didn't want the code to get in then I wouldn't have submitted it in the first place. ;)

Since the contributor agreement exists now, it'd be good to state this explicitly somewhere -- e.g. something like "we may make additional commits in PRs to fix merge conflicts or bugs in cases where the necessary action is clear and unambiguous", or somesuch. I'm not great at wording so that'll need workshopping. :P

Failing that (or, if that's agreed upon, for older cases that pre-date contributing.md), it's also possible to make a pull request _in the forked repo_ that targets the branch the PR uses -- i.e. a PR for this PR. That will give the original contributor (e.g. myself, in this case) the ability to explicitly accept or reject the modification. If it's accepted, then the original PR automatically updates with the merged code (since under the hood it's just new commits on the PR's branch, just like everything else). Git is cool like that.

RicardoLuis0 (approved):

src/rendering/hwrenderer/scene/hw_bsp.cpp
RicardoLuis0, line 345: actually wait, since WALLF_BLOCKRENDERING is toggleable from zscript, it means that a null-pointer access in the renderer can be triggered by user code by simply removing the flag from a one-sided line

RicardoLuis0 (changes requested):

src/rendering/hwrenderer/scene/hw_bsp.cpp
XaserAcheron: Is there any way to handle this on the ZScript level? i.e. when trying to modify that field in ZScript, run a function afterward on the line that forces the flag back on if it's 1-sided? I'm trying to not have to introduce a second check in the renderer code path.

RicardoLuis0: no, zscript has no concept of "properties" for fields, and the ones that try to do something similar (the legacy flags) are busted af

RicardoLuis0: re: #213

XaserAcheron: OK, then what would you suggest?

dileepvr: `if (seg->sidedef->Flags & WALLF_BLOCKRENDERING || !seg->backsector)` ?
I don't think the second check is made if the first one returns true in modern systems.
We could do some tests on linedef heavy maps.

XaserAcheron: That's the exact scenario I was advised to avoid when pitching this idea originally -- that would double the amount of checks for every 2-sided linedef that don't have the flag set in the editor (i.e. the vast majority). If someone has the means to profile this to see if it's a problem in practice, that would be good here, but a 1-sided line without the flag is an invalid state and it'd be best if there's some way of ensuring the line doesn't get into this state at all, rather than add a check to a relatively hot code path.

RicardoLuis0: > That's the exact scenario I was advised to avoid when pitching this idea originally -- that would double the amount of checks for every 2-sided linedef that don't have the flag set in the editor (i.e. the vast majority). If someone has the means to profile this to see if it's a problem in practice, that would be good here, but a 1-sided line without the flag is an invalid state and it'd be best if there's some way of ensuring the line doesn't get into this state at all, rather than add a check to a relatively hot code path.

the thing is, every line already has a ton of checks on `HWDrawInfo::AddLine`, it wouldn't be going from 1 check to 2, it'd be going from 101 checks to 102 (well, not exactly 101 - but it's way more than 1)

XaserAcheron:
Finally got 'round to updating the PR with the requested changes -- the stuff to auto-apply the flag to 1s lines has been reverted in favor of just doing two checks; this fixes up any sort of crashiness if players try and set/unset the flag from ZScript.

RicardoLuis0 (dismissed):

RicardoLuis0 (approved):
had to clean up the git history a bit because of the merge but all good

Issue #378, opened 2025-11-06
libtooltipmenu causes a VM Abort when re-acessing menu(s)


kinker31:
### UZDoom Version

g4.15pre-765-g4be5c66d3

### Game

Doom

### Operating System

Linux x86_64

### OS Version

Arch Linux 6.17.7-zen1-1-zen

### Hardware

AMD GPU and CPU

### Duplicate Confirmation

- [x] I have searched and not found a similar issue.

### Bug Description

When attempting to acess a menu more than twice with the libtooltipmenu library, VM engine immediately crashes and leads right into an Abort Exception screen.

### Replication

1. Load UZDoom with a mod that uses libtooltipmenu, like Gun Bonsai or GZArchipelago. I'll be using Gun Bonsai for the rest of this replication guide.
2. Start the game, and open the upgrade(s) menu with any keybind of your choice.
3. Exit the menu, and open it back up again. The crash should occur.
4. Attempting to open the menu again after that always aborts the VM.

### Configuration

[Not sure if this helps, but this'll do.](https://gist.github.com/kinker31/fedd6b73291d730bb05eb1cee973dfa2)

### Log

```
VM execution aborted: array access out of bounds. Size = 0, current index = 0

Called from OptionMenu.Init at uzdoom.pk3:zscript/engine/ui/menu/optionmenu.zs, line 182
Called from TFLV_TooltipOptionMenu.InitDynamic at GunBonsai-0.10.6.pk3:ca.ancilla.bonsai/libtooltipmenu/tooltipoptionmenu.zsc, line 16
Called from TFLV_Menu_StatusDisplay.Init at GunBonsai-0.10.6.pk3:ca.ancilla.bonsai/menu/statusdisplay.zsc, line 5
Called from TFLV_EventHandler.ShowInfo at GunBonsai-0.10.6.pk3:ca.ancilla.bonsai/eventhandler.zsc, line 135
Called from TFLV_EventHandler.NetworkProcess at GunBonsai-0.10.6.pk3:ca.ancilla.bonsai/eventhandler.zsc, line 166
Called from Menu.SetMenu [Native]
```

kinker31:
I've also raised a similar report [over here](https://github.com/ToxicFrog/doom-mods/issues/56#issuecomment-3482691449), but right now ToxicFrog's waiting for UZDoom to get ported over to nix packages, so I figured I'd let you know as well.

Enjay001:
[...]

Issue #347, opened 2025-11-03
VM Abort when using an options menu colour picker


Enjay001:
### UZDoom Version

UZDoom g4.15pre-733-g34aa137d (Also happens with GZDoom git builds)

### Game

All

### Operating System

Windows 10

### Duplicate Confirmation

- [x] I have searched and not found a similar issue.

### Bug Description

When changing automap colours for a _second time_, UZDoom will give a VM Abort. [edit - any colour picker can cause it]

This also happens in GZDoom Git, but not in GZDoom official.

### Replication

1. Start a game
2. Go to the automap options menu
3. Go to the "Set Custom Colours" menu
4. Pick one of the colours and hit enter to bring up the colour picker (you can change the colour or not, it doesn't matter)
5.  Close all the menus and go back to the game
6. Bring up the menu again and go all the way back to the automap colours menu
7. Select a colour and hit enter
8. Boom! VM abort

See video:
https://youtu.be/v16lS4GHWGo

### Configuration

```ini

```

### Log

MAP01 - Entryway

VM execution aborted: array access out of bounds. Size = 0, current index = 4

Called from OptionMenu.Init at uzdoom.pk3:zscript/engine/ui/menu/optionmenu.zs, line 182
Called from ColorpickerMenu.Init at uzdoom.pk3:zscript/engine/ui/menu/colorpickermenu.zs, line 98
Called from OptionMenuItemColorPicker.Activate at uzdoom.pk3:zscript/engine/ui/menu/optionmenuitems.zs, line 1085
Called from OptionMenu.MenuEvent at uzdoom.pk3:zscript/engine/ui/menu/optionmenu.zs, line 769

Enjay001:
A bit more:
Any colour picker will do (e.g. the ones in the display options).
You do not have to actually go into a game.  You can get this simply by starting UZDoom, going to the options menu from the title screen, activating any colour picker, leaving the colour picker and going back in.

Example from GZDoom Git Build
https://www.youtube.com/watch?v=dnldbHvsNfU

Enjay001:
Also reported for GZDoom:
https://github.com/ZDoom/gzdoom/issues/3408

Enjay001:
Not surprisingly, but I thought I'd confirm, it still happens with a GitHub automatic build from today g4.15pre-746-g96db88eb

RexenSquared:
[...]

PR #411, opened 2025-11-08
Fix for #347 VM Abort when using an options menu colour picker

hhorne:
Per discussion in #347. Reset `mSelectedItem` to `-1` when `OnDestroy()` gets called. This prevents the crash when opening the menu a second time. I believe this should apply to #378 as well.

My editor ate some misc. whitespace in this file. Hope that's not too much "noise" cluttering the change.

### DIFF
--- a/wadsrc/static/zscript/engine/ui/menu/colorpickermenu.zs
+++ b/wadsrc/static/zscript/engine/ui/menu/colorpickermenu.zs
@@ -136,7 +136,7 @@ class ColorpickerMenu : OptionMenu
 				MenuSound ("menu/cursor");
 				mGridPosY = 0;
 				// let it point to the last static item so that the super class code still has a valid item
-				mDesc.mSelectedItem = mStartItem+7;	
+				mDesc.mSelectedItem = mStartItem+7;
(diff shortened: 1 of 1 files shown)


the-phinet:
> My editor ate some misc. whitespace in this file. Hope that's not too much "noise" cluttering the change.

Lol no worries. Cleaning up old code is a good thing

Protip, enable this:
image

---

It looks like this fixes the cause, but would not protect against the abort in #378

To protect against that, something like this should work

```diff
--- a/wadsrc/static/zscript/engine/ui/menu/optionmenu.zs
+++ b/wadsrc/static/zscript/engine/ui/menu/optionmenu.zs
@@ -178,7 +178,7 @@ class OptionMenu : Menu
                        }
                }
 
-               if (mDesc.mSelectedItem >= 0)
+               if (mDesc.mSelectedItem >= 0 && mDesc.mSelectedItem < mDesc.mItems.Size())
                        UpdateTooltip(mDesc.mItems[mDesc.mSelectedItem].GetTooltip());
        }
```

the-phinet (approved):

PR #415, opened 2025-11-08
Updated CI to release rotating dev builds based on latest commit to trunk

DyNaM1Kk:
This PR updates the CI to zip the compiled builds of UZDoom and upload them to a release. Release is rotating, meaning only the latest version will be available for download. The job is only ran if pushed to trunk so the release always represents the latest available dev build of the port. Two things will need to be done if this PR were to be merged:
  1. Create a tag named "uDev".
  2. Create a _pre-release_ named "UZDoom - Latest Dev Build"

The tag and release names are not final and can be changed as needed. The release description will say: 

> This is an automatically generated development version based on the latest commit ${{ github.sha }} pushed to trunk

Where ${{ github.sha }} will be a hyperlink to the last pushed commit. Of course the description can also be changed and expanded to include more information if needed.

I am marking this as a draft until the pre-requisites and requests (if any) are met. The functionality is very versatile and can even be used to set up scheduled nightly builds if desired. If anyone would like to dive deeper into what we can do with this I will include links to documentation on triggers and the `github` context that can be used to apply more specific filters or include more information in the release.

https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context
https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -109,6 +109,9 @@ jobs:
             cp ../fm_banks/GENMIDI.GS.wopl ../fm_banks/gs-by-papiezak-and-sneakernets.wopn package/fm_banks
 
             cp ${{ matrix.config.build_type }}/uzdoom.exe ${{ matrix.config.build_type }}/*.pk3 package
+            
+            cd package
+            7z a "../${{ matrix.config.name }}-${{ matrix.config.build_type }}.zip" *
 
         elif [[ "${{ runner.os }}" == 'macOS' ]]; then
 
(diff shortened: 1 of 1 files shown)


the-phinet:
Nice! I'll look at this as soon as I can :grin: 

Initial feedback without looking at all: I would prefer the tag to be `nightly` and the names to be `UZDoom - Nightly Build`. Those are pretty standard among other projects

arrowgent:
https://github.com/UZDoom/UZDoom/discussions/401
release file naming can be useful and informative


going further, its possible to setup nightly builds that pin the top release and continuously update.  allowing developers and testers to quickly get access to changes
while still allowing stable & RC releases

DyNaM1Kk:
> Nice! I'll look at this as soon as I can 😁
> 
> Initial feedback without looking at all: I would prefer the tag to be `nightly` and the names to be `UZDoom - Nightly Build`. Those are pretty standard among other projects

Cheers.

I'm not familiar with how others do it, but should it be called nightly when it's not on an actual nightly schedule?

the-phinet:
﻿
﻿

PR #420, opened 2025-11-09
Sidedef "alpha" UDMF field

XaserAcheron:
Pretty self-explanatory: this adds an "alpha" UDMF field to sidedefs, which takes precedence over the linedef alpha if set. Pairs well with #405 for making one-way glass.

If desired, I could also do a similar thing to #405 and replace the linedef "alpha" field entirely in the internal struct and save a couple of branch instructions, but I'd like to wait until after #405 is all settled before attempting it since it'd touch the same areas of code and lead to conflicts if I were to do it now. Would rather not couple 'em too tightly together just yet.

### DIFF
--- a/src/gamedata/r_defs.h
+++ b/src/gamedata/r_defs.h
@@ -1259,6 +1259,7 @@ struct side_t
 	int16_t		Light;
 	int16_t		TierLights[3];	// per-tier light levels
 	uint16_t	Flags;
+	double		alpha;
 	int			UDMFIndex;		// needed to access custom UDMF fields which are stored in loading order.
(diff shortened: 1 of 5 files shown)


Boondorl:
> If desired, I could also do a similar thing to https://github.com/UZDoom/UZDoom/pull/405 and replace the linedef "alpha" field entirely in the internal struct and save a couple of branch instructions

I think we won't be able to set it like that since the alpha is changeable via ZScript and probably ACS (unlike lines with no backsector which must remain that way).

RicardoLuis0 (approved):

the-phinet:
﻿
﻿

PR #444, opened 2025-11-11
Added lighting accessibility options

Boondorl:
This includes the ability to both control the light radius around the player and a new option that adds a flat amount of light to the world, trying to preserve the overall lighting model while making it brighter. Meant to be a replacement for the lightmodes as requested by #4. Can be found in the accessibility options menu.

I'm still not entirely sure if this implementation is correct for r_extralight, but it seems to give at least vaguely expected results. Fog is a bit questionable in Hexen but I'm not sure on how that's supposed to interact with the light level.

### DIFF
--- a/src/rendering/hwrenderer/scene/hw_lighting.cpp
+++ b/src/rendering/hwrenderer/scene/hw_lighting.cpp
@@ -187,7 +187,7 @@ float GetFogDensity(FLevelLocals* Level, ELightMode lightmode, int lightlevel, P
 		// case 2: black fog
 		if ((!isDoomSoftwareLighting(lightmode) || blendfactor > 0) && !(Level->flags3 & LEVEL3_NOLIGHTFADE))
 		{
-			density = distfogtable[lightmode != ELightMode::LinearStandard][hw_ClampLight(lightlevel)];
+			density = distfogtable[lightmode != ELightMode::LinearStandard][hw_ClampLight(lightlevel, false)];
 		}
 		else
 		{
--- a/src/rendering/hwrenderer/scene/hw_lighting.h
+++ b/src/rendering/hwrenderer/scene/hw_lighting.h
@@ -7,9 +7,11 @@
 
 struct Colormap;
 
-inline int hw_ClampLight(int lightlevel)
+EXTERN_CVAR(Int, r_extralight)
+
+inline int hw_ClampLight(int lightlevel, bool addExtra = true)
 {
-	return clamp(lightlevel, 0, 255);
+	return clamp(lightlevel + r_extralight * addExtra, 0, 255);
 }
 
 EXTERN_CVAR(Int, gl_weaponlight);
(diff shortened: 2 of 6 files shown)


RicardoLuis0 (dismissed):

TehRealSalt (commented):
I think that the software renderer is not applying `r_extralight` to the walls:

| | Hardware | Truecolor Software |
| --- | --- | --- |
| `r_extralight 0` (default) | Screenshot_Doom_20251111_161051 | Screenshot_Doom_20251111_161058 |
| `r_extralight 8` | Screenshot_Doom_20251111_161042 | Screenshot_Doom_20251111_161038 |

I tested by disabling menu fade and changing it from the Accessibility menu, and changing renderer via the Set Video Mode menu.

This is more anecdotal and might just be placebo, but I also think r_extralight is applying more strongly in Software than in Hardware? I don't expect 100% matching but the difference feels strong between the floors in the two r_extralight screenshots. I think it'd be easier to tell for certain when the walls are all getting the light, though.

Boondorl:
Part of the issue with software seems to be just how the paletted textures work, but I also don't think I'm fully passing the values correctly. It's a lot more confusing to handle than hardware so I wouldn't be surprised if some things were getting missed. I used Monster Condo (Doom 2 - MAP27) to test since it's a very dark map and noticed it seemed to work well enough, but the issue is there's no way for me to verify this stuff without just outright modifying the map itself and doing a comparison. In general it feels like it's stronger than it should be, but Doom's lighting model is also just a bit wonky like that.

I did, however, notice fog wasn't applying correctly to walls with it in Hexen, so it probably is getting missed

RicardoLuis0:
seems that it's only applying to flats in software somehow

the-phinet:
Light radius option looks great!

When #35 ends up getting merged, what purpose does a flat boost to brightness serve?

Boondorl:
So basically, it's different modes of accessibility. A lot of people liked dark mode for the fact it turned the light radius off entirely (something you can do from the menu now), but some maps might still be too dark for their taste so they want to bump the global light level to help out. It works by actually adding light level to the sectors themselves so it's more akin to Minecraft's brightness slider rather than a screen calibrator. Since this is meant to replace the old light mode options, I feel it's important we have ways to emulate what they did in an appropriate context

Boondorl:
So I think I was in the right ballpark of the software functions I was modifying, but where I got it wrong was I needed to apply the light to the core shade calculation from light level. Hardware and truecolor software seem to give matching colors now, but the issue is now the falloff in hardware looks too extreme in comparison (the shadows are brightened way too much). I knew something looked off about it, I think the place I put the additional light was simply not correct (too many things are likely building of each other)

Boondorl:
Ok, think I found the issue. There was indeed a little bit of compounding going on in certain areas. Both the true color sw and hardware seem to give me the same result now + the fog issue in Hexen has been fixed, so I think this is finally ready.

TehRealSalt (approved):
Looks good now!

I think the only suggestion I could make is that if the range on was bigger. I cross-referenced KEX's extra lighting option and its 100% setting is just barely shy of fullbright. But the feature works so :+1:

Boondorl:
Yeah, part of the reason I kept it so clamped was because of the hw extra light accidentally nuking shadows, it made everything look far brighter than intended. I'll probably have to tweak it later but I find the brightness adjustment still noticeable on dark maps like Monster Condo

mc776:
The dynamic lights already do that for a lot of maps...

PR #452, opened 2025-11-11
Use Steam libraryfolders.vdf on Linux & Mac

TehRealSalt:
I was annoyed that UZDoom didn't detect my Steam games and ended up down a rabbit hole...

For a while I had simply thought that separate library folders was not supported in Linux, but as it turns out: it did at one point, but Steam updated how it's handled, and then the Windows' code was updated without looking at other platforms (business as usual, lol)

I considered letting it look at `config.vdf` as a fallback, but decided against it because:
- The Windows version of this code doesn't look for this file at all.
- I looked at my system's own `config.vdf` and it does not have the "BaseInstallFolder" stuff that the old ParseSteamRegistry was trying to look for.
- I blame'd this code and it was around 2-11 years old, so I think it was just really dated in general.

The code already had a TODO for removing the duplication, so I went ahead and fixed that to try and prevent this from happening again in the future. The platform-specific code is only for getting the path to the file now.

**I changed it for Mac because I'm assuming Steam would do this for all 3 platforms if it did it for the other 2, but I'm not able to confirm for certain if this actually is the case or not.**

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -799,6 +799,7 @@ set (PCH_SOURCES
 	d_net.cpp
 	d_netinfo.cpp
 	d_protocol.cpp
+	d_steam.cpp
 	doomstat.cpp
 	g_cvars.cpp
 	g_dumpinfo.cpp
--- a/src/common/platform/posix/i_system.h
+++ b/src/common/platform/posix/i_system.h
@@ -47,7 +47,7 @@ bool I_PickIWad (bool showwin, FStartupSelectionInfo& info);
 
 // [RH] Checks the registry for Steam's install path, so we can scan its
 // directories for IWADs if the user purchased any through Steam.
-TArray<FString> I_GetSteamPath();
+FString I_GetSteamPath();
 
 TArray<FString> I_GetGogPaths();
 
--- a/src/common/platform/win32/i_system.h
+++ b/src/common/platform/win32/i_system.h
@@ -45,7 +45,7 @@ bool I_WriteIniFailed (const char* filename);
 
 // [RH] Checks the registry for Steam's install path, so we can scan its
 // directories for IWADs if the user purchased any through Steam.
-TArray<FString> I_GetSteamPath();
+FString I_GetSteamPath();
 
 // [GZ] Same deal for GOG paths
 TArray<FString> I_GetGogPaths();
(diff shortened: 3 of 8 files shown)


RicardoLuis0:
actually, looking at it, most code in posix/i_steam.cpp and win32/i_steam.cpp should be merged and moved outside of platform-specific folders, seems that 99% of it is the same in windows/linux/macos

TehRealSalt:
> actually, looking at it, most code in posix/i_steam.cpp and win32/i_steam.cpp should be merged and moved outside of platform-specific folders, seems that 99% of it is the same in windows/linux/macos

I was worried about doing that (wanted to avoid increasing the amount of testing this would need), but I've went ahead and done it

arrowgent:
an improvement for sure

so long as i can still turn it off, i dont want it looking in my steam folder.
i have my own folder with all the correct iwads, that is the only folder i want it to look for

arrowgent:
i can mention that
`		regPath.Format("%s/.steam/steam/config/libraryfolders.vdf", home);`
is a symbolic link generated by steam.deb and any debian/ubuntu distro

the factual location installation folder is still
`$HOME/.local/share/Steam/config`


depending on your setup you can check
https://repo.steampowered.com/steam/

for arch & rpm, i cant comment on their install methods.

https://wiki.archlinux.org/title/Steam#Directory_structure
`The default Steam install location is ~/.local/share/Steam`

leahanderson1:
> i can mention that
> `		regPath.Format("%s/.steam/steam/config/libraryfolders.vdf", home);`
> is a symbolic link generated by steam.deb and any debian/ubuntu distro
> 
> the factual location installation folder is still
> `$HOME/.local/share/Steam/config`
> 
> 
> depending on your setup you can check
> https://repo.steampowered.com/steam/
> 
> for arch & rpm, i cant comment on their install methods.
> 
> https://wiki.archlinux.org/title/Steam#Directory_structure
> `The default Steam install location is ~/.local/share/Steam`
> 
> 

On Arch, the .steam folder does in fact exist.
Still would be better to use the non symlinked location though.

leahanderson1:
Oh that has in fact been done

leahanderson1 (approved):
just compiled it, works as intended and it shows all my stuff!

the-phinet (approved):
Nice. Looks correct and works well on my systems.

PR #456, opened 2025-11-12
Add SectorFillColor automap color

XaserAcheron:
One little detail I've always admired about Marathon's automap (despite its myriad other issues :P ) is how it uses a solid color fill for sectors, that way it's easy to distinguish playable areas from void space. Textured mode is nice, but it's a bit too "busy" for my eyes, so I added a new "SectorFillColor" automap color that does what it says on the tin. ;)

A quick demo from Supplice E1M1:
Screenshot_Supplice_20251110_042827

Some technical details:
- This adds a `SectorFillColor` MAPINFO property for setting the color in a mod, and a pair of cvars (`am_sectorfillcolor` and `am_ovsectorfillcolor` for the normal automap and overlay, repsectively) for user customization.
- The fill is only drawn if the color is defined; it doesn't try and draw a solid background-color fill in its absence or anything like that.
- The default for the cvars and all built-in color schemes (i.e. Traditional Doom, Raven, and Strife schemes) is blank, i.e. no fill will be drawn unless the properties are explicitly set.
- Textured mode takes priority; if the `am_textured` cvar is on, textures will be drawn instead and SectorFillColor will be ignored.
- SectorFillColor respects the "Not drawn on textured automap" UDMF flag, just like Textured mode, so maps that take textured mode into account when doing automap cleanup will work great out of the box with this.
- This slots right into the existing automap color system, so user-side overrides (i.e. opting out of mod-defined colors and substituting one's own) Just Work(tm).

A couple of caveats to this PR:
1. Internally, some places in the automap uses the color #010000 (i.e. near-black, with red = 1) as a placeholder value that means "blank; don't draw" -- which is a rather ugly implementation, but color cvars don't have any proper way of defining a blank/null color, so I borrowed this definition for now. It's not ideal, but works in a pinch.
2. The two new color cvars aren't yet exposed in the options menu, since the color picker UI needs to be updated to add support for picking a blank color.

That said, I figured it'd be best to go ahead and open this, since this feature is usable in mods before caveat no.2 is solved and I could use some brainstorming on how best to solve caveat no.1 (or if it even really matters).

I've got some future extensions I'd like to make to this (e.g. UDMF properties that allow sectors to explicitly specify what color or texture to draw), but let's start here first. :P

[EDIT] Welp, I guess it'd be a good idea to post a demo wad. :P -- Here's a quick wad with the color scheme from the Supplice screenshot, fillcolor n' all:
[automap-sectorfill-demo.zip](https://github.com/user-attachments/files/23495252/automap-sectorfill-demo.zip)


RicardoLuis0:
oooooooooooooooooooh, this is really nice

RicardoLuis0:
how does it work across static portals though? your example screenshot doesn't seem to showcase those

XaserAcheron:
Same rules as textured mode -- it will only draw the fill color in the "current" portal group.

Screenshot_Doom_20251112_175631

[Demo wad from the screenshot: [portal-test.zip](https://github.com/user-attachments/files/23512077/portal-test.zip)]

Portal groups are something I'd like to spend some time on in general, i.e. make that sort of stuff more configurable by the mapper, but at minimum it'd need a separate fill color (if anything) anyway.

RicardoLuis0 (approved):
seems good then yeah 👍

src/am_map.cpp
```diff
@@ -3368,7 +3409,7 @@ void DAutomap::Drawer (int bottom)
 	}
 	activateNewScale();
 
-	if (am_textured && !viewactive)
+	if ((am_textured || AMColors.isValid(AMColors.SectorFillColor)) && !viewactive)
```
TehRealSalt, line 3412: I believe this `!viewactive` is preventing the fill color from applying at all in Overlay mode.

TehRealSalt (changes requested):
I love the extra clarity the fill gives to the standard mode!

I noticed it doesn't function in Overlay mode. I've narrowed down which code line that I think is preventing it from being used.

That said, as an Overlay mode enjoyer I'm not sure if having a fill color covering the game view would look very nice. If you do want to support Overlay mode, IMO it'd be nice to also give it another console variable for adjusting the fill's opacity. (Might also be applicable for enabling textured mode for Overlay, or for showing the `AUTOPAGE` background through the fill color in the standard mode. Just a suggestion, tho!)

Otherwise, I also think it'd be fine to just not support it, in which case I'd say just remove the console variable for it, so it doesn't imply that it is a feature.

XaserAcheron:
Hmm, I think the prudent thing to do here is to remove the overlay color for now, then maybe add that later when there's a nicer way to define an alpha value.

It'll be a bit of work, but if Color CVars were updated to be RGBA instead of just RGB, then that would solve a couple of woes all at once (i.e. no need for those "01 00 00" blank placeholders, and the overlay fill color can contain an alpha value directly).

I might take a peek at that shortly, just to see how tricky that is in practice.

the-phinet:
﻿
﻿

mc776:
Is there some way to disable notifications specifically for this bot???

MajorCooke:
Come to think of it, I'd prefer not having this bot leaving comments at all.

the-phinet:
Please discuss the bot [here](https://github.com/UZDoom/UZDoom/discussions/861), instead of leaving off-topic comments in other pull requests

nashmuhandes:
Just wanted to check in on this. If the issue was only about the overlay that TehRealSalt brought up, perhaps best to just remove the overlay component now and add that later, as I think it's too trivial of a thing to block this feature.

RicardoLuis0:
being merged as #1350

PR #457, opened 2025-11-12
Client-sided dynamic lights

Boondorl:
This allows them to run independent of the networking and world state if needed.

### DIFF
--- a/src/playsim/dthinker.cpp
+++ b/src/playsim/dthinker.cpp
@@ -45,6 +45,7 @@
 #include "v_video.h"
 #include "g_cvars.h"
 #include "d_main.h"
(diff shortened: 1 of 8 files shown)


src/playsim/dthinker.cpp
```diff
@@ -56,6 +56,8 @@ extern int BotWTG;
 
 IMPLEMENT_CLASS(DThinker, false, false)
 
+int WorldPaused(bool checkLag = false);
```
RicardoLuis0: this is in a header now so it should probably include r_utility.h instead of the raw function declaration

RicardoLuis0 (approved):

PR #467, opened 2025-11-13
Fix "archimage" skill level name

Gaerzi:
It is archimage, with an i, in vanilla Hexen. Not the eyeless archmage.


Gaerzi:
It's only fitting that my first contribution would be a spelling nitpick. 😛

Boondorl:
Localization changes should go through our Weblate (which allows you to login via a GitHub account). You can find the link on the main repo page at the bottom of the README (it's the image itself). Changes to it will be propagated back to here when synchronized

PR #470, opened 2025-11-14
Update to SDL3

SED4906:
This is, for the most part, the bare minimum porting work.


RicardoLuis0:
one thing, changes to zwidget should be sent to the proper repository, subtrees should not be manually modified

arrowgent:
currently ubuntu LTS do not have SDL3 in their repositories

https://packages.ubuntu.com/search?keywords=libsdl3

if the uzdoom binary compiles with sdl3 static built-in that would work.  otherwise users on current LTS cannot run uzdoom.

until ubuntu 26.04 comes out it is limited to SDL2 libraries

debian however, trixie 13:
https://packages.debian.org/trixie/libsdl3-dev
and as far as i know, bookworm 12 did not have SDL3 in the repos.

as for CI, i am unsure if `ubuntu-latest` is using the LTS 24.04 or 25.10 builds
myself and many others do not consider the interim releases (24.10, 25.04, 25.10) current releases.


unvanquished (custom quake3 engine) has succesfully ported to SDL3 and includes it within its binary packages.  thus it doesnt require a system install of SDL3

https://unvanquished.net/living-in-the-future/


further noting that SDL2 is not deprecated and still receives updates.
so unless there is some burning feature of SDL3 that is beneficial or required

LupertEverett:
Hello there!

I have added SDL3 support to ZWidget very recently, it is in the main repo now!

https://github.com/dpjudas/ZWidget/tree/master/src/window/sdl3

This keeps SDL2 support intact-ish too (if both are present in the system, only the SDL3 backend is built - ofc this can be improved, I just didn't want to deal with linking errors that I encountered when I tried adding SDL3 support a long time ago lol).

SED4906:
I'll probably wait until ZWidget gets updated here with the new changes, from my testing (by replacing `libraries/ZWidget` with the upstream version) it requires some changes to resource loading that shouldn't be lumped in with a pull request meant to update SDL.

src/common/platform/posix/sdl/i_joystick.cpp
```diff
@@ -67,42 +68,44 @@ class SDLInputJoystick: public IJoystickConfig
 public:
 	SDLInputJoystick(int DeviceIndex) :
 	DeviceIndex(DeviceIndex),
-	InstanceID(SDL_JoystickGetDeviceInstanceID(DeviceIndex)),
+	InstanceID(SDL_GetJoysticks(nullptr)[DeviceIndex]),
```
kcat: Might be better to wrap this into a function that checks for a valid index range:

```c++
auto GetJoystickIDFromIndex(int const index) -> SDL_JoystickID
{
    /* 0 is an invalid joystick ID */
    if(index < 0)
        return 0;

    auto count = int{};
    auto *joyids = SDL_GetJoysticks(&count);
    if(joyids && index < count)
        return joyids[index];
    return 0;
}
...
    InstanceID(GetJoystickIDFromIndex(DeviceIndex)),
```

or at least make a function that returns a span:

```c++
auto GetJoystickIDs() -> std::span<SDL_JoystickID>
{
    auto count = int{};
    auto *joyids = SDL_GetJoysticks(&count);
    if(!joyids || count < 0)
        {};
    return std::span{joyids, static_cast<unsigned>(count)};
}
...
    InstanceID(GetJoystickIDs()[DeviceIndex]),
```

the latter of which can get checked with STL hardening enabled (aborting if it's out of range, instead of accessing out of bounds memory).

SED4906:
I said earlier that ZWidget should probably be updated separately, but if ZWidget finds SDL3, it disables SDL2 support implicitly, so updating ZWidget on its own wouldn't work.

RicardoLuis0:
> I said earlier that ZWidget should probably be updated separately, but if ZWidget finds SDL3, it disables SDL2 support implicitly, so updating ZWidget on its own wouldn't work.

yeah but wouldn't any system with sdl3 also have sdl2 as just a wrapper _for_ sdl3? so zwidget uses sdl3, but the game uses sdl2 _on top of_ sdl3, and no issues should happen

kcat:
> yeah but wouldn't any system with sdl3 also have sdl2 as just a wrapper _for_ sdl3?

No, SDL2 can be installed alongside SDL3. As I understand it, the SDL2-to-SDL3 wrapper is more of an experimental thing (or at least, not the default) since SDL3 has differences beyond just the ABI that some existing apps may not play nice with (e.g. some functions not being synchronous like they were, or not flushing or waiting for pending actions like they used to). One of the reasons for SDL3 breaking compatibility was to help tighten up and rework some requirements to work better on newer systems. So having a normal SDL2 build alongside SDL3 may be needed for such apps. 

Also, given how SDL2 and SDL3 have some function/symbol names in common, and how some systems (Linux) handles symbols globally, it can get tricky to manage both libraries being loaded into the same process. If some bit of code tries to call an SDL3 function but ends up calling the SDL2 version (or vice versa), bad things will happen. It's not impossible to deal with, but most code doesn't try, putting it at the mercy of the order the libraries were loaded and which functions end up getting called.

Picking SDL2 or SDL3 is something that should be done at build time. So what probably needs to happen here, if you want ZWidget to be updated independently, is modify ZWidget to support SDL2 and SDL3 (selected at build time), and have UZDoom configure ZWidget to build with SDL2 or SDL3 depending on what UZDoom itself is going to use.

the-phinet:
﻿
﻿

RicardoLuis0:
@SED4906 was this closed by accident?

PR #473, opened 2025-11-14
Disabled all mipmapping on Intel if in None filter modes

### LINKED ISSUE #194: [BUG] Default "unfiltered" textures still look filtered on Intel drivers with Vulkan
### UZDoom version

trunk b04772ef5899ff9e5eca6d8f0771556e2c86556c

Doom 2

Linux x86_64

### Please describe your specific OS version

Mint 22.2

### Relevant hardware info

Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz with i915 graphics

### Have you checked that no other similar issue already exists?

- [x] I have searched and not found similar issues.

### A clear and concise description of what the bug is.

I do not like texture filtering in Doom, and was pleased that UZDoom switched it off by default. However, that default wasn't actually reflected in-game, when I use the Vulkan backend.

Image

When I launch UZDoom with OpenGL, it's fine.

Image

### Steps to reproduce the behaviour.

Explain how to reproduce
1. Start launcher
2. Pick Vulkan
3. Launch

### Your configuration

```ini

```

### Provide a Log

_No response_

Boondorl:
This is an attempt to work around #194. See the discussion in this issue for more details on the problem, but as a quick summary, Intel GPUs always force on filtering when mipmapping is enabled, making the non-standard None texture filter options functionally broken.

As none of us have Intel cards on us to test this, it will need verification. It tries to resolve both menu and texture filtering at the same time, so testing on all 3 backends (OpenGL, GLES, and Vulkan) would be greatly appreciated. This can't really be merged until all 3 cases are resolved.

### DIFF
--- a/src/common/engine/i_interface.h
+++ b/src/common/engine/i_interface.h
@@ -52,6 +52,7 @@ struct SystemCallbacks
 	FConfigFile* (*GetConfig)();
 	bool (*WantEscape)();
 	FTranslationID(*RemapTranslation)(FTranslationID trans);
+	bool (*DisableAnisotropicFiltering)();
 };
 
 extern SystemCallbacks sysCallbacks;
--- a/src/common/rendering/gl/gl_samplers.cpp
+++ b/src/common/rendering/gl/gl_samplers.cpp
@@ -120,12 +120,13 @@ void FSamplerManager::SetTextureFilterMode()
 	}
 
 	int filter = sysCallbacks.DisableTextureFilter && sysCallbacks.DisableTextureFilter() ? 0 : gl_texture_filter;
+	float aniso  = filter <= 0 || (sysCallbacks.DisableAnisotropicFiltering && sysCallbacks.DisableAnisotropicFiltering()) ? 1.0f : gl_texture_filter_anisotropic;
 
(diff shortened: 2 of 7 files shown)


DyNaM1Kk:
Tested on `UHD Graphics 710` using **None (Trilinear)** and **Linear** modes. Looks like it works. Here are the results:

GLES:
None (Trilinear)
Screenshot_Doom_20251114_195718

Linear
Screenshot_Doom_20251114_195735

However, I've noticed that when switching back from Linear to None (Trilinear), this happens:
https://github.com/user-attachments/assets/5f7c2241-7486-4385-8d4f-83d4883afeac

It's almost like the filtering is cached. If I switch between the two modes while the unfiltered frame is visible, the same thing happens and both frames stay filtered.

OpenGL:
None (Trilinear)
Screenshot_Doom_20251114_200500

Linear
Screenshot_Doom_20251114_200506

Vulkan:
None (Trilinear)
Screenshot_Doom_20251114_200536

Linear
Screenshot_Doom_20251114_200540

RicardoLuis0 (approved):
since it appears to be working, all good

PR #474, opened 2025-11-14
Add UZDoom's default paths to an ini from GZDoom

### LINKED ISSUE #425: "Cannot find uzdoom.pk3"
### UZDoom Version

UZDoom 4.14.3

### Game

All

### Operating System

Linux x86_64

### OS Version

Linux Mint

### Hardware

_No response_

### Duplicate Confirmation

### Bug Description

I changed things according to the wiki to transfer from GZDoom, but now that message appears. I don't know where uzdoom.pk3 even is, since I have the appimage.

### Replication

_No response_

### Configuration

```ini

```

### Log

_No response_

TehRealSalt:
Kinda done in a brute-force-y way: it just double-checks all of the system-related default paths exist, instead of only the ones related to the "uzdoom" folder(s). I just wanted to avoid adding even more edge-cases to how paths are defined (the path definitions are already really clustered). Plus GZDoom didn't have XDG stuff so I do think it'd be good to double-check all of the paths. But if need-be, I can make a separate table for the "game name" paths.

Should resolve #425, will need cherry-picked to 4.14.3 as well.

### DIFF
--- a/src/common/utility/configfile.cpp
+++ b/src/common/utility/configfile.cpp
@@ -493,6 +493,37 @@ void FConfigFile::SetValueForKey (const char *key, const char *value, bool dupli
 
 //====================================================================
 //
+// FConfigFile :: EnsureValueForKey
+//
+// Creates a duplicated key, but only if the value is not present
+// across any of the other duplicated keys.
+//
+//====================================================================
+
+void FConfigFile::EnsureValueForKey (const char *key, const char *value)
+{
+	if (CurrentSection != NULL)
+	{
+		FConfigEntry *probe = CurrentSection->RootEntry;
+
+		while (probe != NULL)
+		{
+			if (stricmp(probe->Key, key) == 0
+				&& stricmp(probe->Value, value) == 0)
+			{
+				// Already exists exactly, don't duplicate
+				return;
+			}
+
+			probe = probe->Next;
+		}
+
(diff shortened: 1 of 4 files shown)


TehRealSalt:
(Quick amendment; fixed a rogue line adding a `$HOME` path to Linux configs.)

mc776:
Does this create a uzdoom.ini and populate it with the gzdoom.ini contents, or does it just rely on the gzdoom.ini forever?

(And if the former, this isn't going to overwrite an existing uzdoom.ini or anything, right?)

TehRealSalt:
> Does this create a uzdoom.ini and populate it with the gzdoom.ini contents, or does it just rely on the gzdoom.ini forever?
> 
> (And if the former, this isn't going to overwrite an existing uzdoom.ini or anything, right?)

It doesn't create any files at all, nor does it read anything GZDoom related. This simply fixes not being able to find uzdoom.pk3 on non-Windows systems if you follow the directions on the release notes page to manually copy+rename the GZDoom config to UZDoom's folder.

The worst thing that this branch can do is that it can take your uzdoom.ini and re-add paths that the ini starts with by default if you went in and manually removed those lines. That said, it would only affect people who were playing pre-releases, and it would only happen once.

the-phinet (approved):

PR #479, opened 2025-11-15
Fixed switching from Linear -> None filtering keeping AF level in GLES

Boondorl:


DyNaM1Kk (approved):
Tested on `UHD Graphics 710`. Issue is gone.

TehRealSalt (approved):
I don't have Intel to test but code itself LGTM

Issue #431, opened 2025-11-09
Make include work for at least the most common script lumps (sndinfo, language, etc.)


MartinHowe426:
### Game

All

### Problem Description

The include directive does not work for anything except code script lumps; while one can create **actors** or **zscript** subdirectories for large and complicated projects, one is forced for **language** or **sndinfo** to either have tooling to concatenate them into a single large file (hard to search), or else to give them nonstandard extensions that are the module name (e.g., **sndinfo.quietus.txt**) and thus clutter up the root directory of the archive (a mess).

### Feature Description

It would be good to be able to squirrel away individual files for identifiable modules away in subdirectories to keep the root directory of an archive tidy and aid modularity. One could create separate script directories for each type of script (zscript, sndinfo, etc.) or even per submodule (e.g., sndinfo/weapons/quietus.txt). This would work exactly the same way as splitting zscript up into modules and submodules in an **actors** subdirectory or whatever.

With large projects, this would be a great aid to modularity and also keep the root directory of an archive uncluttered. It would especially make managing large projects easier.

MartinHowe426:
Regarding LANGUAGE: 

The text format is good for small files for individual actors and very easy to read, unless there are a lot of languages which, for most one-person-band projects, there won't be; the format easily lends itself to a "$include" directive in the same manner as SNDINFO.

But the CSV format has no concept of directives and there is no obvious place to put one. Before embarking on that, I'd appreciate suggestions, particularly from the devs due to their internal knowledge of where UZDoom is going in the future.

My initial thoughts are to use a default string with no identifier; this is an illegal combination. Maybe something like:

INCLUDE,,filespec

So "INCLUDE" in the default column, empty key column, and filespec in remarks column, with all other columns blank.

PR #480, opened 2025-11-15
Added a "$include" directive to SNDINFO parser.

MartinHowe426:
Implements #431 for SNDINFO. Tested with multiple directory levels (including archive root) and included files including others.

Please note:

1. My editor removes trailing spaces by default as it keeps files tidy; if unacceptable, please let me know and I'll redo it with the setting turned off.
2. If this accepted, please don't close 431 as there are more script types to do.

### DIFF
--- a/src/sound/s_advsound.cpp
+++ b/src/sound/s_advsound.cpp
@@ -121,6 +121,7 @@ TMap<int, FAmbientSound> Ambients;
 
 enum SICommands
 {
+    SI_Include,
 	SI_Ambient,
 	SI_Random,
 	SI_PlayerSound,
@@ -213,6 +214,7 @@ TMap<int, FString> HexenMusic;
 
 static const char *SICommandStrings[] =
 {
+    "$include",
 	"$ambient",
 	"$random",
 	"$playersound",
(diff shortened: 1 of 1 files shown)


DyNaM1Kk:
This is awesome. Only thing is that I think it should use `#` instead of `$` as that's more standard.

TehRealSalt:
> This is awesome. Only thing is that I think it should use `#` instead of `$` as that's more standard.

The argument for `$` is that all of SNDINFO's commands are prefixed with this. (Personally I can see the arguments for both and don't mind either way)

RicardoLuis0:
yeah, i think `$` is the better way to go, to keep consistency with SNDINFO itself

RicardoLuis0 (approved):

mc776:
I'd much rather # because the $ just makes it look like another part of another actual sound definition instead of doing something else.

MartinHowe426:
FWIW, I used "$" as that is what SNDINFO uses. SNDINFO isn't C, C++, Pascal, or whatever; it defines an entirely declarative language of its own; because there's no inherent distinction between 'directives' and 'commands' in SNDINFO, there's no real need to make it look like C; at least for me.

DyNaM1Kk:
Yes, but SNDINFO didn't have a directive before this PR. Every other lump has their own declarative language and they use `#` for includes.

It's fine though. This argument is not important enough to delay merging this.

Boondorl:
Just a heads up, in the future you should rebase instead of using a merge commit. This won't cause us any issues but it's generally considered better to avoid merge commits like this in PRs.

PR #489, opened 2025-11-16
Interpret and strip comments in response files

### LINKED ISSUE #488: Mod list comments no longer ignored
### UZDoom Version

UZDoom g4.15pre-791-g2e12d8d73

### Game

All

### Operating System

Linux x86_64

### OS Version

CachyOS

### Hardware

Lenovo LNVNB161216 - 11th Gen Intel Core i3-1115G4 - Intel Tiger Lake-LP GT2 [UHD Graphics G4]

### Duplicate Confirmation

### Bug Description

Comments in mod list files are not being interpreted. In GZDoom, they were also not interpreted, but ignoring non-existing files meant they were ignored anyway.

#### Example scenario

modlist.txt:

```
want_this.pk3
#but_not_this.pk3
```

Command:

```
uzdoom [...] -file @modlist.txt
```

#### Expected behavior

Same as GZDoom: `want_this.pk3` is loaded, the other line is ignored.

#### Actual behavior

Line gets interpreted literally, UZDoom complains that it cannot find `#filename` (or `#` if it is spaced).

### Replication

1. Write a mod list file with comments (`#`)
2. Pass the file as a mod list (e.g. `uzdoom -file @my_mod_list.txt`)

### Configuration

```ini
# hdest.fish
#------------
#!/usr/bin/fish
set MAP $argv[1]
set -q SOURCEPORT; or set SOURCEPORT gzdoom
echo Using sourceport $SOURCEPORT # correctly interprets uzdoom
$SOURCEPORT -iwad iwad/doom2.wad -file \
    $MAP \
    mod/audio/pk_doom_sfx_20120224.wad \
    @mod/gameplay/hd-set/hd-set.txt \
    # ^^ the mod list in question
    # [...]
    $argv[2..-1]
# Example actual invocation:
#  uzdoom -iwad iwad/doom2.wad -file mod/audio/pk_doom_sfx_20120224.wad @mod/gameplay/hd-set/hd-set.txt [...]
```

### Log

```
$ SOURCEPORT=uzdoom ./hdest.fish
Using sourceport uzdoom
UZDoom g4.15pre-791-g2e12d8d73 - 2025-11-12 02:11:52 -0600 - SDL version
Compiled on Nov 16 2025
UZDoom version g4.15pre-791-g2e12d8d73
Build: SDL2 version compiled on Nov 16 2025, dated 2025-11-12 02:11:52 -0600
OS: CachyOS, Linux 6.17.7-4-cachyos on x86_64
wad not found: #
```

wallabra:

---

Closes #488.

Interprets `// comments` from response files (`-file @my-response-file.txt`), ignoring their contents, allowing for non-value literate text inside response files.

This fixes the behavior seen in #488.

Example of a response file that should work properly now:

```
mod/yummy/yummy_base.pk3

// Loads my favorite mods :D
mod/yummy/condiment_base.pk3
  mod/yummy/ketchup.pk3
  mod/yummy/mustard.pk3
//--

// Extra ingredients
mod/yummy/extra_base.pk3
  mod/yummy/pickles.pk3
  // - NOTE: My sister is allergic to onions, so comment the following out
  // - only when she's at home ... yes, AT home - I'm evil >:)
  //mod/yummy/onions.pk3
//--
```

### DIFF
--- a/src/m_misc.cpp
+++ b/src/m_misc.cpp
@@ -17,11 +17,15 @@
 **
 */
 
+#include <cctype>
+#include <cstdint>
 #include <sys/stat.h>
 #include <sys/types.h>
+#include <sstream>
 #include <stdlib.h>
 #include <errno.h>
 #include <time.h>
+#include <valarray>
 
 #include "r_defs.h"
 
@@ -78,6 +82,77 @@ static size_t ParseCommandLine (const char *args, int *argc, char **argv);
 //
 //---------------------------------------------------------------------------
 
+// Change this if you want to use a different token to open comments inside a response file.
+static const char *const COMMENT_TOKEN = "#";
+
+static const uint16_t UF_STRIP_WHITESPACE = 1;
+static const uint16_t UF_SKIP_EMPTY = 2;
+static const uint16_t UF_ALLOW_TRAILING = 4;
+
+static std::string Uncomment(const char* from, const uint16_t uncomment_flags = 0)
+{
+	std::stringstream outstream;
+
+	{
+		std::stringstream instream(from);
+		std::string line;
+
+		const bool strip_whitespace = (uncomment_flags & UF_STRIP_WHITESPACE) != 0;
+		const bool skip_empty = (uncomment_flags & UF_SKIP_EMPTY) != 0;
+		const bool allow_trailing = (uncomment_flags & UF_ALLOW_TRAILING) != 0;
+
+		while (std::getline(instream, line)) {
+			// skip comments
+			auto cidx = line.find(COMMENT_TOKEN); // comment token position
+
+			if (cidx == 0) {
+				// starts with token
+				continue;
+			}
+
+			// comment must be first non-whitespace token
+			if (!allow_trailing && cidx != std::string::npos) {
+				size_t start = 0;
+				while (start < cidx && std::isspace(static_cast<unsigned char>(line[start]))) start++;
+				if (start != cidx) cidx = std::string::npos;
+			}
+
+			// case 1: token is in the middle, strip until cidx
+			// case 2: token not present, cidx == npos, aka max size value, so a noop
+			// (faster than adding an IF branch)
+			const std::string_view substring = std::string_view(line).substr(0, cidx);
+
+			// strip whitespace around line
+			size_t start = 0;
+			size_t end = substring.size();
+
+			// skip actually empty lines
+			if (skip_empty && end == 0) {
+				continue;
+			}
+
+			if (strip_whitespace) {
+				// find leading
+				while (start < end && std::isspace(static_cast<unsigned char>(substring[start]))) start++;
+				// trim trailing
+				while (end > start && std::isspace(static_cast<unsigned char>(substring[end - 1]))) end--;
+			}
+
+			const std::string_view trimmed = substring.substr(start, end - start);
+
+			// skip functionally empty lines
+			if (skip_empty && trimmed.size() == 0) {
+				continue;
+			}
+
+			// spit good value
+			outstream << trimmed << ' ';
+		}
+	}
+
+	return outstream.str();
+}
+
 void M_FindResponseFile (void)
 {
 	const int limit = 100;	// avoid infinite recursion
@@ -94,6 +169,7 @@ void M_FindResponseFile (void)
 		{
 			char	**argv;
 			FileSys::FileData file;
+			std::string uncommented;
 			int		argc = 0;
 			size_t	argsize = 0;
 			int 	index;
@@ -112,7 +188,8 @@ void M_FindResponseFile (void)
 				{
 					Printf ("Found response file %s!\n", Args->GetArg(i) + 1);
 					file = fr.ReadPadded(1);
-					argsize = ParseCommandLine (file.string(), &argc, nullptr);
+					uncommented = Uncomment(file.string(), UF_SKIP_EMPTY | UF_STRIP_WHITESPACE);
+					argsize = ParseCommandLine (uncommented.c_str(), &argc, nullptr);
 				}
 			}
 			else
@@ -124,7 +201,7 @@ void M_FindResponseFile (void)
 			{
 				argv = (char **)M_Malloc (argc*sizeof(char *) + argsize);
 				argv[0] = (char *)argv + argc*sizeof(char *);
-				ParseCommandLine (file.string(), nullptr, argv);
+				ParseCommandLine (uncommented.c_str(), nullptr, argv);
 
 				// Create a new argument vector
 				FArgs *newargs = new FArgs;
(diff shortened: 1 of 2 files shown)


the-phinet:
Lol I love the example file 😂

I would prefer `//` as a comment, due to `#` being valid as part of a filesystem path

wallabra:
I'll change the comment token to `//`.

wallabra:
Sorry for the force pushes, just wanted to fix my SSH signing setup.

RicardoLuis0:
btw, don't put references/links to things on github in commit messages, that'll generate a ton of spam

wallabra:
> btw, don't put references/links to things on github in commit messages, that'll generate a ton of spam

wait, did something happen that I'm unaware of?

the-phinet:
> Sorry for the force pushes, just wanted to fix my SSH signing setup.

Don't worry about it at all. This is still a pr lol

> wait, did something happen that I'm unaware of?

Github adds a link to the commit from the issue you a referencing. I honestly think this is a very good thing, and don't see a reason not to do it.

image

RicardoLuis0:
> > Sorry for the force pushes, just wanted to fix my SSH signing setup.
> 
> Don't worry about it at all. This is still a pr lol
> 
> > wait, did something happen that I'm unaware of?
> 
> Github adds a link to the commit from the issue you a referencing. I honestly think this is a very good thing, and don't see a reason not to do it.
> 
> image

the issue is that every force push, every rebase, etc, will create a new 'reference' on the issue, and spam endlessly, it's something on github's end that we have no control over unfortunately

wallabra:
> the issue is that every force push, every rebase, etc, will create a new 'reference' on the issue, and spam endlessly, it's something on github's end that we have no control over unfortunately

Ouch, well I hope a squash merge would sort that out

wallabra:
Now that it uses slashes as the comment token, it should be fine. Should I also update the OP?

the-phinet:
Sorry, I (we) haven't had time to look at this.

> Now that it uses slashes as the comment token, it should be fine. Should I also update the OP?

That's not required. But one thing that is required is adding this to the OP:

```
```

We now require agreeing to our [CONTRIBUTING.md](https://github.com/UZDoom/UZDoom/blob/trunk/CONTRIBUTING.md) file, in order for code to be merged. Since your pr was made before this change, the pr template was not in place for that checkbox to be added.

If you're good with the terms laid out in the file, please click the little checkbox that the above text adds.

wallabra:
OK! Done. Thank you!

the-phinet (dismissed):
Sorry, I entirely forgot to look at this.

The code looks fine and it works on my end.

Only thing left is to add your name to the bottom of the CONTRIBUTORS file :D

the-phinet (dismissed):

wallabra:
Oop! Thanks for reminding me!

kcat:
Not sure I like using `//` since that's a valid path separator. `//home/user`, `//home//user`, `/home//user`, `/home/user`, are all valid to reference the same path.

Perhaps ';' would be better, since I don't think Windows allows that character on paths (it's used as a separator for the `PATH` env var), and I'd be surprised if anyone really uses it on Linux or Mac. Could limit potential misidentification by recognizing ';' as a comment only when it's the first character on a line, or if it's preceded by whitespace?

the-phinet:
> Not sure I like using `//` since that's a valid path separator. `//home/user`, `//home//user`, `/home//user`, `/home/user`, are all valid to reference the same path.

I disagree. POSIX defines `//` as implementation-specific. Some systems use it as the root folder of a networked filesystem. 

I think it's reasonable to expect the user to use only the canonical path of a file, where `//` would never occur.

RicardoLuis0:
could add an extra step to the preprocessing to allow for escapes? ex. instead of `//` for a path with double slashes, you'd do `\/\/`

the-phinet:
> could add an extra step to the preprocessing to allow for escapes? ex. instead of `//` for a path with double slashes, you'd do `\/\/`

You could, but I don't think it's needed. The only time a user would have double slashes in the path to begin with is if they failed to copypaste something properly. There's no unix path that requires double slashes. If they go through the effort of escaping the path, they might as well delete the extra characters

RicardoLuis0:
> could add an extra step to the preprocessing to allow for escapes? ex. instead of `//` for a path with double slashes, you'd do `\/\/`

though i guess it'd make using actual `\` in paths confusing since you'd need to do `\\` for each `\`, but idk, even windows can read `/` in paths fine

kcat:
> I disagree. POSIX defines `//` as implementation-specific. Some systems use it as the root folder of a networked filesystem.

Shouldn't that then be a problem, if someone wants to use such a folder in the response file?

> I think it's reasonable to expect the user to use only the canonical path of a file, where `//` would never occur.

I don't really agree, especially if the response file is generated in a way that can result in a `//`.

I made a quick search and found this: https://isocpp.org/files/papers/P3051R0.html

I'm not suggesting we adopt that, but it does provide examples which show Clang and Intel use `#` for comments in their response files (GCC and MSVC don't specify what, if anything, is for comments).

the-phinet:
> Perhaps ';' would be better, since I don't think Windows allows that character on paths (it's used as a separator for the PATH env var), and I'd be surprised if anyone really uses it on Linux or Mac. Could limit potential misidentification by recognizing ';' as a comment only when it's the first character on a line, or if it's preceded by whitespace?

Semicolons are valid characters in a path for all (afaik) operating systems. The only invalid printing character in Linux is `/`, where it is reserved for being the path separator. Windows has a bunch of other reserved chars

If anything, a `#` at the beginning of a line (or preceded by whitespace?) is better than a semicolon

the-phinet:
> Shouldn't that then be a problem, if someone wants to use such a folder in the response file?

If they want to use a long path than needed, I guess. 

If the network boundary is `/local/networked`, `//` is a 'ignore and start over on the boundary' marker. So this path: `/local/networked/foo/bar/baz//a/b/c` is actually `/local/networked/a/b/c`

Keep in mind that the behavior of `//` is very system dependent. I don't think we should worry about niche use-cases like this

kcat:
> If anything, a # at the beginning of a line (or preceded by whitespace?) is better than a semicolon

I'm fine with that. Could maybe also support hex codes in case a real `#` is needed; a `%` followed by two hex characters, so `Some Mod #2.pk3` could be written as `Some Mod %232.pk3`. A `%` character that is not followed by two hex characters, or `%%`, would be a normal single `%`.

the-phinet:
Alrighty, sorry for the delay. While I still think `//` makes more sense (and I wish that was brought up earlier when the PR was still being worked on), `#` is common enough and any path issues can be solved with quoting. I've swapped it back to be `#`, but only allowing full-line comments. Hopefully that's an implementation that everyone is happy with.

src/m_misc.cpp
```diff
@@ -78,6 +82,77 @@ static size_t ParseCommandLine (const char *args, int *argc, char **argv);
...
+			size_t end = substring.size();
+
+			// skip actually empty lines
+			if (skip_empty && end == 0) {
+				continue;
+			}
+
+			if (strip_whitespace) {
+				// find leading
+				while (std::isspace(static_cast<unsigned char>(substring[start])) && start < end) start++;
```
kcat: Might be a good idea to check `start < end` first, since if `start == end`, this accesses the past the end of the view. Practically speaking, this is fine since we know it's backed by a null-terminated `std::string`, so there will be something there (either the comment token or a null char), but hardened STL implementations may still throw a fit/assert due to `start < size()` failing.

the-phinet: Good call. Thanks!

kcat (approved):

PR #490, opened 2025-11-16
Added auto compatibility option

Boondorl:
This attempts to fix more map-breaking compatibility options by enabling certain behaviors in intelligent ways based on both game and map type. To start, auto missile clipping has been added, making it so in vanilla Doom maps specifically missiles will pass through things like torches, matching the original game.

I'm aware that the UDMF handling on the map type is technically wrong as it doesn't account for namespaces, but this can be fixed at a later time as I imagine even with DSDA UDMF few maps have it at the moment, relatively speaking. The only real upgrade needed here would be to check against the ZDoom namespace since other namespaces are probably for ports that use vanilla behaviors by default. Hexen is also not included in the test since it had Actor over Actor and thus needed correct Actor heights by default (this should help catch any pre-UDMF maps made for ZDoom).

### DIFF
--- a/src/g_levellocals.h
+++ b/src/g_levellocals.h
@@ -48,6 +48,8 @@
 #include "p_visualthinker.h"
 #include <memory>
 
+EXTERN_CVAR(Bool, sv_autocompat)
+
 struct FGlobalDLightLists
 {
 	//TODO add TSet and switch from TMap to TSet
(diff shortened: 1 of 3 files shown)


Boondorl:
This is a pretty big blocking PR so I'd highly appreciate if someone could take a look over this and approve it

the-phinet (approved):
This looks correct to me, and seems to work as intended. Sorry for the delay in looking at this

the-phinet:
﻿
﻿

Issue #498, opened 2025-11-17
Move mod options to their own submenu


MajorCooke:
Mod options being at the bottom and mingling messily with the bottom-most options is quite ugly.

I suggest putting them in their own submenu where they're out of the way inside the Options menu where they're separated from the "Browse" stuff. Can just call it "Mod Options".

Alternatively, it could also be opt-in with a `ModOptionMenu` declaration inside of a GameInfo or something? But either way, [this doesn't look good.](https://puu.sh/KE20L/2861fcf542.png)

Enjay001:
Yep, some sort of neat "mod options" menu would be nice.  If you know xZDoom, then you know to expect mod options to appear down the bottom of the options list, but even then its easy to miss unless you are looking for it.  I also agree that the current system is not pretty.  (Either is the options menu font, IMO, but that's a separate issue.)

I like the idea of an opt-in declaration in gameinfo.  That way there could be no mod options menu entry unless a mod specifically asks for one.  That would be better than having an empty-by-default mod options entry and it appearing would further highlight that there is now something that can be adjusted (unless a user always plays with a mod (or mods) that also have options, in which case the menu option would always be there for them).

Boondorl:
I'm not sure if there's anything we could do about this given how the options menu is structured. The internal MENUDEF doesn't do anything bespoke, it's just the same as all other MENUDEFs. The result is we'd probably have to add some pretty gross hacks to the backend to determine what's a "mod feature", not to mention someone might intentionally want their option at the bottom of the menu. It would also be a mess for standalone ports because they aren't even guaranteed to have the same kind of options menu, basically forcing all of them to either remove this code or rewrite it depending on how they implement their menus.

I think adding a Mods submenu is really the only elegant option here. It can just be a special item that only gets added if it has any items, otherwise being invisible. It wouldn't retroactively fix mods but it would give a way to officially add them going forward.

MajorCooke:
> I think adding a Mods submenu is really the only elegant option here. It can just be a special item that only gets added if it has any items, otherwise being invisible. It wouldn't retroactively fix mods but it would give a way to officially add them going forward.

Exactly. And I would like to suggest that it goes up top, please and thank you. Or right under Customize Controls, that works too. 

Or, a "Configure Mods" or "Mod Options" in the main menu would also be nice.

EdHerdman:
[...]

PR #505, opened 2025-11-17
Added optional gameplay mod submenu to options

Boondorl:
Allows modders to add their mod settings into this menu rather than appending it to the very bottom by adding to GameplayModOptions. Also adds OptionalSubmenu to MENUDEF, a  submenu label type that only displays if its menu has any content to do so.

Implements #498 in a way that I consider unobtrusive to how MENUDEF works. I don't think adding backwords compat for this is possible without potentially messing up a bunch of other menus and how they work.

### DIFF
--- a/wadsrc/static/menudef.txt
+++ b/wadsrc/static/menudef.txt
@@ -362,6 +362,7 @@ OptionMenu "OptionsMenu" protected
 	StaticText " "
 	Submenu "$OPTMNU_PLAYER",			"NewPlayerMenu"
 	Submenu "$OPTMNU_GAMEPLAY",			"GameplayOptions"
+	OptionalSubmenu "$OPTMNU_ADDONS",	"AddOnOptions"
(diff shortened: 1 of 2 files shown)


River-Salmon (dismissed):

Boondorl:
Decided to rename the menu to Add-Ons since I agree as per the issue discussion that this is less ambiguous

Boondorl:
Honestly, the more I think about this and the more I see it discussed, the less I like the idea. I think something different is needed here. That or it should simply be left up to mods to handle, because I don't know if it's really in the best interest of the engine to be trying to manage something like this. Options are already robust enough that submenus can be added anywhere within them, so maybe this should be revisited with the menu rework

MajorCooke:
Honestly I've always been more of a fan of the main menu itself having "Mod Options". I wouldn't go with "addons" because that's what's implied by the re-releases where you can download stuff from in game.

PR #509, opened 2025-11-18
Added missing scale and alpha interpolations

Boondorl:
Fixes missing cases and adds support in the software renderer with the new INTERPOLATEALPHA/SCALE flags.

### DIFF
--- a/src/am_map.cpp
+++ b/src/am_map.cpp
@@ -2917,7 +2917,7 @@ void DAutomap::drawPlayers ()
 			continue;
 		}
 
-		if (p->mo->Alpha < 1.)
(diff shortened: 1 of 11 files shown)


Boondorl:
Tested it. Models + software + the automap seem to be working

TehRealSalt (approved):

MajorCooke:
Please make sure DONTINTERPOLATE is also respected!

Boondorl:
I think that'll need to be a separate discussion. Angle interpolation doesn't respect it either and the wiki says it only applies to positional interpolation

MajorCooke:
[Alright then.](https://github.com/UZDoom/UZDoom/issues/615)

PR #527, opened 2025-11-19
Enable a default crosshair with a more sensible scale size for modern displays

nashmuhandes:
My rationale is that mouselook is already enabled by default _anyway_, therefore not having a crosshair to go with it by default comes off as a missing feature to me.

Here's the tweaked settings on a 1080p display. I think the chosen `crosshairscale` size looks decent here.

Screenshot_Doom_20251119_210057

### DIFF
--- a/src/common/statusbar/base_sbar.cpp
+++ b/src/common/statusbar/base_sbar.cpp
@@ -59,7 +59,7 @@ IMPLEMENT_CLASS(DHUDFont, false, false);
 
 CVAR(Color, crosshaircolor, 0xff0000, CVAR_ARCHIVE);
 CVAR(Int, crosshairhealth, 2, CVAR_ARCHIVE);
-CVARD(Float, crosshairscale, 0.5, CVAR_ARCHIVE, "changes the size of the crosshair");
+CVARD(Float, crosshairscale, 0.2, CVAR_ARCHIVE, "changes the size of the crosshair");
 CVAR(Bool, crosshairgrow, false, CVAR_ARCHIVE);
 
 CUSTOM_CVARD(Float, hud_scalefactor, 1.f, CVAR_ARCHIVE, "changes the hud scale")
--- a/src/g_statusbar/shared_sbar.cpp
+++ b/src/g_statusbar/shared_sbar.cpp
@@ -123,7 +123,7 @@ EXTERN_CVAR(Float, hud_scalefactor)
 EXTERN_CVAR(Bool, hud_aspectscale)
 
 CVAR (Bool, crosshairon, true, CVAR_ARCHIVE);
-CVAR (Int, crosshair, 0, CVAR_ARCHIVE)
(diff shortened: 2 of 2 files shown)


Boondorl:
Not a fan of the crosshair being on by default tbh, but I do think a more sensible default scale should be used

RicardoLuis0:
yeah, we really need an onboarding settings screen for things like this

the-phinet (approved):
I like this change a lot

nashmuhandes:
The game has mouselooking on by default, and I feel crippled being able to mouselook but not have a crosshair on the screen. If this was no mouselooking like KexDoom, or vanilla or whatever, then fine, no crosshair needed

Boondorl:
I'm probably biased here because most of my FPS experience is on games with no crosshair, so it's never really bothered me. I think with the improved crosshairs it'll look better but I find the current set kinda tacky? I think it's the vibrant green color, it just clashes hard unlike the ones Quake ports use. Which I get is the point, but maybe we should have some kind of "inverse color underneath" behavior that some games use to keep visibility while retaining high contrast

mc776:
> but I find the current set kinda tacky? I think it's the vibrant green color, it just clashes hard unlike the ones Quake ports use

I tend to change it to red right away, though for FPSes/mods with more complex options I'll do red a blue-cyan side bits that can still show up in case what I'm pointing at is also red.

Red dot seems to me to be the most neutral and unobtrusive default.

~~also ☺ and ♥️ should be crosshair options~~

kcat:
> Red dot seems to me to be the most neutral and unobtrusive default.

By default(?) the crosshair color reflects health. Red is used to indicate low health, and green is for high/good health, which isn't an uncommon color scheme for health indication. Changing that would be unnecessarily confusing for players already using it.

mc776:
I think my feelings about health crosshairs are similar to Boon's about the crosshair generally...

src/common/statusbar/base_sbar.cpp
```diff
@@ -59,7 +59,7 @@ IMPLEMENT_CLASS(DHUDFont, false, false);
 
 CVAR(Color, crosshaircolor, 0xff0000, CVAR_ARCHIVE);
 CVAR(Int, crosshairhealth, 2, CVAR_ARCHIVE);
-CVARD(Float, crosshairscale, 0.5, CVAR_ARCHIVE, "changes the size of the crosshair");
+CVARD(Float, crosshairscale, 0.2, CVAR_ARCHIVE, "changes the size of the crosshair");
```
the-phinet, line 62: If #538 gets merged, this looks slightly better set to `0.3`. At 1080p, it's basically pixel perfect

nashmuhandes: Noted, will update PR

nashmuhandes:
Will have to redo this PR due to a technical mess-up on my end

PR #528, opened 2025-11-19
Louden UI sounds to match the overall game volume.

nashmuhandes:
https://github.com/user-attachments/assets/3be54a2a-1029-4835-916b-9c7622244699

Due to the fixed audio volume issues in UZDoom (a welcome change), unfortunately now the menu sounds are too quiet.

Listen to the volume differences when the menu sounds are played between the old and new values. The latter has more consistent volume between gameplay and menu, and was in fact how it originally was in GZDoom


Boondorl:
I've been thinking about this one and tbh I don't know if I like how loud the menu is after the fix. If the sounds were specifically made for menuing I think it would be fine but the current ones are extremely obnoxious since it's gun fire and level geometry moving. imo we need a completely new set of sounds before I'd ok cranking the volume up all the way

the-phinet:
I'm in agreement with Boon here. I don't think this should be changed, without changing the sounds themselves, too.

RicardoLuis0:
yeah, the menu was too loud before lol, i always turned it down

PR #531, opened 2025-11-19
Added interpolated movement for dynamic lights

Boondorl:
Interpolation for light movement along the xy axes in particularly is tricky because of the linking cost. While there have been significant improvements to this in 5.0, doing it every single frame is still expensive. As such, updating light xy position in real time has never been easy.

This new strategy makes interpolation opt-in and tries to instead strategically link to all possible lit sectors across the area of the light's movement. When an xy movement is done, it gets the center point of the movement and increases the light's radius to cover the full span. This avoids most sectors potentially being missed from the average point during traversal while still only needing to link once per tick, and only if an xy movement was detected. The interpolation data is stored in the light's velocity to avoid branching during render calculations. To avoid keeping the radius larger than expected, after an interpolated movement the light is relinked to its real position and size. This is only done if another xy movement isn't queued up, ensuring only one link is ever done at most per step.

TODO:
-Add proper support for static portals
-Better optimize performance during rendering (can something other than the Vel strat be used here?)
-Remove branches in ticking? Will need verifying with lots of lights at once to ensure this isn't a waste of time
-Checking if the increased linking size has a major impact with lots of moving lights
-Make sure sector-heavy maps aren't seeing significant decreases in performance
-Don't use r_viewpoint.TicFrac? Not sure how bad this is for cache performance

### DIFF
--- a/src/playsim/a_dynlight.cpp
+++ b/src/playsim/a_dynlight.cpp
@@ -376,58 +376,99 @@ void FDynamicLight::Tick()
 //==========================================================================
 void FDynamicLight::UpdateLocation()
 {
-	double oldx= X();
-	double oldy= Y();
+	if (!IsActive())
+		return;
+
+	const DVector3 oldpos = Pos;
 	float oldradius = radius;
 
-	if (IsActive())
+	AActor *target = this->target;	// perform the read barrier only once.
+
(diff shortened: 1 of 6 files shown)


the-phinet:
﻿
﻿

PR #538, opened 2025-11-20
Remove unnecessary AA from crosshairs, convert to png

SeventhSentinel:
- Crosshairs now have crisp pixels, making them appear less blurry and making it easier to see what's behind them
- Crosshairs are now in png format. This nearly doubles the file sizes in some cases, but we're talking about files that are under 200 bytes in size. We're not on dial-up connections anymore, so this seems like a small price to pay for being able to more easily edit them in the future
- Crosshair 7 is now a single pixel to make it play nicer with scaling

Before/After

Screenshot_Doom_20251119_235818

Screenshot_Doom_20251119_235730


mc776:
I've always wondered why the cross crosshair was so thick like that...

How do the non-square-angle crosshairs like the circle and triangle look in this?

SeventhSentinel:
Those two didn't have unnecessary AA and are therefore unchanged aside from being converted to png.

nashmuhandes:
Although not very likely, the _only_ potential I could think of if some obscure mod referenced the crosshair images via their long name (`"wadsrc/static/graphics/xhairb1.imgz"` instead of `XHAIRB1`).

Otherwise, the assets themselves look good

kcat:
Would something like SVG make more sense? Since UZDoom can run at anything from 320x200 to 8k and beyond, and the crosshairs themselves are fairly simple shapes, a vector-based format would scale better (particularly circles and rounded edges) and could probably offer a bit more customization for things like line thickness.

the-phinet:
> Although not very likely, the only potential I could think of if some obscure mod referenced the crosshair images via their long name ("wadsrc/static/graphics/xhairb1.imgz" instead of XHAIRB1).

I don't think this is a concern specifically because xhairb2.imgz was already removed in favour of a png in the past

If this does become an issue, can *.imgz fallback to *.png in code?

the-phinet (approved):
Played with these for a while, I love them. So much better. Especially the dot one.

nashmuhandes:
> > Although not very likely, the only potential I could think of if some obscure mod referenced the crosshair images via their long name ("wadsrc/static/graphics/xhairb1.imgz" instead of XHAIRB1).
> 
> I don't think this is a concern specifically because xhairb2.imgz was already removed in favour of a png in the past
> 
> If this does become an issue, can *.imgz fallback to *.png in code?

What I mean is, it is possible through scripting (either ACS or ZScript), in the various image drawing functions, to use the full path name instead of the 8-character name for the texture. There's no good way to heuristically fix this because it's the wild west in modding land.

But I also did say it is quite highly unlikely. I suppose just take what I said as "I suspected this might have happened" _in case_ a mod breaks in future. Not as a mark of disapproval. :P

nashmuhandes (approved):

PR #546, opened 2025-11-21
Fully client-sided sound updating

Boondorl:
Also fixes the spycam breaking the listener and sound sequencing logic. Previously sound sequences were only updated once per engine tick ran always, but now use the number of local unpaused ticks that have occurred to know how much to advance. This should hopefully do a better job keeping pace when the engine falls behind, ensuring sounds are still played back correctly (audio is rendered separately from the game loop so they will continue to play back even while the game is lagging out).

### DIFF
--- a/src/d_main.cpp
+++ b/src/d_main.cpp
@@ -706,28 +706,25 @@ CUSTOM_CVAR (Int, dmflags2, 0, CVAR_SERVERINFO | CVAR_NOINITCALL)
 	if ((self & DF2_NO_AUTOMAP) && automapactive)
 		AM_Stop ();
 
+	// Revert our view to our own eyes if spying someone else.
+	if ((self & DF2_DISALLOW_SPYING) && players[consoleplayer].camera != players[consoleplayer].mo)
+	{
+		// The player isn't looking through its own eyes, so make it.
+		players[consoleplayer].camera = players[consoleplayer].mo;
+
+		S_UpdateSounds(players[consoleplayer].camera, 0);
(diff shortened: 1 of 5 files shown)


src/d_main.cpp
```diff
@@ -706,29 +706,26 @@ CUSTOM_CVAR (Int, dmflags2, 0, CVAR_SERVERINFO | CVAR_NOINITCALL)
 	if ((self & DF2_NO_AUTOMAP) && automapactive)
 		AM_Stop ();
 
+	// Revert our view to our own eyes if spying someone else.
+	if ((self & DF2_DISALLOW_SPYING) && players[consoleplayer].camera != players[consoleplayer].mo)
```
River-Salmon, line 710: Nit: can we rename "self" to something else since that usually implies "ptr to self"?

src/sound/s_doomsound.cpp
```diff
@@ -832,17 +832,19 @@ static void S_SetListener(AActor *listenactor)
 // Updates music & sounds
 //==========================================================================
 
-void S_UpdateSounds (AActor *listenactor)
+void S_UpdateSounds (AActor *listenactor, int tics)
```
River-Salmon, line 835: Nit: stronger typing for tics, uint or size_t or something like that to make it clearer it's a count

src/d_main.cpp
Boondorl: Probably not. This is baked into the CVar macro itself, every single custom one would need to be changed

TehRealSalt (approved):

PR #554, opened 2025-11-21
Defaulted WorldPaused to false

Boondorl:
Checking if the game is lagged out should only be needed for niche scenarios since this will mostly be used with the HUD and client-side Thinkers, both which don't care about this.


RicardoLuis0:
> Checking if the game is lagged out should only be needed for niche scenarios since this will mostly be used with the HUD and client-side Thinkers, both which don't care about this.

ah, didn't realize it was that expensive, should probably cache it in the render structs since the way i did it it'll be checked for every model renderered

TehRealSalt (approved):

Issue #517, opened 2025-11-18
Option to change screenshot folder location


JOELwindows7:
### Game

All

### Problem Description

Currently, in GZDoom & hence UZDoom, the Screenshot with `PrntScr` function saves your screenshot in your `My Pictures\Screenshots\UZDoom`, depends on your Windows registry of which user account `Picture` folder would be and such, or anything like that of `XDG_PICTURE` and such.  
**There's no such option to change where location would you able to save the screenshot to where**. Sometimes it is desirable to be able to save screenshot on other place without having to edit your system configuration / registries, e.g. to better sort your screenshot for mod development, streaming, review journalism, etc. reason.

### Feature Description

Provide option to change output directory of where your screenshot image be placed at. When the option was chosen, a Select Folder Dialog opens where you can choose new location for where the next screenshot will be.

And a Reset to Default option to reset the custom location back to your system default, i.e. config or registry.

RicardoLuis0:
there is, just pass `-shotdir` to the launch parameters

RicardoLuis0:
though it certainly could be added as a cvar as well

mc776:
CVar? or a path definition?

EdHerdman:
A configuration UI allowing path definitions would make sense (most emulators have them) and it seems a few Issues are related to this idea.

JOELwindows7:
> CVar? or a path definition?



> A configuration UI allowing path definitions would make sense (most emulators have them) and it seems a few Issues are related to this idea.

Yes. A usual setting thing. A setting option that lets you sets a *Folder path* by dialog, save what its dialog called back to a String CVAR like `MNU_SCREENSHOTPATH`. and so, that CVAR be defined like

```cvarinfo
nouser string MNU_SCREENSHOTPATH = ""
```

it can be e.g. `~/Picture/Screenshots/UZDoom`, `/mnt/CCTV_Harddisk_OBS/UZDoom/Screenshot`, etc. if it's empty, then it automatically goes back to default.

mc776:
TIL on Linux it's now ~/.local/share/games/uzdoom/screenshots/ instead of the much shorter and more convenient and already existing ~/.config/uzdoom/screenshots. Sorry but this is a huge regression as far as not cluttering up someone's hard drive with folders goes.

the-phinet:
Screenshots are not configs, so they do not belong in a config dir. They were moved to the canonical program data dir. 

[...]

PR #565, opened 2025-11-22
Update Linux/Haiku screenshot path

the-phinet:
Previously all gzdoom files lived in `~/.config/gzdoom`, so screenshots lives in `~/.config/gzdoom/screenshots`

ecf42b81699571ee4bca52c66f1b5441e5346e19 updated paths to follow the XDG spec.

This moved all non-config data to `$XDG_DATA_HOME/games/uzdoom`, which is `~/.local/share/games/uzdoom` on most distros. As a side effect of this change, screenshots became even more burried than before.

This was noticed in #517, and it was suggested to use `~/Pictures` (much like window's use of `My Pictures`)

This moves screenshots to `$XDG_PICTURES_DIR/Screenshots/UZDoom` (`~/Pictures/Screenshots/UZDoom` on most distros). This is the same location both Gnome and KDE place screenshots, too.

### DIFF
--- a/src/common/platform/posix/unix/i_specialpaths.cpp
+++ b/src/common/platform/posix/unix/i_specialpaths.cpp
@@ -75,6 +75,7 @@ DEFGETPATH(Data, "XDG_DATA_HOME", "$HOME/config/non-packaged/data");
 DEFGETPATH(Config, "XDG_CONFIG_HOME", "$HOME/.config");
 DEFGETPATH(Cache, "XDG_CACHE_HOME", "$HOME/.cache");
 DEFGETPATH(Data, "XDG_DATA_HOME", "$HOME/.local/share");
+DEFGETPATH(Pictures, "XDG_PICTURES_DIR", "$HOME/Pictures");
 #endif
 #undef DEFGETPATH
 
@@ -197,7 +198,11 @@ FString M_GetConfigPath(bool for_reading)
 
 FString M_GetDocumentsPath()
 {
+#ifdef __HAIKU__
+	return FStringf("%s/" GAMENAMELOWERCASE "/", GetConfigPath());
+#else
 	return M_GetAppDataPath(false) + "/";
+#endif
 }
 
 
@@ -211,7 +216,13 @@ FString M_GetDocumentsPath()
 
 FString M_GetScreenshotsPath()
 {
-	return M_GetDocumentsPath() + "screenshots/";
+#ifdef __HAIKU__
+	static FString path = M_GetDocumentsPath() + "screenshots";
+#else
+	static FString path = FStringf("%s/Screenshots/" GAMENAME, GetPicturesPath());
+#endif
+	path = NicePath(path.GetChars());
+	return path;
 }
 
 //===========================================================================


the-phinet:
@Peppersawce Is this path correct for Haiku? As far as I can tell, there is no canonical pictures dir, so screenshots get dumped in the home folder.

Peppersawce:
'$XDG_PICTURES_DIR' is undefined in Haiku.
GZDoom in Haiku uses `~/config/settings/GZDoom/screenshots`, which is an ok location for it.

Generally speaking I think the proper location would be `~/config/non-packaged/data/UZDoom/screenshots`, `~/config/non-packaged/data/appname` is where users are meant to find editable data or output (`~/config/data` is read-only) and `~/config/settings/appname` is where configuration files are stored (.cfg, saves etc.)

Dropping the screenshots directly in Home could be ok too tbh. That's the default folder for system screenshots.

the-phinet:
I'm cool with whatever you think is most correct/user friendly. This change is being made because a user complained that screenshots were too deeply nested now.

What of these do you think is best?
- in data dir
  - `~/config/non-packaged/data/uzdoom/screenshots/Screenshot_Doom_20251122_184223.png`
- in non-standard screenshot folder
  - `~/Screenshots/UZDoom/Screenshot_Doom_20251122_184223.png`
- directly in home folder
  - `~/Screenshot_Doom_20251122_184223.png`

Peppersawce:
These are the default save locations for the native Screenshot app:
screenshotscreenshot
Desktop is `~/Desktop` and Artwork folder is `~/config/non-packaged/data/artwork`

Peppersawce:
>     * in data dir
>       
>       * `~/config/non-packaged/data/uzdoom/screenshots/Screenshot_Doom_20251122_184223.png`
That's where the users would expect them to be usually so go with that one.

mc776:
> the users

Who!? I would never in a million years think to look for a folder like that.

Please just go with "."

Peppersawce:
"." is not consistent with other apps' behavior, you wouldn't use "." on Linux either.
`$XDG_DATA_HOME` points at `/boot/home/config/non-packaged/data` on Haiku, that is the standard.

I also asked in the users' chat and nobody commented negatively about it so that's an accepted path by "users".

If you use "." I'll have to patch it out on my side, like, 100%.

the-phinet:
> Please just go with "."

Dumping screenshots in the current working directory by default would be such an incredibly bad idea. The working directory is not defined consistently depending on how the user installs the application, and is also not guaranteed to be writable.

jscipione:
Haiku dev here: ~/config/settings/UZDoom/Screenshots/ is the correct directory on Haiku for packaged apps to store data to. (capitalized Screenshots is ok) non-packaged is a space for the user to install their own non-packaged apps, fonts and other data. Haiku packaged apps should not write there. non-packaged apps should not write there either but they do anyway sometimes.

the-phinet:
> Haiku dev here: ~/config/settings/UZDoom/Screenshots/ is the correct directory on Haiku for packaged apps to store data to. (capitalized Screenshots is ok) non-packaged is a space for the user to install their own non-packaged apps, fonts and other data. Haiku packaged apps should not write there. non-packaged apps should not write there either but they do anyway sometimes.

Okay, that makes sense. I will have screenshots placed in XDG_CONFIG_HOME

Unrelated to screenshots:

Currently save data is written to XDG_DATA_HOME on both Linux and Haiku. This makes sense on Linux, but you're saying this dir should be considered read-only. Should save data also be placed in XDG_CONFIG_HOME instead?

Peppersawce:
> Should save data also be placed in XDG_CONFIG_HOME instead?

As far as I know, yes.

mc776:
Whatever, I just realized cropping my scrot output in Kolourpaint somehow still results in a smaller file than UZDoom's screenshots anyway, do whatever y'all want I'm unbinding the screenshot key.

Peppersawce:
@the-phinet Unrelated to this ticket but still related to platform parity, hit me up whenever you guys merge ZWidget and I'll test compilation on Haiku and make fixes if needed.

If the plan is to merge ZWidget after 4.14.3 keep in mind that the source shouldn't compile on Haiku as-is, I'll be setting up the packaging part for the Haiku repos either way but in case you want 4.14.3's source to compile there I can provide patches.

RicardoLuis0:
> @the-phinet Unrelated to this ticket but still related to platform parity, hit me up whenever you guys merge ZWidget and I'll test compilation on Haiku and make fixes if needed.
> 
> If the plan is to merge ZWidget after 4.14.3 keep in mind that the source shouldn't compile on Haiku as-is, I'll be setting up the packaging part for the Haiku repos either way but in case you want 4.14.3's source to compile there I can provide patches.

4.14.3 can't use updated zwidget unfortunately, it's based on a way older codebase, so it'll most likely need patching for haiku yeah

Peppersawce:
Actually, nevermind, I just checked, 4.14.3-rc2 compiles cleanly on Haiku with no patches needed, ZWidget there is from before someone broke my older patch for it. :)
Unfortunately that version of ZWidget freezes on startup for a while (at least visually, it still accepts inputs) on the system but it is what it is.

ZWidget in trunk on the other hand is the one that needs patching (to make it use SDL2 instead of X11), but otherwise works fine on Haiku.

the-phinet:
Is the frozen thing consistent? It might be a bug that we've been trying to squash for a while. It's really inconsistent on Linux, so verifying fixes is basically impossible

Peppersawce:
On Haiku it is consistent and present on every ZWidget version I've seen so far.
On trunk it lasts for a couple of seconds at most. (it's why I consider it "fine")
In 4.14.3-rc2 it's usually way more. And maybe interacting with the window helps in some way?

I've tried building all projects using ZWidget (Raze, GZDoom, SurrealEngine, ZWidget itself) I think back in July and every single one of them had the same issue btw.

Edit: ZWidget on Haiku uses SDL while on Linux it builds with X11 by default. Building ZWidget with SDL on Linux might help reproduce it there.

TehRealSalt (dismissed):
I can't confirm the Haiku stuff but the Linux side works well :+1:

src/common/platform/posix/unix/i_specialpaths.cpp
```diff
@@ -211,7 +212,13 @@ FString M_GetDocumentsPath()
 
 FString M_GetScreenshotsPath()
 {
+#ifdef __HAIKU__
 	return M_GetDocumentsPath() + "screenshots/";
+#else
+	static FString path = FStringf("%s/Screenshots/" GAMENAME, GetPicturesPath());
+	path = NicePath(path.GetChars());
+	return path;
+#endif
```
Peppersawce:
```suggestion
#ifdef __HAIKU__
	static FString path = FStringf("%s/" GAMENAMELOWERCASE "/screenshots", GetConfigPath());
#else
	static FString path = FStringf("%s/Screenshots/" GAMENAME, GetPicturesPath());
#endif
	path = NicePath(path.GetChars());
	return path;
```
This should set the screenshots directory on Haiku to  `~/config/settings/uzdoom/screenshots` as jscipione indicated.

Both directories are lowercase, for the UZDoom one it's because the .ini file creation already makes it like that, for "screenshots" it can be made uppercase no problem.

A similar thing should be done for the save directory but it's outside the scope of this pr.

Edit: fixed a typo, I was using the wrong string at first. :)

the-phinet: Updated (also savegame/demo dirs)

Peppersawce: Looks good to me :)

TehRealSalt (approved):

PR #567, opened 2025-11-23
Made sure Actor-owned Objects match client and save state

Boondorl:


src/scripting/vmthunks_actors.cpp
```diff
@@ -1918,6 +1918,7 @@ static void SetViewPos(AActor *self, double x, double y, double z, int flags)
 	if (!self->ViewPos)
 	{
 		self->ViewPos = Create<DViewPosition>();
+		self->ViewPos->ObjectFlags |= (self->ObjectFlags & (OF_ClientSide | OF_Transient));
```
TehRealSalt: I didn't see a write barrier for ViewPos when I did a cursory glance over the code, whereas ModelData gets it. Not sure if this is actually a problem or if I just missed where it happens, though

TehRealSalt (commented):
Looks right to me and didn't notice any new issues while playing, but I have one little question

src/scripting/vmthunks_actors.cpp
Boondorl: @RicardoLuis0 Does this need to be fixed?

RicardoLuis0: it wouldn't hurt at least

RicardoLuis0 (approved):

PR #568, opened 2025-11-23
Fix sound sequence thing (arg) sound sequence index range

biwa:
In 10c833f the limit of sound sequences was raised from 63 to 4095. This change was only to polyobjs, but not to the (arguably rather obscure) sound sequence thing that takes the index as arg0 ([doomednum 1411](https://zdoom.org/wiki/Sound_sequence_thing)).

This PR fixes this issue. It also improves the message shown, displaying the actual possible range.

Example map: [sndseq-thing-limit.zip](https://github.com/user-attachments/files/23696148/sndseq-thing-limit.zip)
In UZDoom 4.14.3-rc2 It will show a message, saying that sound sequence 4095 is out of range, and not play the sound sequence. With the changes in this PR it will not show the message, and the sound sequence works correctly.

### DIFF
--- a/src/playsim/p_mobj.cpp
+++ b/src/playsim/p_mobj.cpp
@@ -96,13 +96,13 @@
 #include "p_terrain.h"
 #include "r_sky.h"
 #include "r_utility.h"
+#include "s_sndseq.h"
 #include "sbar.h"
 #include "serialize_obj.h"
 #include "serializer_doom.h"
 #include "shadowinlines.h"
 #include "teaminfo.h"
 #include "thingdef.h"
-#include "s_sndseq.h"
 
 // MACROS ------------------------------------------------------------------
 
--- a/src/playsim/p_mobj.cpp
+++ b/src/playsim/p_mobj.cpp
@@ -102,6 +102,7 @@
 #include "shadowinlines.h"
 #include "teaminfo.h"
 #include "thingdef.h"
+#include "s_sndseq.h"
 
 // MACROS ------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


RicardoLuis0 (approved):

src/playsim/p_mobj.cpp
```diff
@@ -6704,9 +6705,9 @@ AActor *FLevelLocals::SpawnMapThing (FMapThing *mthing, int position)
 	{
 		int type = mthing->args[0];
 		if (type == 255) type = -1;
```
TehRealSalt, line 6707: It looks like type 255 is totally inaccessible? I'm not sure what the best solution would be there that wouldn't break existing maps.

Worst case is that this expanded version might need to be a separate thing, and the error message direct users to the new one if it's out of range. Maybe there's something smarter that could be done though.

TehRealSalt (changes requested):

src/playsim/p_mobj.cpp
biwa: That's correct. As per the wiki 255 means "default", so at least it's documented. So I guess the question is if potentially confusing modders that want to use sound sequence 255 with that thing is worse than not having access to additional 4000+ sound sequences.

TehRealSalt: ngl I'm on the side of "it's worse" because the modern way to set these is via name in UDMF maps' sector field, so I consider the thing as being left in for backwards compatibility with older map formats. I don't really gel with encouraging using an older feature by updating it but also leaving its unintuitive behavior in

But that does give me a thought, why not allow arg0str for this object instead of args[0] if it's set? That'd be the best of both worlds IMO, no limits but would remain fully intuitive

biwa: Using arg0str sounds like a pretty good idea.

the-phinet:
﻿
﻿

PR #578, opened 2025-11-25
Add 'NoGiveAll' inventory flag

kevansevans:
Prevents the 'Give All' command from giving inventory items with this flag enabled, but will not prevent 'give everything' or 'give [item name]' from working

### DIFF
--- a/wadsrc/static/zscript/actors/player/player_cheat.zs
+++ b/wadsrc/static/zscript/actors/player/player_cheat.zs
@@ -90,7 +90,7 @@ extend class PlayerPawn
 			if (type != NULL)
 			{
 				let def = getDefaultByType(type);
-				if (giveall == ALL_YES && def.bNoCheat) {}
+				if (giveall == ALL_YES && def.bNoGiveAll) {}
 				else GiveInventory(type, 1, true);
 			}
 
@@ -109,7 +109,7 @@ extend class PlayerPawn
 				if (ammotype && getDefaultByType(ammotype).GetParentAmmo() == ammotype)
 				{
 					let def = getDefaultByType(ammotype);
-					if (def.bNoCheat && giveall != ALL_YESYES) continue;
+					if (def.bNoGiveAll && giveall != ALL_YESYES) continue;
 
 					let ammoitem = FindInventory(ammotype);
 					if (ammoitem == NULL)
@@ -171,7 +171,7 @@ extend class PlayerPawn
 					{
 						let def = getDefaultByType(keytype);
 						
-						if (def.bNoCheat && giveall != ALL_YESYES) continue;
+						if (def.bNoGiveAll && giveall != ALL_YESYES) continue;
 
 						if (def.special1 != 0)
 						{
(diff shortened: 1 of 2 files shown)


XaserAcheron:
I know rename requests are the most annoying thing ever, but I think it'd better to explicitly call this "NoAllCheat" rather than "NoCheat", otherwise players are likely to expect "give [itemname]" to be blocked and think the flag is broken.

Other than that: nice! I'd use the hell out of this myself.

kevansevans:
Fair point, I can update that here real quick. I think ``NoGiveAll`` will be a more transparent name

kcat:
Is there a reason to prevent a cheat from giving items like this? If a user is using the `give all` console command, they expect to get all items, and this feels like a way to tell a user they can't cheat for items when they clearly want to (and still can by using the specific item name for the command, but that's not easy to find making it unnecessarily annoying).

MajorCooke:
> Is there a reason to prevent a cheat from giving items like this? If a user is using the `give all` console command, they expect to get all items, and this feels like a way to tell a user they can't cheat for items when they clearly want to (and still can by using the specific item name for the command, but that's not easy to find making it unnecessarily annoying).

There is +CHEATNOTWEAPON for weapons, but there isn't anything to stop regular inventory items from being given, meaning this flag completes the set. This is just a means to help prevent items that could accidentally screw around with mission critical assets for example. `give everything` bypasses both these flags entirely, so they can use it at their own risk.

One good usage example, I have an inventory item that triggers a jumpscare which I definitely don't want `give all` to give because it's guaranteed to kill the player after it's done.

mc776:
DoEffect() is also a great way to do stuff like bleeding or poison that can work on (basically) any player or monster. (Yes a thinker would be more appropriate but it's a bit clunky to set up especially for someone coming in from Decorate/Decohack)

Boondorl:
> Is there a reason to prevent a cheat from giving items like this? If a user is using the `give all` console command, they expect to get all items, and this feels like a way to tell a user they can't cheat for items when they clearly want to (and still can by using the specific item name for the command, but that's not easy to find making it unnecessarily annoying).

Some things simply aren't meant to be given via cheats but the engine has no way of deciphering this without replacing the player themselves e.g. mechanic tokens. Behaviors are a nice replacement for simple cases but they cannot capture all of them since they don't have the same level of virtualizing as an item. While `give all` does its best to stop common cases, having an explicit flag simply makes this a million times easier. We can outright remove the hardcoded checks (e.g. PowerUps) in favor of just giving them this flag, which also introduces the opportunity to _give_ previously gated things with `give all`.

the-phinet:
﻿
﻿

kevansevans:
Sorry for the late response, I have agreed to the terms. My code was written myself and I agree to it being published under GPL.

RicardoLuis0 (changes requested):
might be a nitpick but, even if zscript is case-insensitive, getDefaultByType should be GetDefaultByType (other than that it's good to merge)

kevansevans:
Sorry for getting to this so late, updated the capitalization

wadsrc/static/zscript/actors/player/player_cheat.zs
RicardoLuis0: still missing here

RicardoLuis0: here

RicardoLuis0: here

RicardoLuis0: and here 😅

kevansevans:
wtf I ctrl+f'd that and only got two results, what the hell?

RicardoLuis0 (approved):

PR #606, opened 2025-11-27
Added owner sanity check to Weapon::CheckAmmo

AKMDM:
This specifically fixes a scenario where if a player class has any `StartItem` that's a weapon forbidden to them, the game throws an error and aborts. Although the player doesn't receive the item as intended, `Player::GiveDefaultInventory` then calls `Weapon::CheckAmmo` on the item, which erroneously assumes that `Owner` is a valid pointer. Adding a sanity check fixes this.

This was first reported by a couple of people in Zandronum, where it crashes the game instead. This is the example `DECORATE` code that they sent me:
```
Actor DoomPlayer2 : DoomPlayer
{
Player.StartItem "Pistol1"
Player.DisplayName "DebugTest"
} 

Actor Pistol1 : Pistol{Inventory.ForbiddenTo "DoomPlayer2"}
```

### DIFF
--- a/wadsrc/static/zscript/actors/inventory/weapons.zs
+++ b/wadsrc/static/zscript/actors/inventory/weapons.zs
@@ -959,6 +959,10 @@ class Weapon : StateProvider
 		int lAmmoUse1;
         int lAmmoUse2 = AmmoUse2;
 
+		if (Owner == null)
+		{
+			return false;
+		}
 		if (sv_infiniteammo || (Owner.FindInventory ('PowerInfiniteAmmo', true) != null))
 		{
 			return true;


mc776:
EDIT: ok if the previous behaviour was just that the whole thing would crash my reply that was here can be safely disregarded

RicardoLuis0:
since this is simply "allowing" something that crashed before, i don't think it's worth adding - or rather, an explicit error should be added for startitems given to the wrong class (so that it isn't a null crash, but doesn't allow broken shit to work silently)

BinarryCode:
I feel like this might be better off fixed on Zandronum’s end. I believe in the server I reported it in, he mentioned  one of the issues doesn’t happen in UZDoom, maybe the base version but idk.

the-phinet:
﻿
﻿

PR #633, opened 2025-11-30
Update a_scroll.cpp

### LINKED ISSUE #631: The floor/ceiling texture scrolling direction is transformed erroneously when texture is rotated(angle not 0)
### UZDoom Version

UZDoom 4.14.3 (and I believe all earlier versions of G/Zdoom as well.)

### Game

All

### Operating System

Windows 10

### OS Version

_No response_

### Hardware

_No response_

### Duplicate Confirmation

### Bug Description

TL;DR Flip the sign(remove the minus) on src/playsim/mapthinkers/a_scroll.cpp:132
*double ca = -an.Cos();*
and/or make a compat flag to preserve the old behaviour in case there are maps that rely on it.

Line action **223: Scroll floor**(UDMF) seems to yield unexpected results when the floor texture rotation is non-zero(same for **224: Scroll ceiling**). I believe this is in error because there is a mismatch between the scrolling direction of textures and the direction things are carried, when textures are rotated.
I made a test map with a ring of scrolling textures, trying to see a pattern in order to figure out exactly how far off the transformation was from what I had intended.
*1 degree* rotation seems to scroll almost, but not quite 180 degrees backwards(179?)
*90 and 270 degrees* rotation scrolls in the expected directions.
*180* degrees rotation scrolls backwards.
And *in-between* angles go more or less toward being *perpendicular* to the expected scroll directions.
I made a video... it's a bit too lengthy, sorry about that...
https://youtu.be/iqGumI6CgWY

also, here's a small demo map:
[ale_scrolltest.zip](https://github.com/user-attachments/files/23836617/ale_scrolltest.zip)

I looked through the code base, and believe I've found the relevant function:
RotationComp (src/playsim/mapthinkers/a_scroll.cpp:122-137)
Going by the function's comment, the intent seems to be to counter-rotate, but the transformation applied is not actually a rotation.

I then made this demonstration in desmos, and i believe it shows both the problem and a solution:
https://www.desmos.com/calculator/9dja13abyz

### Replication

1. Make a scrolling floor(or ceiling) using the line actions 223(or 224).
2. Change the angle of the floor(or ceiling) texture(s) to a non-zero value.

### Configuration

```ini

```

### Log

_No response_

aletheos:
Remove the sign(-) on -an.Cos(), fixing the intended counter rotation.[Fixes #631]

I did *not* make a compat flag(I don't know how), but maybe there should be one just in case there are maps that intentionally use the old behaviour?

### DIFF
--- a/src/playsim/mapthinkers/a_scroll.cpp
+++ b/src/playsim/mapthinkers/a_scroll.cpp
@@ -129,7 +129,7 @@ static void RotationComp(const sector_t *sec, int which, double dx, double dy, d
 	}
 	else
 	{
-		double ca = -an.Cos();
+		double ca = an.Cos();
 		double sa = -an.Sin();
 		tdx = dx*ca - dy*sa;
 		tdy = dy*ca + dx*sa;


the-phinet:
Any updates on getting the compat flags in?

the-phinet:
﻿
﻿

aletheos:
> Any updates on getting the compat flags in?

sorry, life got in the way.
I'm reading source whenever I have the time, but I'm operating on strict Valve/3D-Realms time here(it'll be done when it's done).

PR #657, opened 2025-12-01
No Dynamic Lighting Renderflag

MajorCooke:
Adds +NODYNAMICLIGHTING for actors and SPF_NODYNAMICLIGHTING for particles and VisualThinkers.

This disables all dynamic lighting effects from affecting sprites/models/overlays.


mc776:
TIL this did *not* already exist!

MajorCooke:
Ya, blew my mind discovering that.

And to be clear, this does **not** prevent actors from having/emitting dynamic lights, this just prevents the sprite from being  affected by them.

Now to add this to particles/vthinkers, but that'll be for another PR.

MajorCooke:
You know what, I'll just go ahead and do it now in this one. Why not.

mc776:
> And to be clear, this does not prevent actors from having/emitting dynamic lights, this just prevents the sprite from being affected by them.

I'm tempted to suggest the flag have a name that clarifies this but I don't know what.

(Or if it's even a real concern given a few seconds' thought would make it clear there's no reason why a flag preventing the *emitting* of a light would exist)

MajorCooke:
The term "lighting" I think is enough.

MajorCooke:
All done. I optimized my prior code by simply disabling fetching of dynamic lights for the actors/particles/vthinkers with this flag. Much more optimal.

RicardoLuis0 (approved):

TehRealSalt (approved):
I'd personally prefer a different name (maybe `bDontDynamicLight` or `bExcludeDynamicLighting`). The word 'No' may be confusable with turning off all of the object's lights, at least to me, rather than removing them from lighting lists.

But since I just have the one probably bikeshed complaint, I'm approving anyways. If you disagree, I'm OK with it being merged in as is

MajorCooke:
I thought so at first but I didn't really like those in the end so I'll just keep it as it is. Thanks though!

mc776:
> `bDontDynamicLight`

I almost suggested this but was too pedantic about grammar to do it. It really is the best option.

PR #658, opened 2025-12-02
minor optimizations/improvements to IQM

RicardoLuis0:
* only check unchanging variables once per loop, not once per iteration
* allow bone offsets to work without animations playing
* misc fixes


TehRealSalt (approved):

MajorCooke:
> allow bone offsets to work without animations playing

I was wondering why whenever I summoned a model, it was a mess for the initial frame, only to correct itself the moment the first animation started playing because I never changed anything. Looked like a splatter on the ground. XD

PR #675, opened 2025-12-04
Allow loading LZDoom 4 saves

Peppersawce:
I don't know if it is a smart idea and whether this is the best way to do it, but this allows LZDoom 4 saves to be loaded in UZDoom, currently tested with a Heretic save from LZDoom's 4.11.4 release where everything worked fine.

Should also work with saves from current LZDoom trunk (edit: I mean `l4.14dev`, at version 4.14.3), I tested this by editing `info.json`. 4.15 and above don't show up and that's currently intended by looking at the rest of the code.

GZDoom saves loading shouldn't be affected.

### DIFF
--- a/src/g_game.cpp
+++ b/src/g_game.cpp
@@ -2070,7 +2070,7 @@ void G_DoLoadGame ()
 	#if LOAD_GZDOOM_4142_SAVES
 	FString software = arc.GetString("Software");
 	bool gzdoom_compat_ok = false;
-	if(engine.Compare("GZDOOM") == 0)
+	if(engine.Compare("GZDOOM") == 0 || engine.Compare("LZDOOM") == 0)
 	{
 		gzdoom_compat_ok = CheckGZDoomSaveCompat(engine, software);
 	}
--- a/src/menu/loadsavemenu.cpp
+++ b/src/menu/loadsavemenu.cpp
@@ -62,7 +62,7 @@ EXTERN_CVAR(Int, save_sort_order)
 
 bool CheckGZDoomSaveCompat(FString &engine, FString &software)
 {
-	if(software.IndexOf("GZDoom g") == 0)
+	if(software.IndexOf("GZDoom g") == 0 || software.IndexOf("LZDoom l") == 0)
 	{
 		if(software.Len() < 11) return false; // GZDoom g????????
 
@@ -101,7 +101,7 @@ bool CheckGZDoomSaveCompat(FString &engine, FString &software)
 					//GZDoom g4.14.##, don't allow
 					return false;
 				}
-				else if(v2 > 2)
+				else if(v2 > 2 && software.IndexOf("GZDoom g") == 0)
 				{
 					//GZDoom g4.14.3+, don't allow
 					return false;
@@ -109,7 +109,7 @@ bool CheckGZDoomSaveCompat(FString &engine, FString &software)
 				/*
 				else
 				{
-					//GZDoom g4.14.0 / GZDoom g4.14.1 / GZDoom g4.14.2, allow
+					//GZDoom g4.14.0 / GZDoom g4.14.1 / GZDoom g4.14.2, LZDoom l4.14.#, allow
 				}
 				*/
 			}
@@ -123,7 +123,7 @@ bool CheckGZDoomSaveCompat(FString &engine, FString &software)
 		/*
 		else
 		{
-		//GZDoom g4.0.0 - GZDoom g4.13.2, allow
+		//GZDoom g4.0.0 - GZDoom g4.13.2, LZDoom l4.#, allow
 		}
 		*/
 	}
@@ -182,7 +182,7 @@ void FSavegameManager::ReadSaveStrings()
 						#if LOAD_GZDOOM_4142_SAVES
 						FString software = arc.GetString("Software");
 
-						if(engine.Compare("GZDOOM") == 0)
+						if(engine.Compare("GZDOOM") == 0 || engine.Compare("LZDOOM") == 0)
 						{
 							if(!CheckGZDoomSaveCompat(engine, software))
 							{


Peppersawce:
`goffset` could also be used here to differentiate between a GZDoom after 4.14.2 (not allowed) and a later LZDoom (which would be based on UZDoom) (allowed?)
https://github.com/UZDoom/UZDoom/blob/fbf5d04972dfb34a7f89b22cd8a5902d0fbd38ef/src/menu/loadsavemenu.cpp#L110-L114

drfrag666:
Wenas. There's no trunk in LZDoom xD. The default branch now is l4.14dev, master is l4.15pre. I updated the version number for the upcoming release becouse the version was wrong, it was updated for g4.15 and i forgot to change it.
Edit: where does the 'GZDoom g' come from? The latest LZDoom version is l4.11.4, the tag will be l4.14.3.

Peppersawce:
In LZDoom 4.11.4's saves (`info.json`) the version string is `"Software": "LZDoom 4.11.4",`

Is it any different in `l4.14dev`? Is is `l4.14.3` or something like that?

drfrag666:
It's l4.14.3 and it was l4.11.4, i thought in GZDoom that string didn't contain a 'g'. The 'g' is in the tag but the VERSIONSTR doesn't contain the 'g'. The LZDoom tags change the 'g' for a 'l'.

drfrag666:
But have you pulled the tags? It should be the same for both engines.

Peppersawce:
Ah, I see what's happening, it's a patch I'm applying locally that's messing it up.

drfrag666:
I've just checked:
`    "Software": "LZDoom l4.11.4",`

Peppersawce:
Ok, now it's fixed frfr

drfrag666:
Should i add compatibility for UZDoom saves? xD
Edit: i think i'll do it.

Peppersawce:
Totally :grin:

RicardoLuis0:
don't think this is the best way to do things, that save loading code is specifically a hack _because_ old versions of gzdoom can't be changed, since lzdoom is under active development the best option is to add a new property to the info json, ex. `"AllowLoadIn":["UZDoom"]` (that way uzdoom can explictly check if loading is supported, without needing to parse the description string)

Peppersawce:
I was thinking, does UZDoom really have to go through all these checks and restrictions just to try and load saves from other ZDoom engines?
Can't it just be:

1. Check save format
2. Check minimum version
3. Offer to load the save if format + version are fine

It's not like saves from other versions magically appear in the save folder, so I don't really get why loading saves is limited.
If a player puts a save from another ZDoom in there it's intentional, so the game might as well try and load it if possible imo, maybe with a disclaimer like "This save comes from another ZDoom engine, it might fail to load in UZDoom" or something.

RicardoLuis0:
yes, because a zdoom derived engine can just set its save version to anything, and we must only load saves that we know will be compatible, or else people will just come on here and complain that their saves won't load 🤷‍♀️

Peppersawce:
> because a zdoom derived engine can just set its save version to anything, and we must only load saves that we know will be compatible

If that's the reasoning, what's the point in storing a minimum-compatible save version then? Just remove it if you're just going to not load those saves either way.

This kind of reasoning also makes your `"AllowLoadIn":["UZDoom"]` suggestion from before moot imo. If you can't trust the other engines' saves to work based on the "minimum version" requirement, why would you trust them explicitly saying "this can be loaded in UZDoom"?

> people will just come on here and complain that their saves won't load

First off, they'd have to intentionally try and load these by putting them in UZDoom's save folder, second of, as I said, just put a disclaimer at load-time saying it's not guaranteed to work. If people complain just point them at the disclaimer.

RicardoLuis0:
> If that's the reasoning, what's the point in storing a minimum-compatible save version then? Just remove it if you're just going to not load those saves either way.

it's a minimum version _for gzdoom_ not for any port, because we can't know if a future gzdoom 4.15 or 4.14.3 or anything else will make any incompatible changes as we don't work on it anymore, but we _know_ that 4.14.2 itself or older won't change

> This kind of reasoning also makes your "AllowLoadIn":["UZDoom"] suggestion from before moot imo. If you can't trust the other engines' saves to work based on the "minimum version" requirement, why would you trust them explicitly saying "this can be loaded in UZDoom"?

because if something explicitly says it can be loaded in uzdoom, and doesn't work, that's on them, not us (basically, it'd be a way of a port saying, "we'll keep our save versioning in sync with uzdoom")

Peppersawce:
> it's a minimum version for gzdoom not for any port, because we can't know if a future gzdoom 4.15 or 4.14.3 or anything else will make any incompatible changes as we don't work on it anymore, but we know that 4.14.2 itself or older won't change

I get that, sure, old GZDoom can't break compatibility and I get you'd like to exclude some experimental forks that can break things willy-nilly, but what about LZDoom? It's directly based on the GZDoom/UZDoom code and isn't a project meant to do huge gameplay changes that would break compatibility.

I see little reason to keep their saves out of this, their saves from last release work on UZDoom and it's extremely unlikely (imo) that's going to break, since it's the same codebase.

I feel you're exaggerating this compatibility thing a bit, but then again maybe you're speaking from a place of experience due to your involvement in this community and changes like this could be a sh*tshow waiting to happen.

In other words, I (partially) get your concern, I'll keep this PR open a bit more for discussion's sake.
If the maintainers don't feel it's worth it that's that and I don't feel this is a hill worth dying on.
(not when editing an `info.json` can easily circumvent this entire "issue")

drfrag666:
For what is worth i've added support for loading UZDoom saves. I've reverted some changes since for some reason i don't remember i was preventing loading LZDoom 3 saves and besides that there were other changes in GZDoom and before future saves were deleted, now they are not. I thought about loading future saves too but i think those could crash.
https://github.com/drfrag666/lzdoom/commit/4da4d23df2c861e7b2209cfb7a98d68adbfa456f

RicardoLuis0:
> but what about LZDoom? It's directly based on the GZDoom/UZDoom code and isn't a project meant to do huge gameplay changes that would break compatibility.

the whole point of what i'm saying is that _because_ lzdoom won't break compatibility it doesn't need hackish compat checks in the save loading, a proper way to signify save compat can be added for it instead

RicardoLuis0:
(UZDoom would also add an `"AllowLoadIn":["LZDoom"]` key to _its_ info.json, so that no hacks are needed in _either_ port for compat, all that's needed is a version check)

Peppersawce:
> the whole point of what i'm saying is that because lzdoom won't break compatibility it doesn't need hackish compat checks in the save loading, a proper way to signify save compat can be added for it instead

Okay, that's fair, now I see where you're coming from. No sense in extending a temporary(?) hack for a dead sourceport to one that's still being worked on.

Thanks for the clarification, I'm going to close this PR as this is not the way to implement the feature.

PR #676, opened 2025-12-04
Reworked the scoreboard

Boondorl:
Rewrote the API to be much cleaner and changed the general design. There are now four main virtual functions: InitScoreboard, DrawScoreboard, DrawRemainingTime, and DrawPlayerScores. I feel this is robust enough to modify each major part of the board without having to rewrite lots of things, but more can be added later as needed and I'd rather start cut back and add on than the other way around. Features:
* New layout that's more intuitive and significantly easier to read.
* Cleaner scaling.
* Proper character limit on names (up to 32 characters) and visible player id for better moderation handling.
* Up to 10 players and 12 team scores visible at once.
* Better handling of player amounts > 8 (the player and their team score if applicable are now always rendered even if out of room).
* New API that has reduced indirection and more generalized functionality.
* Minor bug fixes.

Team logos are now drawn behind the team score to avoid clutter and player classes are drawn to the left of the player name column next to each player.

In the future I'll be adding a full scoreboard mode that supports UI input for things like scrolling to see all players, but the main purpose of this PR is to get the scoreboard API rewritten before 5.0 release as it was too messy and difficult to maintain. The intermission scoreboard has not been updated as that itself will also require a rewrite to better use the existing scoreboard infrastructure.

Previous:
Screenshot_Doom_20251204_101227
New:
Screenshot_Doom_20251204_101129

### DIFF
--- a/src/ct_chat.cpp
+++ b/src/ct_chat.cpp
@@ -248,15 +248,16 @@ void CT_Drawer (void)
 
 	if (players[consoleplayer].camera != NULL &&
 		(buttonMap.ButtonDown(Button_ShowScores) ||
-		 players[consoleplayer].camera->health <= 0 ||
-		 SB_ForceActive))
+		 players[consoleplayer].playerstate == PST_DEAD ||
+		 bScoreboardToggled))
 	{
 		bool skipit = false;
-		if (gamestate == GS_CUTSCENE)
+		if (gamestate != GS_LEVEL)
 		{
-			// todo: check for summary screen
+			bScoreboardToggled = false;
+			skipit = true;
 		}
-		if (!skipit) HU_DrawScores (consoleplayer, vp.TicFrac);
+		if (!skipit) HU_DrawScores (vp.TicFrac);
 	}
 	if (chatmodeon)
 	{
--- a/src/hu_scores.cpp
+++ b/src/hu_scores.cpp
@@ -132,7 +132,7 @@ void HU_SortPlayers
 }
 */
 
-bool SB_ForceActive = false;
+bool bScoreboardToggled = false;
 
 
 // PRIVATE DATA DEFINITIONS ------------------------------------------------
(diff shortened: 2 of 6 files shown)


mc776:
How would it look if you had the names (not the numbers or colours) on the rightmost column? That way all the other columns can be made narrower and more compact (and a lot easier to see which item on which row corresponds to which others) but you can still accommodate for names of wildly varying length.

Boondorl:
Ehhhhh, this is a really common layout for modern games to use. I could make the other rows more compact, but I need to make sure there's enough room for localization and really, there's not any reason to have that much room dedicated to the name. Names beyond 32 characters are probably just people trolling anyway. One thing I actually would like to do is remove that ugly ` (ms)` from the Delay title and change it to Latency.

Hypothetically I could make the whole thing more compact (it's still designed to fit in 4:3 comfortably), but that's just minor tweaks that can be done later. I really, _really_ need this API rewrite in because there is no way in hell I am maintaining that monstrosity for the rest of the engine's life. I'd sooner de-scriptify everything than keep that beast around.

mc776:
Keep the (ms), there's no reason not to be told the units being used.

EDIT: Ok yes it is ugly. How about making all the items say "[number] ms" so at least there isn't that disruptive parenthetical?

mc776:
Can we have at least thin lines between the table rows though? I have to deal with a similar format at work and third parties misread the table in question regularly.


EDIT: And since I'm asking for stuff anyway, might as well throw in a less controversial one: light grey or white table labels instead of red?

Boondorl:
The label color is a user option

src/ct_chat.cpp
```diff
@@ -248,15 +248,16 @@ void CT_Drawer (void)
 
 	if (players[consoleplayer].camera != NULL &&
 		(buttonMap.ButtonDown(Button_ShowScores) ||
-		 players[consoleplayer].camera->health <= 0 ||
-		 SB_ForceActive))
+		 players[consoleplayer].playerstate == PST_DEAD ||
+		 bScoreboardToggled))
```
TehRealSalt, line 252: I think this is supposed to be done on `players[consoleplayer].camera` so that it comes up even when you're using spycam on a dead player. If this is intentional though, probably shouldn't have the check for `players[consoleplayer].camera != NULL`.

TehRealSalt (changes requested):
I noticed one minor mistake, but otherwise it looks good to go to me

src/ct_chat.cpp
Boondorl: It was indeed intentional, I think how the scoreboard works should be based solely on the console player's state. That being said, I'm not the biggest fan of this behavior in general and will likely axe it later in part 2 where I examine its behavior a lot more closely. This was a bit of an emergency rewrite and since I was already in there rewriting things, I figured I might as well clean up the design a bit as well.

TehRealSalt: I maintain that it doesn't really need to check for `camera`'s existence but if it's planned for another pass then I won't block over that, especially since I realized afterwards that camera's practically already guaranteed to always exist

TehRealSalt (approved):

Issue #681, opened 2025-12-05
Add suport for prefixed z/mapinfo keywords in UMAPINFO


elf-alchemist:
### Game

All

### Problem Description

Outside of the barely reported parsing issues, one of the big problems with using UMAPINFO for cross-port projects is the lack of support for certain desirable keywords from Z/MapInfo definitions -- such as the ability to disable jumping or crouching.

It is problematic to have to double-define map entries on two separate lumps, when the same information could be provided on just one of them.

While it isn't officially part of the UMAPINFO spec, due to certain delays in the formalization of it, the de-facto way to add port-exclusive keywords is to prepend them with the name of the port, and an underscore, e.g `zdoom_*` or `kex_*`, the latter example is already a strong precedent present in the official Legacy Of Rust wad (id1.wad) as a reference, and is what will be included in the 

### Feature Description

This issue is specifically asking for `zdoom_*` or `uzdoom_*` prefixed versions of ZMAP-exclusive keywords to facilitate work de-duplication, making modders lives easier.

EdHerdman:
Sounds good, but it looks like the text got cut short at the end of the Problem Description.

PR #683, opened 2025-12-05
Add support for new UMAPINFO keys `freeaim`, `jumping`, `crouching`

electricbrass:
A frequent issue with projects using UMAPINFO, is that if they are designed with the intention that crouching, jumping, and freelook are unavailable or disabled, they end up having to redefine all their maps in ZMAPINFO to disable these features in UZDoom or other similar workarounds for other ports. To remedy this, a proposal to add `freeaim`, `jumping`, and `crouching` keys to UMAPINFO has been made. The exact details of the specification haven't been worked out just yet, hence this being just a draft for now, but the current wording can be viewed at https://github.com/kraflab/umapinfo/pull/6

This addresses a large part of the need for #681, but should not be seen as a complete solution for that issue, as this only addresses some of options that are also common in ports other than UZDoom.

### DIFF
--- a/src/gamedata/umapinfo.cpp
+++ b/src/gamedata/umapinfo.cpp
@@ -35,6 +35,14 @@ struct BossAction
 	int tag;
 };
 
+enum struct PlayerActionOption
+{
+	DISALLOW,
+	ALLOW,
+	REQUIRE,
+	UNSET,
+};
+
 struct UMapEntry
 {
 	FString MapName;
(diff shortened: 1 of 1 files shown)


SirBofu:
This is brilliant. It might also be a good idea to add the [ZDoom compatibility options](https://zdoom.org/wiki/Compatibility_options), though there are a lot of those.

bangstk:
> This is brilliant. It might also be a good idea to add the [ZDoom compatibility options](https://zdoom.org/wiki/Compatibility_options), though there are a lot of those.

For this push I think we want to avoid adding a bunch of port specific keys. There is discussion on designing a more suitable successor format which can better handle that sort of thing.

For now, jumping and freelook etc. are ubiquitous enough that these keys are useful to most ports.

For everything else, my current quick proposal would be a ZMAPINFO default block key that tells UZDoom that it's okay to get the map definitions from UMAPINFO while still applying the default block for compat stuff.

electricbrass:
It looks like the discussion has reached an agreement on this revision. Exact wording of the spec hasn't been determined yet, but shouldn't have any further effect on this PR, so I'm marking this as ready.

RicardoLuis0:
sorry to be annoying but can you clean up the commit history a bit? 😅
other than that looks good 👍

electricbrass:
Absolutely, will do

RicardoLuis0 (approved):

PR #689, opened 2025-12-06
Damage-modifying Function Expansion

MajorCooke:
Expanded `<Modify/Absorb>Damage` and `<Do/Take>SpecialDamage` to include flags/angle from DamageMobj where absent.

For the unfamiliar, this does not introduce incompatibilities. `GetModifiedDamage` was expanded once before with no negative effect on mods that don't utilize optional parameters in virtual overrides.

### DIFF
--- a/src/playsim/p_interaction.cpp
+++ b/src/playsim/p_interaction.cpp
@@ -1215,7 +1215,7 @@ static int DamageMobj (AActor *target, AActor *inflictor, AActor *source, int da
 					}
 				}
 
-				damage = inflictor->CallDoSpecialDamage(target, damage, mod);
+				damage = inflictor->CallDoSpecialDamage(target, damage, mod, flags, angle);
 				if (damage < 0)
 				{
 					return -1;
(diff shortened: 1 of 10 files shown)


src/playsim/p_interaction.cpp
```diff
@@ -1848,7 +1848,7 @@ void P_PoisonDamage (player_t *player, AActor *source, int damage, bool playPain
 	// Take half damage in trainer mode
 	damage = int(damage * G_SkillProperty(SKILLP_DamageFactor) * sv_damagefactorplayer);
 	// Handle passive damage modifiers (e.g. PowerProtection)
-	damage = target->GetModifiedDamage(player->poisontype, damage, true, nullptr, source);
+	damage = target->GetModifiedDamage(player->poisontype, damage, true, nullptr, source, 0, DAngle::fromDeg(0));
```
Boondorl: `nullAngle` can be passed as a default here since it's a pre-generated version of `DAngle::fromDeg(0)`

Boondorl (dismissed):

src/playsim/p_interaction.cpp
MajorCooke: Done.

Boondorl (approved):

PR #704, opened 2025-12-08
PP Expansion - Temporal

TheIndiv:
This patch implements a temporal postprocessing system for UZDoom that evolved through several iterations, ultimately settling on a simplified LastInputTexture approach. 

The final version removes the complex per-shader persistent texture system in favor of a single shared LastInputTexture buffer.

This provides all postprocess shaders access to the previous frame's "scene" output (before UI rendering), enabling temporal effects like auto-exposure, motion blur, and temporal anti-aliasing.

This does mean "screen" target shaders (which run after UI rendering) don't receive temporally coherent data for UI-based effects.

The simplified approach introduces UpdateLastInputTexture in PPCustomShaders that copies the scene output after all "scene" shaders run, making previous frame data automatically available to subsequent passes. This reduces complexity while covering most temporal effect use cases. A basic copy.fp shader handles the texture capture. The PPPersistentBuffer class remains but is now used centrally rather than per-shader-instance as in my first proof-of-concept commit.

-------------------------------------------

In a future PR, I'd like to look into expanding the PP pipeline further. I would like to be able to inject shaders on just the first-person weapon, and just the HUD. I would also like to get the "timer" uniform to work with postprocessing shaders.

### DIFF
--- a/src/common/rendering/hwrenderer/postprocessing/hw_postprocess.cpp
+++ b/src/common/rendering/hwrenderer/postprocessing/hw_postprocess.cpp
@@ -906,6 +906,15 @@ void PPCustomShaders::Run(PPRenderState *renderstate, FString target)
 
 	CreateShaders();
 
+	int width = renderstate->Viewport.width;
+	int height = renderstate->Viewport.height;
+
+	if (width > 0 && height > 0 && width <= 16384 && height <= 16384)
+	{
+		mLastWidth = width;
+		mLastHeight = height;
+	}
+
 	for (auto &shader : mShaders)
 	{
 		if (shader->Desc->Target == target && shader->Desc->Enabled)
@@ -913,6 +922,26 @@ void PPCustomShaders::Run(PPRenderState *renderstate, FString target)
 			shader->Run(renderstate);
 		}
 	}
+}
+
+void PPCustomShaders::UpdateLastInputTexture(PPRenderState *renderstate)
+{
+	int width = renderstate->Viewport.width;
+	int height = renderstate->Viewport.height;
+
+	if (width <= 0 || height <= 0 || width > 16384 || height > 16384)
+		return;
+
+	if (!mLastInputTexture ||
+		mLastInputTexture->GetRead()->Width != width ||
+		mLastInputTexture->GetRead()->Height != height)
+	{
+		mLastInputTexture = std::make_unique<PPPersistentBuffer>(width, height, PixelFormat::Rgba16f);
+	}
+
+	renderstate->CopyToTexture(mLastInputTexture->GetWrite());
+
+	mLastInputTexture->Swap();
 }
 
 void PPCustomShaders::CreateShaders()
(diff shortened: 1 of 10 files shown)


src/common/rendering/hwrenderer/postprocessing/hw_postprocess.cpp
```diff
@@ -978,6 +1015,8 @@ PPCustomShaderInstance::PPCustomShaderInstance(PostProcessShader *desc) : Desc(d
 		pipelineInOut += "out vec4 FragColor;\n";
 	}
 
+	uniformTextures += "layout(binding=11) uniform sampler2D LastInputTexture;\n";
```
RicardoLuis0: why are you using a fixed binding for the uniform? this will completely break vulkan/opengl when the custom textures go beyond that, just use the `binding` variable for it and store it in some field of the PPCustomShaderInstance

src/common/rendering/hwrenderer/postprocessing/hw_postprocess.cpp
```diff
@@ -915,6 +924,34 @@ void PPCustomShaders::Run(PPRenderState *renderstate, FString target)
...
+	renderstate->Shader = &CopyShader;
+	renderstate->Viewport = PPViewport{ 0, 0, width, height };
+	renderstate->SetInputSpecialType(0, PPTextureType::CurrentPipelineTexture, PPFilterMode::Nearest, PPWrapMode::Clamp);
+	renderstate->SetOutputTexture(mLastInputTexture->GetWrite());
+	renderstate->SetNoBlend();
+	renderstate->Draw();
+
+	mLastInputTexture->Swap();
+}
+
```
RicardoLuis0, line 946: also this can just be a texture copy from InputTexture to LastInputTexture, ex. glCopyTexSubImage2D or the vulkan equivalent (i think it's vkCmdCopyImage?), no need to do a full render when the shader is just a no-op

RicardoLuis0:
other than that, looks good

dileepvr:
So ... I can't access this outside of PP shader?
I see myself painting an incrementing color value on a corner pixel every frame and reading it as a frame counter in the next.

src/common/rendering/hwrenderer/postprocessing/hw_postprocess.cpp
TheIndiv: Done in latest commit!

src/common/rendering/hwrenderer/postprocessing/hw_postprocess.cpp
TheIndiv: Updated in latest commit! Please take a look! This was a little more complicated, but my testing shows it's working in both Vulkan and OpenGL.

RicardoLuis0: it's mostly good, but blitImage isn't required, since this isn't doing any scaling/filtering/etc, blitImage would just be slower than copyImage for no benefit

TheIndiv: I'll get on that after class! Does the GL implementation look good?

RicardoLuis0: it does yeah

RicardoLuis0: ah wait sorry, just noticed you used glCopyImageSubData instead of glCopyTexSubImage2D, that's an issue

RicardoLuis0: glCopyTexSubImage2D works in any opengl version, but glCopyImageSubData is only core in opengl 4.3, while uzdoom works in opengl 3.3+ (so it would lead to issues for folks using older GPUs)

TheIndiv: I'm home from class and I've made a commit! Do take a look!

RicardoLuis0: looks good 🎉

RicardoLuis0 (approved):
approving from a code review standpoint but still needs testing (i'll try and get to it in a few days but if anyone else tests both opengl and vulkan with this and it's fine then it's good to merge)

RicardoLuis0:
oh also the commit history needs some cleanup

TheIndiv:
Unsure of how to clean the commit history. Or what needs to be cleaned.

RicardoLuis0:
just squashing it down and splitting out relevant parts into significant commits, ex.: (not saying you should do this exactly, but an example)

* add CopyCurrentToImage to VkPostprocess
* save last InputTexture into LastInputTexture
* allow Postprocess shaders to access LastInputTexture

RicardoLuis0:
if you don't want to split it into separate commits just change the commit name to `PP Expansion - Temporal` and it's good to merge 👍

PR #720, opened 2025-12-10
Add automatic time uniforms to PP shaders

TheIndiv:
Basically what the title says. Quick and simple implementation.

### DIFF
--- a/src/common/rendering/hwrenderer/postprocessing/hw_postprocess.cpp
+++ b/src/common/rendering/hwrenderer/postprocessing/hw_postprocess.cpp
@@ -1013,6 +1013,7 @@ PPCustomShaderInstance::PPCustomShaderInstance(PostProcessShader *desc, std::uni
 	FString prolog;
 	prolog += uniformTextures;
 	prolog += pipelineInOut;
+	// Note: Automatic uniforms (InputTimeDelta, InputTime, InputTimeGame) are added by backends
 
 	Shader = PPShader(Desc->ShaderLumpName, prolog, Fields);
 }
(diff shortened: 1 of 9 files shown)


src/common/rendering/hwrenderer/postprocessing/hw_postprocess.cpp
```diff
@@ -1099,6 +1105,19 @@ void PPCustomShaderInstance::SetUniforms(PPRenderState *renderstate)
...
+		uint8_t *dst = &uniforms[timeDeltaOffset->second];
+		memcpy(dst, &renderstate->TimeDelta, sizeof(float));
+	}
+	auto timeOffset = FieldOffset.find("InputTime");
+	if (timeOffset != FieldOffset.end())
+	{
+		uint8_t *dst = &uniforms[timeOffset->second];
+		memcpy(dst, &renderstate->Time, sizeof(float));
+	}
+
```
RicardoLuis0, line 1167: just a nitpick, but the uniform struct is for custom uniforms, should this be added as a uniform field instead? otherwise mods near the size limit of uniform structs may stop working since it was increased in size

TheIndiv: Made a push! Do take a look

RicardoLuis0: the issue isn't UniformStructSize, it's the raw size of the uniform buffer used for the custom uniforms, that buffer is best left only for said custom uniforms, and the input time as just a regular `uniform float InputTime`

RicardoLuis0: (like, this isn't as much of a problem for opengl, where the minimum guaranteed size is 16kb, but for vulkan, the post-process uniform struct uses push constants, and those have a minimum guaranteed size of 128 **bytes**)

TheIndiv: Made another push, redoing everything from basically scratch. Do take a gander

RicardoLuis0: looks good yeah 👍
(i'll hold off on approving/merging this until the other post-process PR is in though, since it'll most likely have a merge conflict)

RicardoLuis0:
re-stating this here:
> looks good yeah 👍
> (i'll hold off on approving/merging this until the other post-process PR is in though, since it'll most likely have a merge conflict)

RicardoLuis0 (approved):

PR #734, opened 2025-12-12
Make menu dimming more tolerable

the-phinet:
As suggested numerous times by several different people. Improves readability of menus

Before / After

image

### DIFF
--- a/src/menu/doommenu.cpp
+++ b/src/menu/doommenu.cpp
@@ -330,7 +330,7 @@ void OnMenuOpen(bool makeSound)
 //
 //==========================================================================
 
-CUSTOM_CVAR(Float, dimamount, -1.f, CVAR_ARCHIVE)
+CUSTOM_CVAR(Float, dimamount, 0.8f, CVAR_ARCHIVE)
 {
 	if (self < 0.f && self != -1.f)
 	{
@@ -341,7 +341,7 @@ CUSTOM_CVAR(Float, dimamount, -1.f, CVAR_ARCHIVE)
 		self = 1.f;
 	}
 }
-CVAR(Color, dimcolor, 0xffd700, CVAR_ARCHIVE)
+CVAR(Color, dimcolor, 0x000000, CVAR_ARCHIVE)
 
 void System_M_Dim()
 {


RicardoLuis0:
i like the idea of making things more readable, but the dimming here is way too strong, almost an entirely black screen (also list menus should be dimmed a lot less than option menus, that way the titlepic/titlemap are still visible)

the-phinet:
I don't think the titlepic needs to be visible since you can look at it when not in a menu. All it tale is not pressing enter (or hitting escape)

Accessibility is more important than aesthetics. Any dimming less than 80% does pass (with DOOM2 titlepic) accessibility guidelines.

| | white on red | white on yellow | red on red | red on yellow |
|---|---|---|---|---|
|90%|[AAA](https://www.whocanuse.com/?bg=170000&fg=c7c7c7&fs=28&fw=b)|[AAA](https://www.whocanuse.com/?bg=1a1a04&fg=c7c7c7&fs=28&fw=b)|[AA](https://www.whocanuse.com/?bg=170000&fg=db0000&fs=28&fw=b)|[AA](https://www.whocanuse.com/?bg=1a1a04&fg=db0000&fs=28&fw=b)|
|80%|[AAA](https://www.whocanuse.com/?bg=150303&fg=c7c7c7&fs=28&fw=b)|[AAA](https://www.whocanuse.com/?bg=231906&fg=c7c7c7&fs=28&fw=b)|[AA](https://www.whocanuse.com/?bg=150303&fg=db0000&fs=28&fw=b)|[AA](https://www.whocanuse.com/?bg=231906&fg=db0000&fs=28&fw=b)|
|70%|[AAA](https://www.whocanuse.com/?bg=200505&fg=c7c7c7&fs=28&fw=b)|[AAA](https://www.whocanuse.com/?bg=352509&fg=c7c7c7&fs=28&fw=b)|[AA](https://www.whocanuse.com/?bg=200505&fg=db0000&fs=28&fw=b)|[FAIL](https://www.whocanuse.com/?bg=352509&fg=db0000&fs=28&fw=b)|
|60%|[AAA](https://www.whocanuse.com/?bg=2b0606&fg=c7c7c7&fs=28&fw=b)|[AAA](https://www.whocanuse.com/?bg=47320d&fg=c7c7c7&fs=28&fw=b)|[AA](https://www.whocanuse.com/?bg=2b0606&fg=db0000&fs=28&fw=b)|[FAIL](https://www.whocanuse.com/?bg=47320d&fg=db0000&fs=28&fw=b)|
|50%|[AAA](https://www.whocanuse.com/?bg=360808&fg=c7c7c7&fs=28&fw=b)|[AAA](https://www.whocanuse.com/?bg=583e10&fg=c7c7c7&fs=28&fw=b)|[AA](https://www.whocanuse.com/?bg=360808&fg=db0000&fs=28&fw=b)|[FAIL](https://www.whocanuse.com/?bg=583e10&fg=db0000&fs=28&fw=b)|

To me (someone with good eyesight), 60% dimming is already basically unreadable. 70% is tolerable, but I can only imagine how someone with poor colour differentiation would struggle to read.

image

the-phinet:
Actually, looking at the titlepic in the main menu. When the menu is visible, you can barely see it as is?

image

Dimming the screen is not obscuring the art any more than the text already is.

And again: you can just hit escape to hit the text and remove the dimming

mc776:
Is the dim amount not customizable?

the-phinet:
> Is the dim amount not customizable?

Yup! If a user's personal preference is to have less intense dimming (or the weird old yellow dim thing), they can do that.

Point is that the default should be accessible to all, not just the able bodied.

RicardoLuis0:
i think maybe 60% would be better as a default

mc776:
Seconding 60... of all those screenshots it's the one where it's dark enough that the menu is plainly visible but I can still make out the contents of the dimmed part without straining

SeventhSentinel:
I personally run 50% dimming with flat black as the dim color. 60-70% dimming looks good to me, too.

TehRealSalt:
+1 to 80% personally

TehRealSalt (approved):

Boondorl:
I like 80%, however, an exception should be made for the graphics options menu. That should have no dimming until we can get a better solution rolling (I believe menus can opt of out dimming manually in MENUDEF)

the-phinet:
For #35, since The screen cannot be dimmed during calibration, I implemented a dimming function just for the menu text. How do we feel about me adding this to menus in general, having 50% dimmed everywhere, and a further 70% dim under the text? This second dim thing could be customizable in the accessibility option

image

the-phinet:
> I believe menus can opt of out dimming manually in MENUDEF

I don't see any way to do this outside of making a zscript menu. I could add a check in the menu renderer to disable dimming of the current menu is the video menu I guess. I don't think that would be ideal

RicardoLuis0:
> For #35, since The screen cannot be dimmed during calibration, I implemented a dimming function just for the menu text. How do we feel about me adding this to menus in general, having 50% dimmed everywhere, and a further 70% dim under the text? This second dim thing could be customizable in the accessibility option
> 
> image

making it dimmer near text is nice, but still way too dark, how does it look like with 40% on background and another 40% on the text?

Boondorl:
> I don't see any way to do this outside of making a zscript menu. I could add a check in the menu renderer to disable dimming of the current menu is the video menu I guess. I don't think that would be ideal

Whoa, really? That's actually incredibly lame because it literally has the field value for it lol. MENUDEF should be expanded to allow this

the-phinet:
> Whoa, really? That's actually incredibly lame because it literally has the field value for it lol. MENUDEF should be expanded to allow this

I mean, I very well might be wrong. I can't find docs for it (only for zscript menus)

> making it dimmer near text is nice, but still way too dark, how does it look like with 40% on background and another 40% on the text?

Lol, I thought you liked 60% everywhere a moment ago 😅 I'll take a look at what values pass accessibility guidelines with a double dim.

---

In any case, I'm going to merge this before further bikeshedding can happen. I'll add the double dim thing in another pr, and we'll see what to do in the display menu at that point

jackrjli:
The default dimamount and dimcolor are set on a per-game basis in [MAPINFO](https://zdoom.org/w/index.php?title=MAPINFO/GameInfo_definition); for example, if you play Heretic then you would notice it uses a bluish dim rather than Doom's yellowish dim. The `dimamount` and `dimcolor` CVARs override the MAPINFO definitions if `dimamount` is set to a value other than -1, so changing the `dimamount` default as done here means that by default the MAPINFO settings are not respected. If you want to change the default for Doom, I would recommend doing so in [its MAPINFO](https://github.com/UZDoom/UZDoom/blob/047942e216be21adae73d7a8e8bc1f34ecd0f275/wadsrc/static/mapinfo/doomcommon.txt#L39-L40) rather than here.

the-phinet:
For the sake of accessibility, not respecting MapInfo is a good thing.

Once the double dim thing gets implemented, going back to `-1` is probably a good thing. That way the text can have a basically black background, and the rest of the screen can be whatever the mod author wants

jackrjli:
If you change the cvar default, then people migrating from GZDoom or UZDoom 4.14.3 won't get that change because they already have the cvar in their config file. New users would get the new default, but then changing the default back to -1 later on wouldn't do anything for them. Making the change in MAPINFO would cover both those cases; it would only not cover PWADs or standalone IWADs that set the dim to an obnoxious value (do people do that?) if you want to block that.

RicardoLuis0:
> For the sake of accessibility, not respecting MapInfo is a good thing.
> 
> Once the double dim thing gets implemented, going back to `-1` is probably a good thing. That way the text can have a basically black background, and the rest of the screen can be whatever the mod author wants

the issue is that white text on a black background is fine (but still somewhat eye-straining), but non-white text (ex. red) on a black background is genuinely eye-straining, at least for me

PR #749, opened 2025-12-15
Fix building on VS2026

dwing4g:
Fix a compiling error on Visual Studio 2026:
```
rapidxml.hpp(385,24): error C2059: syntax error: '<parameter-list>'
```
Maybe it's a VS bug, but we can avoid it easily.

### DIFF
--- a/src/common/thirdparty/rapidxml/rapidxml.hpp
+++ b/src/common/thirdparty/rapidxml/rapidxml.hpp
@@ -382,8 +382,8 @@ namespace rapidxml
     public:
 
         //! \cond internal
-        typedef void *((alloc_func)(std::size_t));       // Type of user-defined function used to allocate memory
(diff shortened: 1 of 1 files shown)


the-phinet (approved):
Looks fine to me.

Does the issue still persist if we instead pull in the updates from upstream? If doing that fixes things (and doesn't cause other issues), I would prefer that.

dwing4g:
The official website of RapidXML is https://rapidxml.sourceforge.net/, that appears to have been unmaintained for long time. While there are several personal forks on GitHub, none of them are considered the true upstream repository.

GZDoom had also previously attempted to fix this line of code, but the issue was not completely resolved, and it was further exposed by Visual Studio 2026.

This specific line of code is intended to define a function type, not a function pointer. Therefore, the function name should not be enclosed in parentheses. Although adding parentheses might be handled without error by some compilers, it represents a potential underlying pitfall.

PR #755, opened 2025-12-16
make r_extralight rescale the 0-255 range into r_extralight-255 instead of simply adding

RicardoLuis0:
this lets the range of values it can work with be greatly extended without harming lighting effects on maps

### DIFF
--- a/src/rendering/swrenderer/scene/r_light.cpp
+++ b/src/rendering/swrenderer/scene/r_light.cpp
@@ -148,7 +148,7 @@ namespace swrenderer
 
 	fixed_t LightVisibility::LightLevelToShadeImpl(RenderViewport *viewport, int lightlevel, bool foggy)
 	{
-		lightlevel = hw_ClampLight(lightlevel);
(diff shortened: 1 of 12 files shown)


RicardoLuis0:
(also changes the allowed range from `-8, +8` to `-64, +128`)

src/rendering/swrenderer/scene/r_light.cpp
```diff
@@ -147,7 +148,7 @@ namespace swrenderer
 
 	fixed_t LightVisibility::LightLevelToShadeImpl(RenderViewport *viewport, int lightlevel, bool foggy)
 	{
-		lightlevel += r_extralight;
+		lightlevel = hw_ClampLight(lightlevel);
```
Boondorl: Is this safe? The original sw renderer doesn't clamp to 255, at least not this early in, that seemed to be a hw specific thing. I'm not sure what knock on effects this could have

RicardoLuis0: removed the clamp from software mode

Boondorl (approved):

Issue #726, opened 2025-12-11
UZDoom Multiplayer Improvements


MegaShinySnivy:
### Game

All

### Problem Description

At present, our group cannot play multiplayer Doom together effectively due to Zdoom's reliance on P2P networking, and said P2P networking breaking down over long distances. Even when I set Netmode 1*, our group faces heavy lag, and its normally only a few minutes before someone gets out of sync, even with no mods. These problems seemed to have become worse when we migrated over to UZdoom 4.X.X, but I am not confident enough to say that is the definitive issue. 

(*netmode 1 is required as some of our members cannot do port forwarding due to their ISPs not implementing IPV6 and instead forcing a CGNAT. in other news, I hate the internet.)

### Feature Description

A more stable and latency-insensitive multiplayer implementation would make long range dooming more possible, and allow my group to get back together. Potentials include a server-client architecture like most games, meaning netmode 1 would no longer be needed, as well as lowering the lockstep syncing requirements for doom itself. 

I am not familiar with UZdoom's networking enough to determine if this is a good suggestion, however a theoretical fix I could think of is having a server be the authoritative source for the current game state. If a player is out of sync, force their game back into sync using data from the running game on the server. 

This may break compatibility with some mods that rely on the lockstep nature of Doom, so I suggest it is added as either an optional command line flag or an in-game option.

the-phinet:
Is this feedback based on a `4.x` release, or a `5.x` devbuild?

MegaShinySnivy:
4.x release! My apologies, I should have put that in the original.

MegaShinySnivy:
I've updated the issue with additional details.

RicardoLuis0:
if so, you should try multiplayer with 5.0 devbuilds instead, there has been an entire rewrite of the netcode, and you'll find many of your issues already addressed

RicardoLuis0:
(basically the issue wasn't P2P, it was shit netcode lol - 5.0 is still P2P but runs smooth as butter)

MegaShinySnivy:
I'll try it out! Do you know where I can grab a copy of these builds for windows and linux? We have players on both systems.

the-phinet:
You can grab recent build artifacts from CI. Keep in mind they're very "here be dragons", so there may be issues. Make sure all users are using builds from the same commit

https://github.com/UZDoom/UZDoom/actions/workflows/continuous_integration.yml?query=event%3Apush

MegaShinySnivy:
[...]

PR #760, opened 2025-12-16
Don't predict if the player's pawn hasn't called PostBeginPlay yet

### LINKED ISSUE #748: [Multiplayer] Host peer's PlayerPawn not reading user cvar from its PostBeginPlay()
### UZDoom Version

UZDoom-Begin-251-g0712d1d51 (trunk)

### Game

All

### Operating System

Linux x86_64

### OS Version

Debian 11

### Hardware

N/A

### Duplicate Confirmation

### Bug Description

I define a `User Int` `CVar` in a `CVARINFO` lump, define a custom `PlayerPawn`, and use `CVar.GetCVar("varname", player)` to get a pointer in its `PostBeginPlay()`, and try to access it using `GetInt()` in its `Tick()`.

The "host" peer (and only the host peer) aborts with the message:

`VM execution aborted: tried to read from address zero. In function parameter "self"`

See "Replication" section for minimal example pk3. There are no problems in singleplayer mode.

### Replication

Launch a 2-player multiplayer game on localhost with the `Doom2.wad` iwad and this mod: 

[multiplr_cvartest.zip](https://github.com/user-attachments/files/24168437/multiplr_cvartest.zip)

Join it from localhost using another instance. The "host" instance should abort and show the error. The "guest" instance will continue running.

### Configuration

```ini

```

### Log

_No response_

Boondorl:
This does some important networking and therefore should not be predicted at all. Instead, simply don't predict until this is handled properly.

Fixes #748

### DIFF
--- a/src/playsim/p_user.cpp
+++ b/src/playsim/p_user.cpp
@@ -1480,6 +1480,7 @@ void P_PredictPlayer (player_t *player)
 	if (demoplayback || gamestate != GS_LEVEL ||
 		player->mo == NULL ||
 		player != player->mo->Level->GetConsolePlayer() ||
+		(player->mo->ObjectFlags & OF_JustSpawned) ||
 		(player->cheats & CF_PREDICTING))
 	{
 		return;


Boondorl:
@mc776 I'm wondering if this is also related to HD desyncing instantly on 5.0. Might be worth seeing if this fixes it since the linked issue was indeed replicable locally

TehRealSalt (approved):

mc776:
> @mc776 I'm wondering if this is also related to HD desyncing instantly on 5.0. Might be worth seeing if this fixes it since the linked issue was indeed replicable locally

I had the same thought when I saw this in my notifs.

I can't seem to tag [MegaShinySnivy](https://github.com/MegaShinySnivy) directly so I'll mention #726 here.

MegaShinySnivy:
@mc776 I'll try with this commit as soon as I can!

PR #781, opened 2025-12-19
Fixes #706 and other OOB exceptions thrown when Argv FString is empty…

### LINKED ISSUE #706: OOB with invalid -iwad input
### UZDoom Version

trunk

### Game

All

### Operating System

Windows 10

### OS Version

_No response_

### Hardware

_No response_

### Duplicate Confirmation

### Bug Description

There is an out-of-bounds error when starting the engine like this:

```
uzdoom.exe -iwad ""
```

(yes, with the quotes after `-iwad`)

A user who's running their projects from within SLADE reported this, apparently that is what SLADE passes to the engine if SLADE's "base resource" is set to None

### Replication

_No response_

### Configuration

```ini

```

### Log

_No response_

TorrentialFire:
When an empty string is passed as an argument value to the engine, `m_argv.cpp` attempts to index into the string without checking its length, and an exception is thrown.

Assuming this is desired behavior for sniffing out other places in the engine where FStrings are indexed OOB, I have added the appropriate check to FArgs::CheckValue, which short-circuits the check if the corresponding FString in Argv is empty.

EDIT: Dang, my commit message is wrong. This was introduced earlier than the commit I mentioned. Looking in a bit more detail...

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -252,3 +252,4 @@ Heath
 Carlos "Cardboard Marty" Sanchez
 JeremWatts
 Faizan "IndieSunPraiser" Mughal
+William "TorrentialFire" Merzon-Wilson
--- a/src/common/utility/m_argv.cpp
+++ b/src/common/utility/m_argv.cpp
@@ -419,7 +419,7 @@ const char * FArgs::CheckValue(const FArg check) const
 	if (i > 0 && i < (int)Argv.Size() - 1)
 	{
 		i++;
-		return Argv[i][0] != '+' && Argv[i][0] != '-' ? Argv[i].GetChars() : nullptr;
+		return Argv[i].IsNotEmpty() && Argv[i][0] != '+' && Argv[i][0] != '-' ? Argv[i].GetChars() : nullptr;
 	}
 	else
 	{


TorrentialFire:
Apologies for the errant commit callout in my commit message. This was introduced by 605d627 which was explicitly to add bounds checking to FString.

the-phinet (dismissed):
Looks good to me.

the-phinet:
I think this is your first contribution! Can you do me a favour and add your name to the bottom of the CONTRIBUTORS file?

TorrentialFire:
Yeah, a pretty straightforward fix. I did test locally and confirmed the iwad selection dialog appears as you would expect, instead of the exception message as Nash described.

I have updated the CONTRIBUTORS file and took the opportunity to correct the previous commit message.

the-phinet (dismissed):

TorrentialFire:
Hmm, I would have liked Github to rebase my forked `trunk` instead of creating a merge commit.

the-phinet:
Sorry, could you do us another favour and rebase instead of merging to sync? We try to keep the commit history linear

TorrentialFire:
Everything should be linear now. 😄

the-phinet (approved):

PR #783, opened 2025-12-19
Fixes #778 setting nointerscrollabort now correctly prevents skipping…

### LINKED ISSUE #778: nointerscrollabort doesn't do what I suspect it's meant to
### UZDoom Version

4.14.3 and begin 257

### Game

All

### Operating System

Windows 10

### OS Version

_No response_

### Hardware

_No response_

### Duplicate Confirmation

### Bug Description

This could just be me misinterpreting what the cvar is meant to do.  I assumed that if "Allow skipping of intermission scrollers" ( `nointerscrollabort` ) was set to "off" then it wouldn't be possible to skip the bunny scroller at the end of Doom E3M8 (or any other scroller) by pressing the use key or fire key.

However, regardless of how I set this option, if I go to e3m8, kill the spiderdemon and go to the bunny scroller, as soon as it starts, I can press use/fire and the scroller skips to the end.  Also tested with the enddemon scroller in Heretic.

I've checked my ini file and `nointerscrollabort` is changing between true and false when I change the menu option.
So, either it's not working, or I've misunderstood what it's meant to do (sorry for the false report if that's the case).

### Replication

1. Load up Doom
2. go to e3m8
3. kill the spiderdemon
4. click through the exit text until the bunny scroller starts
5. hit use or fire
6. the game will skip to the end of the scrolling animation
7. change the `nointerscrollabort` to the opposite value
8. repeat the above
9. you will get the same behaviour as in 6.

### Configuration

```ini

```

### Log

_No response_

TorrentialFire:
This fix allows `nointerscrollabort` to actually prevent a keypress event from advancing a scrolling intermission screen (e.g. the E3M8 bunny cutscene). If this was the intended behavior of that setting, it hasn't worked correctly in a long time.

~~This fix removes a bit of behavior that I do not believe had any effect. If `nointerscrollabort` was false and a keypress event was sent, the intermission controller advanced the screen anyway, so setting the background and adjusting the ticker wouldn't do much. I may be misunderstanding. I am open to suggestions.~~

### DIFF
--- a/src/intermission/intermission.cpp
+++ b/src/intermission/intermission.cpp
@@ -803,11 +803,18 @@ void DIntermissionScreenScroller::Init(FIntermissionAction *desc, bool first)
 int DIntermissionScreenScroller::Responder (FInputEvent *ev)
 {
 	int res = Super::Responder(ev);
-	if (res == -1 && !nointerscrollabort)
+
+	if (res == -1)
 	{
+		if (nointerscrollabort) {
+			// Consume the key press to prevent advancing the scroller
+			return 0;
+		}
+
 		mBackground = mSecondPic;
 		mTicker = mScrollDelay + mScrollTime;
 	}
+	
 	return res;
 }
 


TorrentialFire:
I realized that if the scrolling intermission screen is the final intermission screen, then removing the ticker and background setting behavior when `nointerscrollabort` is false would be a bug. So, I reintroduced that behavior alongside the fix for when `nointerscrollabort` is true.

RicardoLuis0:
one thing, please don't reference issues directly in the commit messages, they create notification spam because of github

RicardoLuis0:
ex.
image

TorrentialFire:
Sure! No problem. 🫡

Boondorl (approved):

PR #788, opened 2025-12-19
Improved LastSafePos logic

Boondorl:
Now validates the spawn point is still safe when respawning in case the sector properties changed.  Added more checks for potential damaging effects when determining if a sector is safe or not.

Checking sector tags on the sector action for making it damage is still missing, but I consider this enough of an edge case to likely not be disruptive.

Needs in-game testing + testing to make sure old saves won't break when loading.

### DIFF
--- a/src/g_game.cpp
+++ b/src/g_game.cpp
@@ -1397,7 +1397,7 @@ void FLevelLocals::PlayerReborn (int player)
 	p->cheats |= chasecam;
 	p->Bot = Bot;			//Added by MC:
 	p->settings_controller = settings_controller;
-	p->LastSafePos = p->mo->Pos();
+	p->LastSafePos.Update(*p->mo, true);
 
 	p->oldbuttons = ~0, p->attackdown = true; p->usedown = true;	// don't do anything immediately
 	p->original_oldbuttons = ~0;
--- a/src/g_level.cpp
+++ b/src/g_level.cpp
@@ -1867,7 +1867,10 @@ int FLevelLocals::FinishTravel()
 	for (size_t i = 0u; i < MAXPLAYERS; ++i)
 	{
 		if (PlayerInGame(i) && !(Players[i]->mo->ObjectFlags & OF_EuthanizeMe) && toCallBack.Find(Players[i]->mo) < toCallBack.Size())
+		{
+			Players[i]->LastSafePos.Update(*Players[i]->mo, true);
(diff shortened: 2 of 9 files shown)


inkoalawetrust:
Not directly related to the player respawn code itself but also exposing those IsDamaging(), IsCrushing() and IsDangerous() functions (Or at least the last one) to ZScript could be really useful for mods for both players and monsters.

Boondorl:
I plan to do it later on down the line once I have a better idea of how they should be structured. Atm it's a bit "here be dragons"

inkoalawetrust:
Sounds good, it could also help with making AVOIDHAZARDS actually do something because P_IsUnderDamage() is much more crude than your functions, since with the way Graf implemented the flag I'm pretty sure it just,, doesn't work at all (Sure didn't seem to when I tested it), the implementation for it like the [P_IsUnderDamage()](https://github.com/UZDoom/UZDoom/blob/50c40d74c4e4dbe129576dd9a54edae072a0f8d2/src/playsim/p_enemy.cpp#L434-L466) and [P_SmartMove()](https://github.com/UZDoom/UZDoom/blob/50c40d74c4e4dbe129576dd9a54edae072a0f8d2/src/playsim/p_enemy.cpp#L725-L763) seem to be just directly ripped from DSDA's implementation of [both](https://github.com/kraflab/dsda-doom/blob/c6fa4c30b4dbeda52bd777624d25bd408a1e7bba/prboom2/src/p_enemy.c#L319-L338) [functions](https://github.com/kraflab/dsda-doom/blob/c6fa4c30b4dbeda52bd777624d25bd408a1e7bba/prboom2/src/p_enemy.c#L493-L543)*, converted to C++, and touched up to fit with GZDoom's codebase.

*Except that looking at how DSDA does it again while sourcing the links above, while P_SmartMove returns DI_NODIR (So nothing), P_NewChaseDir() *does* do something, it's just [that block of code has been commented out since forever ago](https://github.com/UZDoom/UZDoom/blob/50c40d74c4e4dbe129576dd9a54edae072a0f8d2/src/playsim/p_enemy.cpp#L1035-L1046).
FWIW it seems like that code "for later implementation" could probably be reimplemented by making a new actor property and then doing Distance2D(), just with the property set to 0 for backwards compat instead of 128 like it is in DSDA. But that's for a separate PR, maybe.

the-phinet:
﻿
﻿

TehRealSalt (approved):

PR #794, opened 2025-12-20
[Win32] Try to trick DWM into keeping the window (and all background windows) running in "fullscreen" mode

madame-rachelle:
This tries to address some of the complaints that, when UZDoom is running in Fullscreen mode, the video freezes. This might still need some further tweaks before it actually works.


SanyaWaffles:
Tested this in my fork and it oddly fixes the Discord fullscreen streaming bug too!

RicardoLuis0 (approved):

the-phinet:
~~Aaaaaaaaaaaa crap sorry I hit the update with merge button~~

Fixed

PR #801, opened 2025-12-21
History

### LINKED ISSUE #32: [Meta] Remove deadnaming in the source code and git `blame` history
Many instances of deadnaming are still pervasive in copyright headers, old git commits, etc. Git provides a [`.mailmap`](https://git-scm.com/docs/gitmailmap) file, that helps handle multiple names and emails as belonging to the same user, this can be use to clean up Git `blame`s for the old commits. Copyright headers and code comments will need updating, to remove Heit's/Jay's/etc old names. Perhaps not a critical need, but I've always found it personally troublesome that these were never updated, even after years.

the-phinet:
Closes #32

Progress: Done :face_exhaling: 

Also taking the time to normalize the header style. Keeping Marisa's original style as described in empty.cpp


TorrentialFire:
This seems like a dull task to do alone. Would you like some help?

the-phinet:
> This seems like a dull task to do alone. Would you like some help?

It's just something I'm  doing between doing other work.

Sure :) Feel free to submit changes to my fork. I do intend to squash everything into on big commit, when everything is done.

Basically fix deadnames, update copyright notices (this is the tedious part, gotta look at the actual file history for every file...), and ensure the notices all follow the same format style.

For header files, since it's inconsistent if copyright notices exist or not, I am adding them to each one. ~~I think I may remove them all in a separate commit after this is all done.~~ No, too many of them contain actual code

Went and precomputed the first contributions of people for each source file. 2006-02-24 if the earliest the data goes:
- src: https://gist.github.com/the-phinet/07160d4b1e6b14cf554227d5232faa11
- wadsrc: https://gist.github.com/the-phinet/4ee271dd90ef08cf8ccb72174e7b9c5d

RicardoLuis0 (approved):

PR #811, opened 2025-12-23
Added default bots

Boondorl:
Adds 5 hardcoded bots for the purposes of playtesting if no cfg is found.

### DIFF
--- a/src/playsim/bots/b_game.cpp
+++ b/src/playsim/bots/b_game.cpp
@@ -522,7 +522,19 @@ bool FCajunMaster::LoadBots ()
 	tmp = M_GetCajunPath(BOTFILENAME);
 	if (tmp.IsEmpty())
 	{
-		DPrintf (DMSG_ERROR, "No " BOTFILENAME ", so no bots\n");
+		// Adds a few default bots that can always be summoned for testing.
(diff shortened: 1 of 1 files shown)


src/playsim/bots/b_game.cpp
```diff
@@ -522,7 +522,19 @@ bool FCajunMaster::LoadBots ()
 	tmp = M_GetCajunPath(BOTFILENAME);
 	if (tmp.IsEmpty())
 	{
-		DPrintf (DMSG_ERROR, "No " BOTFILENAME ", so no bots\n");
```
the-phinet, line 525: Maybe add an info message?

```cpp
DPrintf(DMSG_NOTIFY, "Adding default bot. Add " BOTFILENAME " in order to customize\n");
```

the-phinet (approved):

src/playsim/bots/b_game.cpp
Boondorl: I actually got tired of seeing this message every time a level loaded. I'll let you ponder the implications of that. This code is dreadful and needs to be chucked in its entirety, so for now I'd rather silence it seeing as the bots are quite literally broken if you actually play with them. This is basically just to make testing easier

PR #831, opened 2025-12-30
eval expression improvements

RicardoLuis0:
* allow referencing cvars in eval expressions using `$var`
* add `repeat` to run a command multiple times based on expression results
image


RicardoLuis0:
oh and also added `numeval` that only runs numeric logic instead of string logic

River-Salmon (approved):

PR #843, opened 2026-01-01
Properly networked item selection logic

Boondorl:
Weapon and flechette selection functions are play scoped, so they need to be called server side to prevent desyncs. This presented a similar issue to BobWeapon.

### DIFF
--- a/src/d_net.cpp
+++ b/src/d_net.cpp
@@ -2371,6 +2371,96 @@ static int RemoveClass(FLevelLocals *Level, const PClass *cls)
 	return removecount;
 
 }
+
+EXTERN_CVAR(Int, displaynametags)
+EXTERN_CVAR(Int, nametagcolor)
+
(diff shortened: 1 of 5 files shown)


RicardoLuis0:
possibly add a cvar (default on?) to play the weapon switch sounds for other players in mp? (as just a sound attached to the pawn)

Boondorl:
I'm not even sure what the intent of this feature is tbh since weapons actually have selection sounds you can build into them. I think it was meant to be a completely optional sound file so it's not even guaranteed everyone will have it/have the same one. The original sound actually played back statically, I just thought that was incredibly silly. In this case I'd prefer to leave it untouched because I really don't feel like pondering the logistics of the best way to handle this that doesn't break shit in deathmatch (someone could load up a file to easily tell when other players are switching weapons and tell where they are).

RicardoLuis0 (approved):

PR #848, opened 2026-01-02
Browser platform support

ololoken:
## Description
Adds wasm32 platform support. By default works in 32-bit mode with single thread.
Demo: https://turch.in/uzdoom+ashes2063/index.html?lang=en

Right now it's more PoC for discussion than PR ready to merge.

TODOs:

- [x] Sync FS callback to persist config changes and saves.
- [x] rendering, shaders and most annoying bugs.
- [x] fix save game screenshots
- [ ] Network.
- [x] Make tools work via nodejs to avoid requirement to build them manually.
- [ ] Add module launcher webpage with options to select WADs and PK3s.
- [ ] Build pipeline

## Checklist


uncreativeCultist:
image
probably the first person to run myhouse.wad inside of UZDoom on a web browser.
good work, would be cool to see this get merged

RicardoLuis0:
> TODOs:
> Sync FS callback to persist config changes and saves.
> Network.
> Make tools work via nodejs to avoid requirement to build them manually.
> Add module launcher webpage with options to select WADs and PK3s.

it'd probably also need porting to 64-bit wasm before merging as was mentioned in the forum thread, but really, amazing work on this 🎉

CMakeLists.txt
```diff
@@ -50,6 +50,21 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 find_package(Threads REQUIRED)
 find_package(Python3 3.6 REQUIRED)
 
+if(CMAKE_SYSTEM MATCHES "Emscripten")
```
RicardoLuis0: would `if(EMSCRIPTEN)` work here? as far as i can look online it seems that the variable is defined if emscripten is the platform

RicardoLuis0: (same for anywhere else that checks `CMAKE_SYSTEM MATCHES "Emscripten"`)

src/common/platform/posix/sdl/sdlglvideo.cpp
```diff
@@ -194,6 +194,11 @@ namespace Priv
 
 	void SetupPixelFormat(int multisample, const int *glver)
 	{
+#ifdef __EMSCRIPTEN__
+		SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 2);
+		SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 0);
+		SDL_GL_SetAttribute(SDL_GL_ALPHA_SIZE, 0);
+#else
```
RicardoLuis0: this should probably be done by forcing gles2 on browser, rather than `#ifdef`-ing the version, that way if GLES3 support is ever added to the opengl3 backend (as a delta touch backport possibly?), webgl2 can be used with full user shader support

src/d_main.cpp
```diff
@@ -1418,75 +1422,83 @@ void D_ErrorCleanup ()
...
-	// Clamp the timer to TICRATE until the playloop has been entered.
-	r_NoInterpolate = true;
-	Page.SetInvalid();
-	Subtitle = nullptr;
-	Advisory.SetInvalid();
-
-	vid_cursor->Callback();
-
-	for (;;)
+	int lasttic = *reinterpret_cast<int*>(lasttic_);
```
RicardoLuis0: shouldn't this be `int &lasttic`? that way any assignments to it don't need code changes

src/rendering/hwrenderer/scene/hw_bsp.cpp
```diff
@@ -40,14 +40,20 @@
 #include <immintrin.h>
 #endif // ARCH_IA32
 
+#if defined(__EMSCRIPTEN_PTHREADS__) || !defined(__EMSCRIPTEN__)
 CVAR(Bool, gl_multithread, true, CVAR_ARCHIVE | CVAR_GLOBALCONFIG)
+#else
+CVAR(Bool, gl_multithread, false, CVAR_ARCHIVE | CVAR_GLOBALCONFIG)
```
RicardoLuis0: if multithreading isn't available, this should be:
```
CVAR(Bool, gl_multithread, false, CVAR_NOSET)
```
to disallow users from changing it

src/common/rendering/gles/gles_shader.cpp
```diff
@@ -235,16 +235,21 @@ bool FShader::Load(const char * name, const char * vert_prog_lump_, const char *
 	FString	proc_prog_lump = proc_prog_lump_;
 	FString light_fragprog = light_fragprog_;
 
-	vert_prog_lump.Substitute("shaders/", "shaders_gles/");
-	frag_prog_lump.Substitute("shaders/", "shaders_gles/");
-	proc_prog_lump.Substitute("shaders/", "shaders_gles/");
-	light_fragprog.Substitute("shaders/", "shaders_gles/");
+	const char *shaders_folder = gles.glesMode == GLES_MODE_WEBGL
+		? "shaders_webgl/"
+		: "shaders_gles/";
```
RicardoLuis0, line 240: are the shaders different enough that they need duplicating? if the changes are small enough, it'd be nicer if this was possible to just do it with a define plus ifdefs in the gles shaders themselves

CMakeLists.txt
ololoken: Done

src/common/rendering/gles/gles_shader.cpp
ololoken: Major changes are related to in/out keywords, texture2d -> texture, absence of declared gl_FragColor and some types mismatch. Unfortunately webgl glsl v100 is far from gles v100 in terms of available features, so I had to migrate to glsl v300 es

src/rendering/hwrenderer/scene/hw_bsp.cpp
ololoken: Thanks

src/d_main.cpp
ololoken: Yep, just forgot about it.

src/common/platform/posix/sdl/sdlglvideo.cpp
ololoken: GL context creation is deep in js module guts generated by emscripten. So it's better to set it from "native" code.

src/common/rendering/gles/gles_shader.cpp
ololoken: It's possible there is option to update original gles shaders version, but I'm not ready advice it.

RicardoLuis0: > Major changes are related to in/out keywords, texture2d -> texture, absence of declared gl_FragColor and some types mismatch. Unfortunately webgl glsl v100 is far from gles v100 in terms of available features, so I had to migrate to glsl v300 es

yeah, that's a fair reason for separate shaders 👍

src/CMakeLists.txt
```diff
@@ -131,7 +131,7 @@ else()
 	endif()
 
 	# Non-Windows version also needs SDL except native OS X backend
-	if( NOT APPLE OR NOT OSX_COCOA_BACKEND )
+	if( NOT APPLE OR NOT OSX_COCOA_BACKEND AND NOT EMSCRIPTEN )
```
RicardoLuis0, line 104: the logic itself might not need it, but it'd be good for parenthesis to be used (since mixing OR and AND)

src/common/platform/posix/sdl/sdlglvideo.cpp
RicardoLuis0: > GL context creation is deep in js module guts generated by emscripten. So it's better to set it from "native" code.

yeah, but the ifdef'd out code already sets 2.0.0 for GLES in the very same function, that's what i'm confused about

src/common/rendering/r_thread.cpp
```diff
@@ -30,7 +30,11 @@
 #include "printf.h"
 #include <chrono>
 
+#ifdef __EMSCRIPTEN__
+CVAR(Int, r_multithreaded, 0, CVAR_ARCHIVE | CVAR_GLOBALCONFIG);
```
RicardoLuis0: ah, missed this when looking over the first time but this should also have `CVAR_ARCHIVE | CVAR_GLOBALCONFIG` changed to `CVAR_NOSET`

RicardoLuis0:
also this:

https://github.com/UZDoom/UZDoom/blob/trunk/src/common/audio/sound/oalsound.cpp#L639-L643

probably needs an ifdef to check for the emscripten version of openalsoft, or else a bogus warning will be printed

src/common/platform/posix/sdl/sdlglvideo.cpp
ololoken: To be honest I just didn't want to search SDL_GL_* attribute which is not supported and ruins context creation, later I will try hunt it.

ololoken:
> also this:
> 
> https://github.com/UZDoom/UZDoom/blob/trunk/src/common/audio/sound/oalsound.cpp#L639-L643
> 
> probably needs an ifdef to check for the emscripten version of openalsoft, or else a bogus warning will be printed

It's better to check if AL_VENDOR == "Emscripten"

RicardoLuis0:
> > also this:
> > https://github.com/UZDoom/UZDoom/blob/trunk/src/common/audio/sound/oalsound.cpp#L639-L643
> > probably needs an ifdef to check for the emscripten version of openalsoft, or else a bogus warning will be printed
> 
> It's better to check if AL_VENDOR == "Emscripten"

image

yeah, just something so that this warning doesn't print

kcat:
> yeah, just something so that this warning doesn't print

Certain things won't work without OpenAL Soft features, like sounds that use loop points (which don't cover the whole sound) or use stereo sound for 3D/world sounds, and "area sounds" may not sound correct. Some warning that the audio may sound wrong may be appropriate.

RicardoLuis0:
> > yeah, just something so that this warning doesn't print
> 
> Certain things won't work without OpenAL Soft features, like sounds that use loop points (which don't cover the whole sound) or use stereo sound for 3D/world sounds, and "area sounds" may not sound correct. Some warning that the audio may sound wrong may be appropriate.

afaik emscripten supports a ton of openal features

RicardoLuis0:
yep, loop points, spatial sounds, hrtf, etc
image
<https://emscripten.org/docs/porting/Audio.html>

Lunascaped:
not sure if its just me but i get a lot of weird mouse snapping with this, would be cool if it was fixed before merging.

https://github.com/user-attachments/assets/d226d5db-8f30-4d1c-a639-8acfb467d5ee

ololoken:
Demo updated. Fixed colors and optimised textures loading.
There is one thing I want to fix:
image

It's related to this color https://github.com/UZDoom/UZDoom/blob/trunk/src/common/statusbar/base_sbar.cpp#L592
when commented everything is ok.
I suspect PalEntry
```cpp
struct
{
	uint8_t b,g,r,a; // here due to texture format was changed?
};
uint32_t d;
```

RicardoLuis0:
odd, isn't wasm little-endian? that's the only thing that would mess up how palentry values are passed i'd reckon, but i guess something else is breaking it

RicardoLuis0:
btw, does this also break in 64-bit wasm?

Lunascaped:
The mouse jumping still happens sadly so this is still kinda unplayable for me

ololoken:
> btw, does this also break in 64-bit wasm?

Yep, behaviour is the same. I found that removing of this line https://github.com/UZDoom/UZDoom/blob/trunk/src/common/2d/v_2ddrawer.cpp#L370 fixes the problem. Will try to figure out the reason tomorrow. And this issue with save game screenshot is also for tomorrow or even next weekend 
image


> The mouse jumping still happens sadly so this is still kinda unplayable for me

Well, I suppose it's some browser-related problem. I'm testing build playing brutality mod and have fun.
image

Lunascaped:
Weird, I tried using a chrome browser  with no extensions other than uBlock Origin and it still happens. However it seems to work perfectly fine on Firefox so idk.

ololoken:
ok, the problem with color is fixed here https://github.com/UZDoom/UZDoom/pull/848/commits/f5b473132385d210c9517750bfe7826dda065ada

RicardoLuis0:
also seems that to fix the `none` texture filtering modes it may need to unconditionally return true here for webgl: https://github.com/UZDoom/UZDoom/blob/trunk/src/d_main.cpp#L3045-L3048

ololoken:
> also seems that to fix the `none` texture filtering modes it may need to unconditionally return true here for webgl: https://github.com/UZDoom/UZDoom/blob/trunk/src/d_main.cpp#L3045-L3048

Would you give more input about none texture filtering? I'm not aware of this problem.

RicardoLuis0:
> > also seems that to fix the `none` texture filtering modes it may need to unconditionally return true here for webgl: https://github.com/UZDoom/UZDoom/blob/trunk/src/d_main.cpp#L3045-L3048
> 
> Would you give more input about none texture filtering? I'm not aware of this problem.

basically, uzdoom defaults texture filtering to none, but none texture filtering with mipmaps doesn't work on intel (and apparently webgl too) if anisotropic filtering is enabled

RicardoLuis0:
also it seems like the window on your preview (at https://turch.in/uzdoom+freedoom/index.html) isn't resizing itself properly on load? it needs resizing the window around for it to size the viewport properly, though idk if that's from the code itself or just some bug in the emscripten version of SDL

RicardoLuis0:
ex. resolution on load (stretched, lower-res) vs after de-maximizing/re-maximizing the window (proper full res)
image
image

RicardoLuis0:
but idk if this is an engine bug or just a bug with the preview website

ololoken:
> but idk if this is an engine bug or just a bug with the preview website

Yep there is some problem with actual html canvas size calculation. WxH size on load doesn't match with bounding client rect. Will try fix.

ololoken:
> > > also seems that to fix the `none` texture filtering modes it may need to unconditionally return true here for webgl: https://github.com/UZDoom/UZDoom/blob/trunk/src/d_main.cpp#L3045-L3048
> > 
> > 
> > Would you give more input about none texture filtering? I'm not aware of this problem.
> 
> basically, uzdoom defaults texture filtering to none, but none texture filtering with mipmaps doesn't work on intel (and apparently webgl too) if anisotropic filtering is enabled

I tried different combinations of textures filtering and anisotropic filter and didn't notice glitches. Sorry, can run only wasm version of zdoom and have no references. Proper scale on boot is fixed, will update demo later.

RicardoLuis0:
> > > > also seems that to fix the `none` texture filtering modes it may need to unconditionally return true here for webgl: https://github.com/UZDoom/UZDoom/blob/trunk/src/d_main.cpp#L3045-L3048
> > > 
> > > 
> > > Would you give more input about none texture filtering? I'm not aware of this problem.
> > 
> > 
> > basically, uzdoom defaults texture filtering to none, but none texture filtering with mipmaps doesn't work on intel (and apparently webgl too) if anisotropic filtering is enabled
> 
> I tried different combinations of textures filtering and anisotropic filter and didn't notice glitches. Sorry, can run only wasm version of zdoom and have no references. Proper scale on boot is fixed, will update demo later.

basically any texture filtering that says "none" shouldn't look blurry, but if it's bugged that only works for plain "none", and not for "none (whatever)" (the part in parenthesis means it's mip-mapped) when anisotropic filtering is enabled

RicardoLuis0:
bugged vs correct:
image
image

ololoken:
@RicardoLuis0 I noticed this huge commented block https://github.com/UZDoom/UZDoom/blob/trunk/src/common/rendering/gles/gles_samplers.cpp#L176
is it connect somehow to the filters issue we discussing?

RicardoLuis0:
> @RicardoLuis0 I noticed this huge commented block https://github.com/UZDoom/UZDoom/blob/trunk/src/common/rendering/gles/gles_samplers.cpp#L176
> is it connect somehow to the filters issue we discussing?

nah, that's the non-gles texture filtering choice, gles does it always every time it binds a texture IIRC, idk why it was left there as a comment tbh

ololoken:
> nah, that's the non-gles texture filtering choice, gles does it always every time it binds a texture IIRC, idk why it was left there as a comment tbh

hmm, it appears in webgl LOD is calculated strictly, and driver don't "magically" sharpen mip. 
Well, potentially it can be fixed with manual sharpening in shader.

RicardoLuis0:
> > nah, that's the non-gles texture filtering choice, gles does it always every time it binds a texture IIRC, idk why it was left there as a comment tbh
> 
> hmm, it appears in webgl LOD is calculated strictly, and driver don't "magically" sharpen mip. Well, potentially it can be fixed with manual sharpening in shader.

it's not about magically sharpening, it's about not respecting the mag filter, those mipmapped none modes, the magnification filter is none, and the minification filter is what's in parenthesis, but in bugged drivers like intel/webgl, when aniso filtering is enabled it just forces the min filter for everything, leaving shit blurry

RicardoLuis0:
you can see here:
https://github.com/UZDoom/UZDoom/blob/17b4f76c7f92c78789fa09467cda3a8eef4e6146/src/common/rendering/gl/gl_hwtexture.cpp#L41-L49

the four "none" modes are:
* none - index 0 - min GL_NEAREST, mag GL_NEAREST
* none (nearest) - index 1 - min GL_NEAREST_MIPMAP_NEAREST, mag GL_NEAREST
* none (linear) - index 5 - min GL_NEAREST_MIPMAP_LINEAR, mag GL_NEAREST
* none (trilinear) - index 6 - min GL_LINEAR_MIPMAP_LINEAR, mag GL_NEAREST

in bugged driversm, with aniso enabled they become:
* none - index 0 - min GL_NEAREST, mag GL_NEAREST
* none (nearest) - index 1 - min GL_NEAREST_MIPMAP_NEAREST, mag GL_NEAREST_MIPMAP_NEAREST 
* none (linear) - index 5 - min GL_NEAREST_MIPMAP_LINEAR, mag GL_NEAREST_MIPMAP_LINEAR
* none (trilinear) - index 6 - min GL_LINEAR_MIPMAP_LINEAR, mag GL_LINEAR_MIPMAP_LINEAR

(or possibly all but index 0 becomes GL_LINEAR_MIPMAP_LINEAR? hard to tell what the drivers are doing exactly)

ololoken:
yeah I agree it's a bug. Just want to avoid feature disabling based on vendor. My chromium based browser produces this:
image
image
image

Can you try this https://turch.in/webgl-mipmaps-explainer/ ? I'm curious what params of filtering/mipmap will work for you.

RicardoLuis0:
> Can you try this https://turch.in/webgl-mipmaps-explainer/ ? I'm curious what params of filtering/mipmap will work for you.

hmm, they all seem to work fine on that website, odd

RicardoLuis0:
maybe simply the way the renderer is setting the filters doesn't work properly for webgl?

RicardoLuis0:
also both firefox and chromium use the same library for webgl ([ANGLE](https://github.com/google/angle)) so it shouldn't differ between browsers

ololoken:
I updated demo https://turch.in/uzdoom+ashes2063/index.html?lang=en
Would you try filter settings in this build?

inkoalawetrust:
I tried them, and it seems only "None" is unblurred. All the other "None" filters blur the UI and world. The PSprites seem to be the only thing that is properly affected.

ololoken:
ok, then report your browser and os, lets collect some stats.
on my m1 tahoe 26.2 Chromium Version 142.0.7444.267 textures are not blured.
same safari
firefox 146.0.1 (aarch64) wasm start took ages(), but works smooth. No blur.

all browsers report devicePixelRatio = 2

inkoalawetrust:
I'm on Windows 10 22H2 build 19045.3324 and using Chrome Version 143.0.7499.170.

RicardoLuis0:
still blurred here as well, windows 10 22H2 19045.6691, blurred on both firefox and chromium

ololoken:
In browser's dev console type `devicePixelRatio`. What value do you see?

RicardoLuis0:
image

ololoken:
Ok, we have a suspect, as in emulation mode when devicePixelRatio=1 I too see blurry textures.
It's not a problem to disable aniso filters for devices with DPR<=1.

RicardoLuis0:
the pixel ratio probably hides the filtering

kcat:
> basically, uzdoom defaults texture filtering to none, but none texture filtering with mipmaps doesn't work on intel (and apparently webgl too) if anisotropic filtering is enabled

It would be hardware dependent, wouldn't it? Webgl is just passing the properties on to the system opengl driver, so if the gl driver doesn't have a problem with nearest+mipmaps+anisotropic filtering, webgl won't either. If Intel drivers/hardware does, then it wouldn't matter if it's webgl setting those properties or a native app.

I ran it on Firefox in Linux, with an AMD RX 580, and it all looks fine to me just like native UZDoom. `devicePixelRatio` reports `1`.

ololoken:
I'm not a guru in gl, but playing https://turch.in/webgl-mipmaps-explainer/ noticed that TEXTURE_MAX_ANISOTROPY_EXT affects only TEXTURE_MIN_FILTER or "far" textures. The only way to see anisotropy blur closer is to increase mip bias. 
Lesser devicePixelRatio relatively increases mip bias.
Assuming intel bug is 100% proven I can suggest to add detection unmasked gl(real driver name instead of abstract "WebKit") vendor here https://github.com/UZDoom/UZDoom/blob/trunk/src/d_main.cpp#L3045-L3048

RicardoLuis0:
> I'm not a guru in gl, but playing https://turch.in/webgl-mipmaps-explainer/ noticed that TEXTURE_MAX_ANISOTROPY_EXT affects only TEXTURE_MIN_FILTER or "far" textures. The only way to see anisotropy blur closer is to increase mip bias. Lesser devicePixelRatio relatively increases mip bias. Assuming intel bug is 100% proven I can suggest to add detection unmasked gl(real driver name instead of abstract "WebKit") vendor here https://github.com/UZDoom/UZDoom/blob/trunk/src/d_main.cpp#L3045-L3048

this isn't on intel hardware though (AMD GPU + AMD CPU), and standalone uzdoom has no issues with mipmapping at least for me, in any backend (GLES/OpenGL/Vulkan, all render things unblurred)

> It would be hardware dependent, wouldn't it? Webgl is just passing the properties on to the system opengl driver, so if the gl driver doesn't have a problem with nearest+mipmaps+anisotropic filtering, webgl won't either. If Intel drivers/hardware does, then it wouldn't matter if it's webgl setting those properties or a native app.

also i'm pretty sure ANGLE reimplements opengl with directx calls, at least on windows? it doesn't simply proxy through to the driver

RicardoLuis0:
> You can use anisotropic texture filtering in conjunction with linear texture filtering or mipmap texture filtering to improve rendering results. Your application enables anisotropic texture filtering by calling the [IDirect3DDevice9::SetSamplerState](https://learn.microsoft.com/en-us/windows/win32/api/d3d9helper/nf-d3d9helper-idirect3ddevice9-setsamplerstate) method. Set the value of the first parameter to the integer index number (0-7) of the texture for which you are selecting a texture filtering method. Pass D3DSAMP_MAGFILTER, D3DSAMP_MINFILTER, or D3DSAMP_MIPFILTER for the second parameter to set the magnification, minification, or mipmapping filter. Set the third parameter to D3DTEXF_ANISOTROPIC.

https://learn.microsoft.com/en-us/windows/win32/direct3d9/anisotropic-texture-filtering

seems that directx (at least 9) possibly plain doesn't support anisotropic filtering with nearest filtering? if so it'd need to do the intel workaround if the backing renderer is directx/direct3d-based

Ser-Gen:
I have macos sonoma 14.5, chrome 143.0.7499.193. I don't see any blur with "None (trilinear)" on both monitors with devicePixelRatio 1 and 2 at the current link

https://turch.in/uzdoom+ashes2063/index.html?lang=en

`devicePixelRatio  1`

image

`devicePixelRatio  2`

image

RicardoLuis0:
> I have macos sonoma 14.5, chrome 143.0.7499.193. I don't see any blur with "None (trilinear)" on both monitors with devicePixelRatio 1 and 2 at the current link

yeah, from all i can see it's a directx bug, so it'd be windows-only

babycommando:
Absolutely unbelievable. It works so damn well!!! Congratulations, this is absolutely legendary.

Just tested and its working like a charm. What is preventing this to be merged officially? Are you working on in something specific at the moment?

About networking, do you have a plan to approach it?

Thanks!!

RicardoLuis0:
> What is preventing this to be merged officially?

1. 5.0 is mostly under feature freeze for unplanned stuff
2. commit history is incredibly messy, so it is unmergeable and unreviewable as-is, would need cleaning up before it can be reviewed, and it obviously needs to be reviewed before it can be merged

babycommando:
Thanks @RicardoLuis0 for the kind response. Perhaps I can try helping making sense of stuff.

Honestly this is the only gzdoom/uzdoom wasm implementation working on the internet at the moment, and it works very well. Legendary works. Will be amazing to make all that ztuff run in the browser, open many possibilities for the future of videogames lol

@ololoken what are you working on at the moment?

ololoken:
@babycommando I'm working on Serios Sam.
Network implementation for browser requires infrastructure like running signalling server. We can emulate local network in browser using webrtc datachannels. You can find example how it's done here:
https://github.com/ololoken/xash3d-fwgs/tree/master/engine/platform/emscripten and https://github.com/ololoken/xash3d-launcher/blob/main/src/assets/module/vnet.ts

@RicardoLuis0 totally agree

I will continue wasm fork support.

Another demo:
https://turch.in/voxeldoom/index.html

babycommando:
Hey @ololoken, woah Voxel Doom <3

I just cloned your branch at https://github.com/ololoken/UZDoom/tree/emscripten-trunk and tried building locally, but got stuck around `libvpx` and `Zmusic` / `fluidsynth`.

Currently trying to wrap my brain around how you made a couple things there.
* have you built libvpx manually separately with emscripten, or used some prebuilt lib?
* how do you handle Zmusic fluidsynth and the audio deps stuff?
* do you first build native tools like lemon, re2c, zipdir, then run the wasm build?

Could you share how you usually build it? Perhaps you have the exact emcmake cmake commands or a rough script, if you could share the sequence it would be helpful.

The demos work beautifully, so I'm clearly missing something in the build setup.

Thanks again, this fork is seriously amazing!!

ololoken:
@babycommando it's supposed that you have to build 3rd party libraries using emscripten toolchain. Just `emmake make install` and in common case dependency will be automatically resolved
And yes for now build lemon, re2c and zipdir as native. I know there are ways to configure crosscompilation for such cases, but first want to try build them as node wasm modules .

babycommando:
@ololoken Got it, that makes sense now.
I did a clean emcmake configure. Native lemon/re2c/zipdir are already built and copied, and that part seems fine. But still haven't figured out libvpx and glib-2.0 required by Zmusic and fluidsynth.

So for your local build, do you just build/install libvpx and glib with emconfigure/emmake into the Emscripten environment, then rerun UZDoom cmake? Or do you avoid glib by patching bundled fluidsynth to use the glib stubs for EMSCRIPTEN? Also, for VPX, are you using upstream libvpx with emconfigure ./configure && emmake make install?

update: managed to compile libvpx locally so the project dont cry more about it, but glib-2.0 used by ZMusic still a problem, when trying to build it I'm getting: 
`subprojects/libffi/meson.build: Unsupported pair: system "emscripten", cpu family "wasm32"` 
Wonder how you got past that at all if you even built glib-2.0 in first place. Looks like libffi have an Emscripten port at hoodmane/libffi-emscripten repo. Have you used that? Yep, lets build the dependence of a dependence to see if it works cause why not, right haha. If UZDoom uses it can be helpful anyways I guess.

update 2: finally got to build glib-2.0 manually, I ended up indeed replacing libffi with that emscripten version and it worked. Also later on when compiling it, it also asked about building zlib manually. After that, I finally got glib-2.0 far enough to stop failing at configure.
Now UZDoom is actually compiling much deeper, but Im hitting a wave of C and Cpp portability issues: missing includes like cstring, algorithm, cstdint, C headers accidentally getting C++ includes, etc. So got some progress, but honestly its still pretty cursed. You said to just emmake make install, so I am probably doing something wrong. Spent the day trying to compile this lol.

update 3: got past the huge C/Cpp error wave! The mistake was building it with vulkan not disabled. Now I'm passing the flag -DHAVE_VULKAN=OFF and it seems to be building.

update 4: Phew, after a long time it finally compiled, I put it over some dumb html file to see what happens, initially it was only throwing a cpp pointer errors 🤣 but it was just me not mounting the wad correctly.

It worked!!! Music, performance, everything working just so well.
Captura de tela 2026-05-28 034156
freedoom2 voxelized man!

It is lovely to finally see this running. Also could you provide the html and js files you are hosting the wasm into?

Perhaps it would be way more helpful if you could just share an installation instruction of what to do from the moment I clone the repo to build and run it so we don't fall on those traps anymore. Also did everything on windows powershell which might have been an extra problem. Anyway this is legendary ::) Thanks!

babycommando (approved):
This is indeed solving the problem of having to do the hackish workaround for lemon, re2c and zipdir. Gracefully fixes that part, thanks. Perhaps we can make libvpx and FluidSynth (glib-2.0 -> libffi/zlib) build automatically as well as part of the next commits

RicardoLuis0:
taking a bit of a closer look at this, there's a lot of duplication, what's the difference between `shaders_webgl` and the regular shader files? if it's small enough it should be done as `#ifdef`s

ololoken:
> taking a bit of a closer look at this, there's a lot of duplication, what's the difference between `shaders_webgl` and the regular shader files? if it's small enough it should be done as `#ifdef`s

Good question. The duplication started as a safety measure — I didn't want WebGL work to risk regressions on the native GLES path. At this point, I think keeping separate shader trees is the better than #if / #else / #endif noise. Yes about 50% of the files are identical, but the split keeps that logic clean.

The two folders aren't just copies of the same GLSL with minor tweaks. They're different dialects with different CPU/GPU contracts on WebGL: RGBA texture upload with a .bgra swizzle in getTexel(), vec3 vertex positions promoted to vec4, and bone indices passed as vec4 instead of uvec4.

Merging into one tree would mean:

Wrapping essentially every I/O declaration and fragment output in version/platform guards
Relying on something beyond __VERSION__, since it can't tell WebGL (300 es) apart from desktop GL 3.x (330)
Adding new C++-injected defines and re-testing affected backend modes

RicardoLuis0:
> Wrapping essentially every I/O declaration and fragment output in version/platform guards Relying on something beyond **VERSION**, since it can't tell WebGL (300 es) apart from desktop GL 3.x (330) Adding new C++-injected defines and re-testing affected backend modes

that's still better IMO, it'd be a problem for the regular GL backend since those shaders can have custom user input, but keeping both in the same files for GLES makes sure any shader changes take the webgl version into consideration

babycommando:
@ololoken have you had any advances in the networked mode with webrtc datachannels? Do you believe this could be multiplayer at some point? Thanks! 
Also if possible, can you share your build process to make things more fluid? I ended up having to compile many dependencies of the dependencies by hand and stuff, lost some time on that. But it is working very well after I compiled everything (for hours lol)
[https://babycommando.itch.io/gzdoom-wasm](https://babycommando.itch.io/gzdoom-wasm)

ololoken:
@RicardoLuis0  I'd agree with merging if we were still doing this by hand. 
With llm in loop the cost of duplication has basically collapsed, while the cost of #if noise, harder reviews, and unintended regressions hasn't. The maintenance burden is no longer a real argument.

@babycommando Same pattern: compile all dependencies first, then compile uzdoom. Network support will be added in the future. For now, I only have time to periodically update merge request.

RicardoLuis0:
> @RicardoLuis0 I'd agree with merging if we were still doing this by hand.

i sure hope you're doing it by hand, we don't allow AI code https://github.com/UZDoom/UZDoom/blob/trunk/CONTRIBUTING.md#ai-generated-code

babycommando:
> Network support will be added in the future

This will be absolutely revolutionary for the game makers :]

Issue #585, opened 2025-11-25
Use the JSON serializer for rollback


Boondorl:
Currently the rollback method relies on a simple copy operation of the player. This is an extremely unsafe method of handling this due to Object references getting caught up in this and all kinds of things will break such as TArrays and TMaps. The serializer properly handles all of these things and, should it be used, will allow us to safely verify Object references as well and even recreate Objects as needed if they were destroyed.

This will still require correct processing of certain lists e.g. Behaviors to make sure lost references are dealt with properly (the level won't mark these so they'll eventually disappear, but it's better to make sure they're all properly unlinked). On top of this, the physics relinking will need a better strategy. The current one is dangerous and a "best fit" approach should instead be taken for getting inserted back into the lists.

Going forward this will make it much easier to apply rollback to _all_ Actors as the current rollback is heavily coupled to the player.

Issue #560, opened 2025-11-22
Crash when autosaving at map start


kcat:
### UZDoom Version

UZDoom version g4.15pre-820-gdf7ff511e

### Game

All

### Operating System

Linux x86_64

### OS Version

Debian Unstable

### Hardware

AMD

### Duplicate Confirmation

- [x] I have searched and not found a similar issue.

### Bug Description

Also a crash that has started happening with recent commits, and isn't consistently reproduceable. When a map starts, the game will create an autosave, which sometimes crashes. It might be an issue triggered by the mods I'm using since I just noclip-ran several unmodded doom2 levels without an autosave crash, but with these mods, it tends to happen within two level transitions. Regardless, as mentioned, it had been fine until recently. Attached crash log, which shows it happening during serialization:

```
#6  <signal handler called>
No locals.
#7  FName::GetChars (this=<optimized out>) at /home/kitty/projects/gzdoom/src/common/engine/serializer_internal.h:118
No locals.
#8  FSerializer::WriteObjects (this=this@entry=0x7fff5a04c780) at /home/kitty/projects/gzdoom/src/common/engine/serializer.cpp:652
        obj = 0x5589204ab8d0
        i = 678
#9  0x00005588f130df18 in FSerializer::WriteObjects (this=0x7fff5a04c780) at /home/kitty/projects/gzdoom/src/common/engine/serializer.cpp:840
        i = <optimized out>
        obj = <optimized out>
#10 FSerializer::GetCompressedOutput (this=this@entry=0x7fff5a04c780) at /home/kitty/projects/gzdoom/src/common/engine/serializer.cpp:792
        buff = <optimized out>
        compressbuf = <optimized out>
        stream = {next_in = 0x5588f2731e18 <level+888> "@\"\257!\211U", avail_in = 4044095571, total_in = 94047298260400, next_out = 0x558921b72a30 "", avail_out = 1869903169, total_out = 94046671553552, msg = 0x558920f57250 "@\3645\025\211U", state = 0x60c0300000018, zalloc = 0x0, zfree = 0x558922135820, opaque = 0x558920f57250, data_type = 230848512, adler = 140734703650688, reserved = 94046650419626}
        err = <optimized out>
[...]
```

Interestingly, in this run it says the faulty address is null, while I'm pretty sure previous runs would say address `0x58`.

[uzdoom-crash.log](https://github.com/user-attachments/files/23686667/uzdoom-crash.log)

### Configuration

```ini

```

[...]

Issue #362, opened 2025-11-05
UZDoom crashes on saving


inkoalawetrust:
### UZDoom Version

b04772e (Actually this https://github.com/inkoalawetrust/gzdoom/tree/82af1b3291e807490cd54a2aa1d55b4defc1d54e)

### Game

Doom

### Operating System

Windows 10

### Duplicate Confirmation

- [x] I have searched and not found a similar issue.

### Bug Description

[It's basically this bug I had already reported to GZDoom last year.](https://github.com/ZDoom/gzdoom/issues/2407) However, it does happen with RelWithDebInfo builds too after all. Anyway, bug has kept happening since then when I play PB with Obsidian maps (Surely happens elsewhere too since GZDoom has [a similar PB saving bug report](https://github.com/ZDoom/gzdoom/issues/2487)).

All that aside, pretty straightforward, I play Project Brutality with an Obsidian mapset, and eventually it starts often crashing when I try saving. Unlike previous times, this now begun happening just 1-2 maps in, not ~5 hours in. So hopefully it means that this playthrough setup can also reproduce the bug after less than 30 minutes of gameplay from scratch, even without the savefile or exact UZDoom version etc. So that y'all don't have to basically play the whole game to get it.

[Here's my exact version of PB.](https://github.com/inkoalawetrust/Project_Brutality/tree/88f00d44f54a88c2c0a2adc2b848c16b41f8e9d4)

My entire playthrough setup (ZScript music randomizer, extra Doom 1 music, the mapset, the Obsidian texture packs, the save file): https://drive.google.com/file/d/1pPlICdFqf-pe8wLALomw9Vvdr0TcPw3A/view?usp=sharing

Edit: Also, might as well put this here instead of just on Discord, but the crash seemed to happen because there was some kind of garbage memory entity in mDObjects during FSerializer::WriteObjects().
<img width="1004" height="171" alt="Image" src="https://github.com/user-attachments/assets/7078efb8-c373-4ec3-8435-814c45477f59" />

### Replication

1. Use either my zscript-polyobjects-new branch, or if save files are not THAT rigid, it should also work on mainline UZDoom.
2. Load up the mapset and all the assets (With ZDL for ease of use probably, like I do)
3. Load the save file.
4. Play for as long as possible and save, I usually save pretty often, like after every minor event.
5. Eventually UZDoom should crash trying to save.

### Configuration

```ini
[...]

PR #874, opened 2026-01-05
[discuss] Prediction GC hack

TehRealSalt:
## Description

I spent a lot of time looking at #560 + #362 + my own mod getting this crash, w/ ASAN + debugger + print spam, and I'm fairly sure that these are all the same problem: a save occurring soon after a GC sweep happening in the middle of a player prediction. Unmodded doesn't do as much garbage collection as some crazy ZScript mods can, so I wasn't able to fully confirm that it happens there, but it's theoretically possible all of the time.

The proper fix for this problem would be #585, but that's planned for 5.1. I came up with this temporary band-aid instead; don't run GC during predictions, make sure GC is checked before + after every prediction to accommodate instead. This appears to be the best solution that doesn't introduce any full sweeps or break GC behavior.

From testing, this change has made my game significantly more stable; I haven't experienced a new crash since integrating this change in my fork, when I used to be able to reliably crash multiple times a day.  But it still feels pretty gross, and I'm not *completely* convinced this alone can get rid of the crash entirely.

I opened this to feel out what other people think about doing this as a temporary measure for 5.0. Feel free to close if there's no consensus. If accepted, I'd suggest not closing #560 or #362 until #585 has been done instead.

## Checklist

### DIFF
--- a/src/common/objects/dobjgc.cpp
+++ b/src/common/objects/dobjgc.cpp
@@ -154,6 +154,17 @@ static cycle_t GCTime;			// Track time spent in GC
 
 void CheckGC()
 {
+	if (bPredictionGuard)
+	{
+		// HACK: GC running during prediction is terrifying,
+		// because it runs off of copious amounts of memcpy.
+
+		// This (and the related hack in P_UnPredictPlayer)
+		// will not be needed whenever the prediction
+		// is updated to use serialization instead.
+		return;
+	}
+
 	AllocHistory.AddAlloc(RunningAllocBytes);
 	RunningAllocBytes = 0;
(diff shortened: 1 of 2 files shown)


RicardoLuis0 (approved):

inkoalawetrust:
I find it kinda funny how much of a problem that random bug I discovered 2 years ago has been. For it to *still* be an issue even after multiple attempts to squash it. And if anything it seems to have just gotten worse on my end over time, last time I was playtesting it, it had begun happening after just 2 maps.

the-phinet (approved):
I've been testing this for a while and everything seems well. Is this good to merge?

PR #879, opened 2026-01-07
Feature: use fonts generated from TTF Files

River-Salmon:
## Description

Features:
Adds capability of rendering bitmap fonts rendered from ttf files.
Adds user options for font overrides, remappings, and preferences.
Font API for semantic font choice (ex: Font.GetSmallTextFont() rather than using NewSmallFont directly)

Primary use cases:
Accessibility: fonts like Atkinson Hyperlegible are substantially easier to read for some users.
Customization / modding: making it easier to control the presentation of text in UZDoom.

What this does NOT do:
-Pixel-perfect vector rendering of fonts: The fonts generate a bitmap atlas at startup. 

Third party dependencies:
-FreeType
-Harfbuzz
-Trex
-simdutf
-stb_image_write
-Included font files

## Checklist


src/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 cmake_minimum_required( VERSION 3.16.0 )
 
 include(precompiled_headers)
+include(FetchContent)
```
kcat: I don't like the idea of dynamically pulling from an external website during the build. Dependencies are usually handled by including a copy in the codebase and/or relying on an already-installed package.

the-phinet: Agreed.

River-Salmon: Addressed in recent changes

River-Salmon:
I've moved this out of draft. We're at the point where I think we can start taking deeper feedback about the PR

wadsrc/static/fonts/dynamic/New Text Document.txt
Niterux, line 1: What does this file do?

River-Salmon:
Nothing. I'll remove it

the-phinet:
> > What does this file do?
>
> Nothing. I'll remove it

No, keep it. It [might be important](http://www.catb.org/jargon/html/magic-story.html) :joy:

River-Salmon:
For those following this; a fairly recent build off this PR is available on the Experimental Releases. Please give it a shake and post feedback!

giwake:
gave this a go, so far it's really promising!

some small bugs/thoughts:

the console cursor seems to have a small 1px(?) space after hitting the spacebar, which then expands to a proper sized space when another character is added afterwards. this doesn't happen in the chat text input. it's particularly visible after "brown" and "dog" in this clip. this is a little disorienting and made me think i didn't hit the spacebar.

uzdoomfontjumping

~~additionally it appears that having something like cleartype or font smoothing turned on or off in windows applies it in uzdoom. you can see how the text in the gif above looks weird because i have cleartype and font smoothing turned off (personal preference)~~ struckthrough - cleartype/font smoothing doesn't seem to actually cause this bug? strange...
the console cursor bug happens regardless of if cleartype/font smoothing is enabled.

i don't know the intricacies of TTF rendering but i think it would be neat to have the ability to enable/disable font smoothing within uzdoom for modders/devs, what with doom being retro and all. some fonts (my goto being any of [microsoft's old "core fonts for web"](https://en.wikipedia.org/wiki/Core_fonts_for_the_Web)) look particularly nice with it disabled.

i'm pretty excited for this ~~especially because i could no longer have to faff around with trying to import comic sans in a roundabout way.~~

the-phinet:
Minor issue with non-ttf rendering: Listmenu is colourless sometimes and the cursor is off-centre

[Screencast_20260416_215344.webm](https://github.com/user-attachments/assets/09d85b80-2039-4270-a620-4e8a40a29b2c)

RicardoLuis0 (approved):

PR #910, opened 2026-01-13
Fix appstream linter errors and edit migration instructions

fpiesche:
## Description

- hard wrap long lines for readability
- lowercase developer id
- fix branch name for screenshots
- edit UZDoom release notes with more detailed and Linux-specific info on migrating from GZDoom

## Checklist

### DIFF
--- a/src/posix/freedesktop/org.zdoom.UZDoom.metainfo.xml
+++ b/src/posix/freedesktop/org.zdoom.UZDoom.metainfo.xml
@@ -6,7 +6,7 @@
 	<project_license>GPL-3.0</project_license>
 
 	<name>UZDoom</name>
-	<developer id="org.zdoom.UZDoom">
+	<developer id="org.zdoom.uzdoom">
 		<name>UZDoom team</name>
 		<url>https://github.com/UZDoom/UZDoom/graphs/contributors</url>
 	</developer>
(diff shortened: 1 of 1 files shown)


fpiesche:
Just running a local test build to check the formatting on the release notes looks reasonable before I mark this as ready...

src/posix/freedesktop/org.zdoom.UZDoom.metainfo.xml
```diff
@@ -92,104 +103,174 @@
 	</categories>
 
 	<releases>
-		<!-- Keep 5 full descriptions, and 2 years of releases with just urls -->
```
the-phinet, line 95: Can you keep this note here?

src/posix/freedesktop/org.zdoom.UZDoom.metainfo.xml
```diff
@@ -95,101 +106,172 @@
...
+						in <code>~/.var/app/org.zdoom.GZDoom/.config/zdoom/cache</code></li>
+					<li>For other installations of GZDoom, this data can be found in
+						<code>~/.config/zdoom/cache/</code></li>
 				</ul>
-				<p>In practice, you want to:</p>
-				<ol>
-					<li>Create folder <code>~/.local/share/games/uzdoom</code>, and move everything from <code>~/.config/gzdoom</code> except <code>gzdoom.ini</code> and <code>autoexec.cfg</code> into it.</li>
-					<li>Rename <code>gzdoom.ini</code> to <code>uzdoom.ini</code>.</li>
-					<li>Rename <code>~/.config/gzdoom</code> to <code>~/.config/uzdoom</code>.</li>
-				</ol>
```
the-phinet, line 144: So this file isn't just used for Linux. Please keep information about other platforms, too

the-phinet: Actually I'm noticing other edits (like removing the headers). We'd like to keep the release notes the same as the release notes on the forum. 

On a personal note, I don't love hard-wrapping xml files (xml spec doesn't collapse whitespace by default, so hard-wrapping lines actually affects the data). I much prefer my editor to soft-wrap them. Unless it's a thing that flathub will complain about, I'd rather keep the long lines.

fpiesche: The headers cannot be in the release notes; the FreeDesktop spec [only allows `p` and `ul`/`ol` as base level tags and only `code` and `em` inside the text portions](https://freedesktop.org/software/appstream/docs/chap-Validation.html#asv-description-markup-invalid). Note that while the spec here only refers to the `description` element of the metainfo, this does also apply to the `description` elements for `release`s.

fpiesche: Also, what other platforms is this file used for? Neither Windows nor macOS support FreeDesktop standards.

fpiesche: The other option would be to just detach this release info completely from what's on the forum, and just have a separate short summary of changes ("First release of UZDoom with some security fixes and changes to default options" or something along those lines, and leaving the more elaborate content for the link to the full release notes. Would make maintaining the `releases.xml` a bit less hassle, too.

the-phinet: > The headers cannot be in the release notes; Flatpak only allows p and ul/ol as base level tags and only code and em inside the text portions

Okay, sounds good. ~~Out of curiosity, where is that defined? I couldn't find it I the freedesktop spec last time I looked at it~~ Just saw the edit. Thanks!

> Also, what other platforms is this file used for? Neither Windows nor macOS support FreeDesktop standards

All platforms. Specifically the release notes section is used in the launcher. I guess the file could be generated from the cross-platform notes that could be kept in `docs` or something. This was just easier and ensured things stay up to date 

> The other option would be to just detach this release info completely from what's on the forum, and just have a separate short summary of changes ("First release of UZDoom with some security fixes and changes to default options" or something along those lines, and leaving the more elaborate content for the link to the full release notes.

I'd prefer if we didn't do that. I don't want to point the user to an external source to see what updates came with

the-phinet: > The headers cannot be in the release notes; Flatpak only allows p and ul/ol as base level tags and only code and em inside the text portions

Okay, sounds good

> Also, what other platforms is this file used for? Neither Windows nor macOS support FreeDesktop standards

All platforms. Specifically the release notes section is used in the launcher. I guess the file could be generated from the cross-platform notes that could be kept in `docs` or something. This was just easier and ensured things stay up to date

fpiesche: Ahhh, I didn't know about the launcher making use of it! I'll go and revert most of my rewrites then; the main issue with the migration instructions is that the Linux ones are incomplete and wouldn't work for someone using either the ZDoom or UZDoom Flatpak...

the-phinet: > Ahhh, I didn't know about the launcher making use of it! I'll go and revert most of my rewrites then; the main issue with the migration instructions is that the Linux ones are incomplete and wouldn't work for someone using either the ZDoom or UZDoom Flatpak...

Instead of reverting, can you add the flatpak info? I assume you're already going to do that, but just in case

the-phinet: Man, GH sucks. Editing messages is now duplicating them?

the-phinet:
I'm going to go ahead and merge this to keep the CI check happy in other PRs. I'll open another pr (unless you beat me to it) adding the cross-platform stuff back when I have a moment

the-phinet (approved):

PR #912, opened 2026-01-13
Expand autoaim range

### LINKED ISSUE #906: Increase vertical auto aim range
### Game

All

### Problem Description

Ever since the beginning, the auto aim range has been less than the projectile range. This is quite the problem as plenty of maps will have hitscan enemies above or below the horizon, able to hit the player but not vice versa, not without the player getting dangerously close. Playing without mouselook is part of the DOOM charm, and upgrading it seems natural for the next build, especially since there aren't any mods or scripts available to do so. 

### Feature Description

The smart autoaim feature should have its range extended to be the same as the limits of projectile range, thus eliminating an unfair NPC enemy advantage when firing hitscans at range.

### Additional Media

_No response_

the-phinet:
## Description

Allow for larger auto-aim range during single player games

Closes #906

## Checklist

### DIFF
--- a/src/playsim/d_player.h
+++ b/src/playsim/d_player.h
@@ -17,33 +17,28 @@
 **
 **---------------------------------------------------------------------------
 **
+** Finally, for odd reasons, the player input is buffered within
+** the player data struct, as commands per game tick.
+**
+** The player data structure depends on a number of other structs:
+** items (internal inventory), animation states (closely tied to
+** the sprites used to represent them, unfortunately).
+**
+** In addition, the player is just a special case of the generic
+** moving object/actor.
 */
 
 #ifndef __D_PLAYER_H__
 #define __D_PLAYER_H__
 
-// Finally, for odd reasons, the player input
-// is buffered within the player data struct,
-// as commands per game tick.
+#include "a_weapons.h"
+#include "actor.h"
+#include "b_bot.h"
+#include "basics.h"
+#include "d_netinf.h"
 #include "d_protocol.h"
 #include "doomstat.h"
-
-#include "a_weapons.h"
-
-#include "d_netinf.h"
-
-// The player data structure depends on a number
-// of other structs: items (internal inventory),
-// animation states (closely tied to the sprites
-// used to represent them, unfortunately).
 #include "p_pspr.h"
-
-// In addition, the player is just a special
-// case of the generic moving object/actor.
-#include "actor.h"
-
-//Added by MC:
-#include "b_bot.h"
 
 class player_t;
 
--- a/src/playsim/d_player.h
+++ b/src/playsim/d_player.h
@@ -37,6 +37,7 @@
 #include "basics.h"
 #include "d_netinf.h"
 #include "d_protocol.h"
+#include "doomdef.h"
 #include "doomstat.h"
 #include "p_pspr.h"
 
@@ -192,9 +193,10 @@ struct userinfo_t : TMap<FName,FBaseCVar *>
 		}
 
 		float aim = *static_cast<FFloatCVar *>(*CheckKey(NAME_Autoaim));
-		if (aim > 35 || aim < 0)
+		float bound = (dmflags & DF_NO_FREELOOK)? 35: 70;
+		if (aim > bound || aim < 0)
 		{
-			return 35.;
+			return bound;
 		}
 		else
 		{
--- a/src/playsim/p_map.cpp
+++ b/src/playsim/p_map.cpp
@@ -4546,7 +4546,8 @@ DAngle P_AimLineAttack(AActor *t1, DAngle angle, double distance, FTranslatedLin
 				// vrange of 0 degrees, because then toppitch and bottompitch will
 				// be equal, and PTR_AimTraverse will never find anything to shoot at
 				// if it crosses a line.
-				vrange = DAngle::fromDeg(clamp(t1->player->userinfo.GetAimDist(), 0.5, 35.));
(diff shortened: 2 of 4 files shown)


RicardoLuis0:
i can't think of any off the top of my head, but would there be any vanilla maps that explicitly rely on the autoaim max angle? if so, it'd be good to at least add a compat flag to clamp it to the vanilla 35 degrees

the-phinet:
> maps that explicitly rely on the autoaim max angle

I can't think of any either.

I think it's probably fine without a compat flag?

Ralgor:
I think there's at least one Plutonia map that relies on it. There's a room with a bunch of chaingunners above the player and just out of auto aim range. It's intentional.

the-phinet:
> There's a room with a bunch of chaingunners above the player and just out of auto aim range

Okay, cool! That's really mean, actually 😂

In this case, is adding a compat flag needed? I assume the compat flag would be off by default, and the angle would default to the old value

Ralgor:
Since the default hasn't changed, and the maps don't actually break if it does, I don't think a compat flag is necessary. All that happens is the maps end up easier than expected. I expect that many harder map sets were designed with this limit in mind, but we'd have to ask some actual mappers to know for sure.

mc776:
> I think there's at least one Plutonia map that relies on it. There's a room with a bunch of chaingunners above the player and just out of auto aim range. It's intentional.

The final area in Freedoom Map11 does this too, you basically play it like a Mass Effect-style cover shooter until you're at the final pillar.

I don't think a compat flag is needed, for the same reasons Ralgor said.

HexaDoken:
I can't imagine any map would outright _break_ from this, and if a player is worried about making maps too easy, I imagine they can just set the autoaim limit to default?

What's the reason for making this SP only?

> I think there's at least one Plutonia map that relies on it. There's a room with a bunch of chaingunners above the player and just out of auto aim range. It's intentional.

Wouldn't that be completely unaffected by this though? This just increases the vertical angle limit, it doesn't touch the horizontal range at which autoaim triggers as far as I can tell? The chaingunners are too far away, not too high up.

the-phinet:
> What's the reason for making this SP only?

In my head, it's an unfair advantage. There would be no reason to not crank it to the max. (If the consensus is to remove this limitation, I'm happy to change it)

HexaDoken:
> In my head, it's an unfair advantage. There would be no reason to not crank it to the max. (If the consensus is to remove this limitation, I'm happy to change it)

Not really an unfair advantage if anyone could do it though? Besides in DM it's much better to manually aim with mouselook than trying to rely on fickle autoaim.

I _suppose_ it might be a factor if somebody is going to use UZDoom to play strict vanilla, no-freelook DM, but... is that worth worrying about? We don't exactly have a rich competitive scene, and there are frankly better ports for vanilla DM anyway.

I definitely see this limit being annoying for coop games, where it doesn't accomplish much and which accounts for much more of actual ZDoom multiplayer games.

the-phinet:
> I definitely see this limit being annoying for coop games, where it doesn't accomplish much and which accounts for much more of actual ZDoom multiplayer games.

That's a really good point. ~~Unless others disagree, I'll remove the limit~~ Done.

RicardoLuis0:
> > I definitely see this limit being annoying for coop games, where it doesn't accomplish much and which accounts for much more of actual ZDoom multiplayer games.
> 
> That's a really good point. Unless others disagree, I'll remove the limit

i don't mind it being both mp and sp yeah

ConcernedCoder:
> I think there's at least one Plutonia map that relies on it. There's a room with a bunch of chaingunners above the player and just out of auto aim range. It's intentional.

Funny, I was thinking of that TNT map with heavy weapon dudes below the line of sight.

TehRealSalt:
> I _suppose_ it might be a factor if somebody is going to use UZDoom to play strict vanilla, no-freelook DM, but... is that worth worrying about? We don't exactly have a rich competitive scene, and there are frankly better ports for vanilla DM anyway.

I still think it'd be nice to account for people who want the overall "feel" of these settings. Obviously UZDoom has extremely low overall accuracy, but a casual group of friends might not care and just want something that *generally* plays like OG Doom but with a ZScript mod.

Not to pull Marcus around ( :sweat_smile: ), but in my personal opinion it'd probably be good to only do the smaller cap for the existing `DF_NO_FREELOOK` dmflag. If no-mouselook is being explicitly enforced for every player by server gameplay settings, then I think that's a safer bet that they care about purity here.

the-phinet:
> Not to pull Marcus around

Lmao. I'd rather things be correct than worry about my sanity 

> it'd probably be good to only do the smaller cap for the existing DF_NO_FREELOOK dmflag

This makes a lot of sense. What are Jay's thoughts?

RicardoLuis0:
> > Not to pull Marcus around
> 
> Lmao. I'd rather things be correct than worry about my sanity
> 
> > it'd probably be good to only do the smaller cap for the existing DF_NO_FREELOOK dmflag
> 
> This makes a lot of sense. What are Jay's thoughts?

i don't see why not, but maybe i'd ask this to someone who actually plays dm to be fair xD

HexaDoken:
> but maybe i'd ask this to someone who actually plays dm to be fair xD

Sounds good to me yeah. (I thought of that idea but forgot to actually write it WHOOPS LMAO)

the-phinet:
It respects DF_NO_FREELOOK now :)

RicardoLuis0 (approved):

PR #937, opened 2026-01-20
Add gl_showboundingbox

RageSpark:
Ported and adapted r_showhitbox from VKDoom, also added Projectile Pass Height.

## Checklist


nashmuhandes:
As was discussed in Discord, this isn't suitable to add in its current form - the way VKDoom did it is a mess. There are plans to do this properly for 5.1, so I'd rather not the codebase be unnecessarily polluted with sub-par code.

Sorry!

PR #941, opened 2026-01-21
Stopped ticking client-sided Thinkers when paused

Boondorl:
For now this will be done to keep parity between the Actor ticking and model rendering (this relies on the game being unpaused to advance properly).

## Checklist

### DIFF
--- a/src/p_tick.cpp
+++ b/src/p_tick.cpp
@@ -61,10 +61,13 @@ void P_RunClientSideLogic()
 
 	if (gamestate == GS_LEVEL || gamestate == GS_TITLELEVEL)
 	{
-		for (unsigned int i = 0; i < MAXPLAYERS; ++i)
+		if (!WorldPaused(false))
 		{
-			if (playeringame[i] && players[i].inventorytics > 0)
-				--players[i].inventorytics;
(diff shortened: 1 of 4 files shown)


RicardoLuis0:
should cherry-pick this commit into here as well, or else the interpolations will be wrong by one tick https://github.com/RicardoLuis0/UZDoom/commit/5b375f3eb4db98e726aa7550671eb079cb6bf1f9

Boondorl:
Since this disables any ticking at all when paused I think the use of LocalTimer will need to be removed. If you're willing to clean that part up I'll cherrypick it into this PR

RicardoLuis0:
> Since this disables any ticking at all when paused I think the use of LocalTimer will need to be removed. If you're willing to clean that part up I'll cherrypick it into this PR

nah, the animating itself doesn't rely on ticking, so it'd still work, the actor just wouldn't be able to _change_ the animations from outside of ui scope

Boondorl:
Gonna be honest, don't know if I'm gonna have the energy to get around to cherrypicking that lol. I'd rather just send it over via a new PR since it's not entirely related to this one

TehRealSalt (approved):

PR #943, opened 2026-01-21
Extract macOS plist bundle version from version.h

stared:
The macOS app bundle (`UZDoom.app/Contents/Info.plist`) showed "Development Version" instead of the actual version.

I discovered while developing https://github.com/stared/rusted-doom-launcher, as I wanted to provide information of the Doom engine for debugging. GZDoom provides correct version.

```bash
$ defaults read "/Applications/GZDoom.app/Contents/Info.plist" CFBundleShortVersionString
g4.14.2
```

```bash
$ defaults read "/Applications/UZDoom.app/Contents/Info.plist" CFBundleShortVersionString
Development Version
```

At the same time, running it gives the correct version:

```bash
$ /Applications/UZDoom.app/Contents/MacOS/uzdoom
UZDoom 4.14.3 - 2025-11-28 23:17:36 +0000 - Cocoa version
Compiled on Nov 28 2025

MacBookPro18,1 running macOS Unknown version 15.7.3 (24G419) 64-bit ARM
UZDoom version 4.14.3
```

So, I expect that the version in `Info.plist` is the same.

Screenshot 2026-01-21 at 23 41 47
Screenshot 2026-01-21 at 23 42 04

See Apple documentation for [CFBundleShortVersionString](https://developer.apple.com/documentation/bundleresources/information-property-list/cfbundleshortversionstring).

One think I do not understand though is that GZDoom also has "Development Version" in https://github.com/ZDoom/gzdoom/blob/master/src/posix/osx/zdoom-info.plist. It seems that they have some private release pipeline not shown here. Yet, it is good have a proper way to do it.

As an important disclaimer - I am not an expert in macOS build systems. So I would be grateful if someone with better knowledge checked that.

## Checklist

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -261,3 +261,4 @@ RichardDS90
 Frank Kloster
 Santtu "MFG38" Pesonen
 heiko123abc
+Piotr "stared" Migdał
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,24 +1,5 @@
 cmake_minimum_required( VERSION 3.16.0 )
 
-function(extract_version_define output_var define_name capture_pattern)
-    set(strict_regex "^#define ${define_name} ${capture_pattern}$")
-    file(STRINGS "${CMAKE_CURRENT_SOURCE_DIR}/version.h" match_lines REGEX "${strict_regex}")
-    list(LENGTH match_lines match_count)
-    if(NOT match_count EQUAL 1)
-        message(FATAL_ERROR "Expected exactly one ${define_name}, found ${match_count}")
-    endif()
-    list(GET match_lines 0 match_line)
-    string(REGEX REPLACE "${strict_regex}" "\\1" extracted_value "${match_line}")
-    set(${output_var} "${extracted_value}" PARENT_SCOPE)
-endfunction()
-
-extract_version_define(UZDOOM_VERSION_STRING   "VERSIONSTR"   "\"([^\"]+)\"")
-extract_version_define(UZDOOM_VERSION_MAJOR    "VER_MAJOR"    "([0-9]+)")
-extract_version_define(UZDOOM_VERSION_MINOR    "VER_MINOR"    "([0-9]+)")
-extract_version_define(UZDOOM_VERSION_REVISION "VER_REVISION" "([0-9]+)")
-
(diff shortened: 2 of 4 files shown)


the-phinet:
Hey, thanks for this! I have not had time to look deeply at this yet, but it looks good to me.

I believe it would be more correct to grab the version string from `gitinfo.h`. It's generated during build time with [`UpdateRevision.cmake`](https://github.com/UZDoom/UZDoom/blob/trunk/tools/updaterevision/UpdateRevision.cmake)

stared:
Thank you @the-phinet for looking at this PR.

I wanted to directly related it to the single source of truth.

Also, as far as I understand Apple seems to require version string to be just`[Major].[Minor].[Patch]`, not hashes or so, vide https://developer.apple.com/documentation/bundleresources/information-property-list/cfbundleshortversionstring

> This key is a user-visible string for the version of the bundle. The required format is three period-separated integers, such as 10.14.1. The string can only contain numeric characters (0-9) and periods.

TehRealSalt:
The main concern I have with the current approach is that you're reusing the ZScript/networking version for it. I'm not sure what other org members' opinion is on reusing those constants for release version, because they could potentially differ and are not ever presented to the user currently. (This probably won't happen for this repo but is much more likely for forks that want to make releases that maintain compatibility.) The version that is printed by the game to the user in every other scenario is the one from git.

Even if Apple needs to be in that format, imo it would be more correct to derive from the git tag. If that isn't reasonable, then I think it'd be better to add a new version string define instead of reusing technical constants. (I've noticed Windows is currently doing whatever it wants for its version too when it shouldn't, so it'd be a good way to merge those two together as well)

stared:
Thank you for your remarks, @TehRealSalt.

If git tags are the source of truth, then yes, Apple builds should take from it.
I modified code. Let me know if it is good to go, or any other changes are needed.

TehRealSalt:
Thank you! (Sorry for the prolonged delay btw, we're light on Mac maintainers.) The changes look good to me, but I'll give @the-phinet a chance to check since they have an actual Mac to test on

stared:
@TehRealSalt No rush!

I appreciate your time.

stared:
@the-phinet By any chance, do you have time to check it?

(Since it is not much content it, squash-merge might work the best.)

the-phinet:
I have not had a chance to yet. My office is in disarray. I will take a look by the weekend. Sorry for the delay.

the-phinet (approved):
Looks good to me. Sorry about how long it took me to take an actual look.

Due to another issue (not yours) it doesn't work through ci properly. I will need to fix this soon. It's messing with the build numbers on the prerelease builds.

PR #947, opened 2026-01-24
Theme Update

### LINKED ISSUE #381: Icon and Progress bar colours to be better harmonised
### Game

All

### Problem Description

I mentioned this in the ZDoom.org UZDoom announcement thread, but I may as well put it here too as a minor suggestion for consideration.

The background for the new icon is a cyan-teal-like colour ( #237887 or 35, 120, 135).  The progress bar in UZDoom itself is a much brighter cyan (#2EEDFC or 46, 237, 252 according to a screenshot I took) colour.

I just thought it might be nice if the icon colour and the progress bar colour were closer to each other.  It kind of looks like there was an attempt to match colours but it didn't quite work out (like when someone wears two versions of the same colour clothing - but they don't match, so it looks like there was a failed attempt to do so).

My preference would be to subdue the progress bar to something closer to (or maybe exactly the same as) the icon background.  The bar is very bright against the black background of the splash screen and could do with being a bit more subdued anyway IMO.

It's just a gentle suggestion though.  Take it or leave it. :)

### Feature Description

Present splash screen and progress bar colour (with icon inserted)
Image

Suggestion (progress bar the same colour as the icon)
Image

### Additional Media

_No response_

### LINKED ISSUE #901: Update launcher theme
- Tweak lightmode to be less yellow
- Add darkmode

https://phinet.ca/projects/zwidget/

the-phinet:
Updates theme, makes theme mod configurable

Closes #901 
Closed #381 

Screenshot_20260124_002158
Screenshot_20260124_002237

## Checklist


TehRealSalt (commented):
Looks good to me! Just a couple minor ?s:

- Maybe `theme.txt` should be in the widgets sub-folder along with the banner images for organization reasons (not sure of what this folder's fate will be with that other launcher, but I imagine if these assets get reused then they'd stay in that sub-folder)

- I wasn't able to get theme to change from a mod, but was able to replace the file in `uzdoom.pk3`. I think WidgetResources simply doesn't include the mod file list. If it's too much effort to change, still definitely keep the theme.txt code in because it makes it easy for non-coders to push theme changes, but it probably shouldn't be advertised as mod-configurable. (My test file, I tried just using `./uzdoom -file theme.zip`: [theme.zip](https://github.com/user-attachments/files/24838967/theme.zip))

the-phinet:
> Maybe theme.txt should be in the widgets sub-folder along with the banner images for organization reasons (not sure of what this folder's fate will be with that other launcher, but I imagine if these assets get reused then they'd stay in that sub-folder)

That's a good point. I think it would make sense to rename the folder to `ui` or something like that, since if the theme data gets put in there, it would not be just for the launcher.

> I wasn't able to get theme to change from a mod, but was able to replace the file in uzdoom.pk3. I think WidgetResources simply doesn't include the mod file list. If it's too much effort to change, still definitely keep the theme.txt code in because it makes it easy for non-coders to push theme changes, but it probably shouldn't be advertised as mod-configurable.

Okay, good to know it's not working. I'm not sure how to get the filesystem to layer files. This is something that needs to be fixed. If someone can offer some guidance, I would be very grateful.

This is the current implementation:
```cpp
auto file = "theme.txt";
auto buffer = LoadWidgetData(file);
FString str;
for (auto c: buffer)
	str.AppendCharacter(c);
FScanner sc;
sc.OpenString(file, str);

while (sc.GetString())
{
	switch (sc.MatchString(ThemeCommandStrings))
	{
	// ...
	}
}
```

I'm 90% sure I'm opening the file in the worst possible way, and 80% sure I'm not closing it.

TehRealSalt:
Your current implementation looks alright to me personally, that's kind of just how the rest of the code uses LoadWidgetData. (The only thing is maybe use `FScanner::OpenMem` instead of `FScanner::OpenString`, to skip that character appending process. See also how menudef is loaded to get the supported languages list)

I took a cursory glance, the `WidgetResources` variable is a single `FResourceFile`, so it definitely doesn't have access to any more than the single file. `InitWidgetResources` is called extremely early into `D_DoomMain_Internal`, before `fileSystem.InitMultipleFiles` has been done, so you can't rely on the usual file system.

You could potentially change WidgetResources into its own `FileSys::FileSystem` instead of a single FResourceFile. You'd then have to load in all of the files, but LoadWidgetData could then just use some standard filesystem functions. I'm concerned doing that might add some delay to showing the launcher if you have lots of large mods on command line, but it's hard to determine the impact without doing the work first...

Depending on what your target audience is (mod creators looking to make the launcher themed around their mod, or users that just want to make their copy look unique), you might also just want to consider a different approach. If you just care about users, it'd be enough to let LoadWidgetData load files from a theme pk3 from somewhere on the OS first, and fallback to uzdoom.pk3 for things that weren't found. That would prevent slowing down startup too much, and would also let the theme be customized without needing to run from command line every time. (But also, platform-specific code for determining the best folders to expect users to put this stuff in is it's own headache ^^;)

the-phinet:
> The only thing is maybe use FScanner::OpenMem instead of FScanner::OpenString

Thanks, I totally missed that.

> Depending on what your target audience is (mod creators looking to make the launcher themed around their mod, or users that just want to make their copy look unique), you might also just want to consider a different approach.

I think I'm mainly concerned about people making standalone mods (though the other 2 use cases you listed would be nice to cover, too). For Snap, do you customize the uzdoom.pk3? If so, the current implementation is probably fine. If not, I agree things should be re-thought a bit.

> I think it would make sense to rename the folder to `ui` or something like that, since if the theme data gets put in there, it would not be just for the launcher.

Actually, I think this would be a good change, since it would give #879 a place to put fonts along with the launcher fonts

TehRealSalt:
> I think I'm mainly concerned about people making standalone mods (though the other 2 use cases you listed would be nice to cover, too). For Snap, do you customize the uzdoom.pk3? If so, the current implementation is probably fine. If not, I agree things should be re-thought a bit.

I wouldn't worry about my case at all! Snap does have its own fork, and does already replaces the lighting model by simply overwriting the shaders in uzdoom.pk3. If I got a feature to do that properly, I'd probably make it use that because I want to minimize my fork's changes. But because I'm doing my own distribution, I have uzdoom.pk3 prepped to be edited anyways, so I may as well just edit it. I will always appreciate having canonized ways to do standalone game changes, but it should not be a focus point for the engine itself IMO.

How this is implemented now is already streamlines the process of trying out different colors for engine devs, so as I said earlier I still think this is a great improvement as-is, if you don't want to worry about targeting broader gameplay mods or users

the-phinet:
Alrighty, I think I have mods working in an acceptable fashion. Load times are not impacted by much. Loading WolfenDoomBOA (a 700mb wad) 100 times adds about 10 seconds to the initial load, so about 0.1s per big mod. Loading just a theme patch adds a negligible load time.

If you're good with the implementation, I'll clean up the last couple commits and I think this is good to go

TehRealSalt:
If I'm to nitpick, could be extended to the banner images too, but otherwise LGTM!

the-phinet:
> If I'm to nitpick, could be extended to the banner images too, but otherwise LGTM!

Do we care if the banner images are kept in memory for the lifetime of the game? If we don't care, adding this would be a trivial change.

TehRealSalt:
> Do we care if the banner images are kept in memory for the lifetime of the game? If we don't care, adding this would be a trivial change.

Doesn't sound too bad to me personally

the-phinet:
Implemented image overwriting and sped up data loading. And fixed another longstanding (minor) zwidget bug lol.

Boondorl (approved):

PR #952, opened 2026-01-25
code in "_dbg_common.h" removed duplicate macro definitions UNDNAME_NO_…

fcarvajalbrown:
\UZDoom\bin\windows\vpx\src\_dbg_common.h presents a file that has duplicate macro definitions affecting redundancy and clean-up. Macros duplicated that were removed and kept:

- [x] UNDNAME_NO_ARGUMENTS
- [x] UNDNAME_NO_SPECIAL_SYMS

## Checklist


RicardoLuis0:
vpx is a vendored library (on windows only, on macos/*nix it's pulled from the system), commits that modify it will not be accepted

fcarvajalbrown:
ok good to know thanks!

PR #904, opened 2026-01-11
New Launcher Prototype

heiko123abc:
## Description
 After some work,  i have integerated the Launcher into the current UZDoom executeable. I have tested it on Arch Linux and Windows 11 without major issues. But i want to keep it as a draft for now as there are still some unfinished stuff and to test if the workflow suits everyone. It also fully removes ZWidget.
 
Especially when it comes to the backend, it is currently crudely hooked into the platform files to get it working. The language switching also does not work.  If there are any other features or issues please post them below. The launcher was made using wxWidgets and thus uses windows native ui, gtk for linux and cocoa for Macos.

Screenshots:

grafik

grafik

grafik

grafik

There is also a new crash window that also shows on linux: 

Screenshot From 2026-01-11 16-22-24

Screenshot From 2026-01-11 16-23-44

The way it works right now is that it that it simply spawns a process from the launcher to track playtime and cmd launch paramters.
When the game process ends, the launcher collects the playtime.

The current executeables can be found here: https://github.com/heiko123abc/UZDoomNewLauncher/actions/runs/21781417388
**THE MACOS BUILD IS UNTESTED SINCE I DONT HAVE ONE**
## Checklist
- [x] I have reviewed [CONTRIBUTING.md](../blob/trunk/CONTRIBUTING.md) and agree to its terms.

## TODOs
- [ ] Dump startup information into the linux/macOS crash reporter
- [ ] Rework Platform code and remove too platform specific code
- [ ] Fix language switching

Issue #260, opened 2025-10-27
Platform Parity


the-phinet:
GZDoom was primarily developed on Windows, but it targets multiple operating systems. Some things are implemented only on one platform, but not on another. UZDoom would like to reduce the number of differences between platforms

This issue tracks known differences between platforms. If you use UZDoom on more than one operating system and you know of differences between them, please mention them below.

|Feature|🪟|🐧|🍎|🍁|ℹ️|🗒️|
|-------|--|--|--|--|--|--|
|Launcher|✅|✅|✅|✅|❓||
|Netgame Lobby|✅|✅|❌|❌|❓||
|Abort Screen|✅|❌|❌|❓|❓|#900|
|Crash Screen|❓|❓|❓|❓|❓||
|Gamepad support|✅|✅|❌|❌|❓||
|Modern gamepad support|❌|✅|❌|❌|❓||
|Haptics|✅|✅|❌|❌|❓||
|Fullscreen shortcut|✅|✅|❌|❌|❓||
|Portable Config|✅|❌|❓|❓|❓|#802|

<details><summary>Notes</summary>

- 🪟: Windows
- 🐧: Linux
- 🍎: MacOS
- 🍁: Haiku (not officially supported)
- ℹ️: Other unofficial platforms (Solaris, BSD, etc)
- 🗒️: Notes / links
- ✅: Intended behaviour
- ❌: Unintended behaviour
- ❓: Unknown

</details>

RexenSquared:
If by crash screen you mean the one that lets one report a crash then Windows does have it but since most things trigger a vm abort instead, it might be hard to trigger it unless there's a way to force it, as I don't know if reading from address zero still triggers a crash.

RicardoLuis0:
i think SDL runs on all that? so hopefully with the SDL3 port most of the code in the backends can be merged, leaving only a small amount of highly-platform-specific code

Enjay001:
> If by crash screen you mean the one that lets one report a crash then Windows does have it but since most things trigger a vm abort instead, it might be hard to trigger it unless there's a way to force it, as I don't know if reading from address zero still triggers a crash.

There is the CrashOut console command that does exactly what the name says.  It was added a very long time ago (in ZDoom by Marisa Heit IIRC) and it produces this in Windows 10:

<img width="625" height="533" alt="Image" src="https://github.com/user-attachments/assets/f9ef1187-ff39-4242-a65b-680beaec423d" />

[...]

Issue #449, opened 2025-11-11
Headless mode


andOlga:
### Game

All

### Problem Description

Port forwarding is becoming increasingly difficult in the modern world due to more and more ISPs employing CGNAT. It is still possible to bypass this with some fake LAN software, or by using UDP forwarding services. However, these services are not abundant, and the configuration is painful. At this rate, whatever imporvements are made for multiplayer are likely to be nearly-worthless for a lot of users due to overhead introduced by those methods and/or impossibility to use them.

### Feature Description

Given that one of the primary focuses in UZDoom right now seems to be multiplayer, the increase in maximum player count, and the fact that the engine is switching to the packet-server model as its default, it would be great if a flag was added to allow UZDoom to start without any graphical output. This way you could throw a copy of UZDoom together with your WADs onto a cheap, temporary VPS and fire up a multiplayer session, without having to worry about complicated setups.

Note that this is not a request for a full *dedicated server* mode. While that would be amazing to have in the long run, I think just "don't try to connect to X/Wayland" would be a fantastic start.

I realize that it's likely to result in a "ghost" player in the game who doesn't do anything, and that the multiplayer sessions will still be temporary, and that it would be needed to restart UZDoom between them, and so on. Honestly, I see those as non-issues compared to having to invent magic tricks to bypass the near-unavoidable blockages from modern ISPs.

RicardoLuis0:
this is explictly planned yeah, but it's something for 5.1/5.2/possibly further because it does still need even more extensive changes to networking

andOlga:
> but it's something for 5.1/5.2/possibly further

Of course, I fully understand and I do not expect this to be some oneliner change. I wouldn't even know where to begin tackling this myself. I'm really glad to hear that this *is* happening, though.

arrowgent:
dedicated-server --switch in the command line option would be a great feature when its available
much like chocolate-doom has an interactive `-server` multiplayer and a dedicated-server `-dedicated` within the same binary


ISPs obviously arent something players can control.  buying an ISP approved modem that doesnt allow managing the network is a choice, although a choice some people cant afford to avoid.
personally, i bought my modem 5 years ago and paid for itself in 11 months (versus the subscription lasting indefinitely) but i also sacrifice unlimited bandwidth for this...as a punishment...because of course.

andOlga:
>  buying an ISP approved modem that doesnt allow managing the network is a choice

[...]

PR #957, opened 2026-01-26
Show netgame widget on all systems

### LINKED ISSUE #255: Lobby window is missing from Linux builds (?)
[More info to be provided later, just putting this here so that we don't forget]

TehRealSalt:
Resolves #255. Is related to #260

image

Cleans up the code to remove the platform-specific indirection to calling the netgame widget, and always call the ZWidget functions on all systems. Additionally, due to a myriad of issues (crashes on Linux + a giant black box in the background that can't be interacted with on Windows), the main game window is now initialized after the netgame lobby. This makes it possible to host a netgame on Linux. I've tested hosting a game from Linux and having someone on Windows join me, and that worked well.

#904 will resolve this in a proper way, but we will likely be taking that in for 5.1. I tried doing this as a simple fix we can potentially integrate for 5.0. Plus, the window change here should apply to that branch as well.

This should also apply to Mac. IIRC there's something up with Mac's ZWidget support, bur I'm unable to test this. Regardless, the old console-based fallback window that I deleted was also non-functional. If ZWidget support is fixed on Mac, then this should also start appearing there too. I propose that making a handful of tweaks to ZWidget to try and get Mac to work sooner is a lot more likely than us making a platform-specific solution, especially for 5.0. (And again, 904 is the actual long-term solution here.)

This comes with the caveat that the loading screen is now dummied out, practically speaking. There's some potential alternatives we can do to restore it, but all of them will be much easier if we get rid of ZWidget first. (Once more: 904.)

## Checklist


the-phinet:
Nice! It's pretty cool to have this working on Linux now.

- The loading screen being neutered is a bit of an issue. I fear leaving a fixme will lead to it being forgotten. Can the window allocation happen earlier conditionally?
- Since the textmode lobby has been removed, does that mean multiplayer is broken on mac? As soon as I unpack my mac, I'll start working on getting this screen wired up, too.
- Longterm, a headless server mode will be worked on (#449). Would it not make sense to keep the textmode lobby around for that? Or would building an actual tui from the ground up make more sense?

TehRealSalt:
> The loading screen being neutered is a bit of an issue. I fear leaving a fixme will lead to it being forgotten. Can the window allocation happen earlier conditionally?

I think having `if singleplayer then don't` switches are just as much as a fix-later-type issue personally. Messing with startup like this might introduce some subtle bugs because of how sensitive this code is, and multiplayer-only changes get significantly less testing than changes that affect singleplayer, so I opted to prefer doing what will get more reports.

I'm not confident in my stance there though, so I'll do it anyways if you still feel strongly about it.

> Since the textmode lobby has been removed, does that mean multiplayer is broken on mac? As soon as I unpack my mac, I'll start working on getting this screen wired up, too.

Prooobably. I don't think it comes with very many regressions to behavior, because the text mode was in really poor upkeep, and the platform code doesn't manage any of the lobby's state. It'll still function in the background, but yeah it'd be good if it displayed something to confirm this fact.

I don't really want to re-introduce text mode as it was before because it made the code a lot more complex, but it should be possible to add prints each time the widget is updated. That isn't quite the same thing as having a GOOD text mode, but it will capture similar functionality as it had before. I'll do this shortly

> Longterm, a headless server mode will be worked on ([Headless mode #449](https://github.com/UZDoom/UZDoom/issues/449)). Would it not make sense to keep the textmode lobby around for that? Or would building an actual tui from the ground up make more sense?

The current requirements for hosting a game is either reaching a specific player count OR mashing the lobby UI's "GO" button, and neither are very conducive to a 24/7 server. I think the plan for true dedicated servers is to combine headless mode with in-game joining, and dedicated servers just would skip the lobby entirely. I could be wrong though, asking @Boondorl to be sure.

To be clear I'm not against having a proper text mode, but IMO the old one was nested in too much platform-specific code. It'd be nice if Windows could use it too, and it became a first-class feature instead of something to neglect ^^;

the-phinet:
> I'm not confident in my stance there though, so I'll do it anyways if you still feel strongly about it.

I don't feel strongly about it at all. I trust your judgement more than I trust mine 😅

> Text mode stuff 

By the sounds of it, stripping it out is the correct thing to do. I think waiting for Boon's input would make the most sense 

---

Other than that, this looks good to me. I'll test it a bit more and take a look at the code as soon as I can.

the-phinet (dismissed):
Works perfectly for me. I have not had a chance to test on the mac, but that is rather low priority for me, and will take me a (long) while to be able to test.

the-phinet (approved):

Issue #900, opened 2026-01-11
Abort Screen not implemented on macOS


the-phinet:
MacOS ~~and Linux~~ must open the terminal to see logs

PR #958, opened 2026-01-26
Show error pane widget on Linux

TehRealSalt:
Partially addresses #900, is related to #260

image

Allows the error pane widget to appear on Linux. The Copy Clipboard and Restart buttons should function completely.

#957 had a clear path to refactoring to make its widget happen on all systems. However, the flow for showing errors is very specialized, even if it might be possible to reconcile. I've made the call to do purely as a Linux-only change since it's what I can test, and I don't want to make guesses for Mac.

## Checklist


the-phinet (dismissed):
Nice :)

Looks good and works well!

the-phinet (approved):

Issue #586, opened 2025-11-25
Avoid rollback on UI-scoped fields


Boondorl:
Since these are ran locally, they have no reason to roll back and it's likely to cause issues. Once the rollback is updated to the serializer, use special logic here that ignores UI fields (and also marks transient things regardless since the rollback is not the same as proper saving).

PR #963, opened 2026-01-31
Reworked prediction rollback system

Boondorl:
This was previously slated for 5.1, but it seems some new 5.0 features have been particularly nasty about triggering crashes from the existing memcpy method. Since it was already unstable, I decided to take the months of work I had built up slowly chipping away at this rewrite and finish it now. This is still largely untested so will need a ton of verification before it can be safely merged, but might be necessary for 5.0. If it's deemed too large of a rework, client-side Thinkers will need to be temporarily disabled so that the predictor can be turned back off in singleplayer without causing behavior mismatch problems with multiplayer in the future.

The new strategy uses the serializer to create a snapshot instead of memcpy. Not only is this far safer as it will ignore references to stale pointers, but it also backs up complex types like maps and arrays correctly. `norollback` has been added as a keyword for data-scoped fields to avoid this behavior, and UI fields and Objects are automatically marked to not be rolled back, making it play nicer with client-side features. Physics relinking is also much safer now as this won't copy the list but will dynamically link it back into the correct spot, avoiding crashes from malformed lists.

To accommodate this, the concept of what's client-side or not has been necessarily expanded. All UI and data-scoped Objects are now automatically marked as client-side which also allows for better error handling. To avoid too much havoc, play-scoped Objects that were spawned while predicting will now be wiped before rolling back (this will be needed in the future). Any fully backed up Object that was destroyed while predicting will be recreated (others will be turned into null pointers when restoring references).

To avoid too much overhead from the more complex snapshotting behavior, rollback will no longer occur every tick but only when a new server tick is received. Thinker and TID relinking are not implemented so have to still be avoided while predicting, but this shouldn't result in any broken behaviors, just possible desyncs.

Implements #585 and #586.

## Checklist


z9ctn4fn:
HDest community member here. We tried playing Doom 2 with HDest and UaS, and desyncs were still happening reliably.
If you use the IED item, it causes a desync- maybe take a look into that? We were also having lots of issues connecting to eachother.

mc776:
I would love to try to debug this but pretty much every single noun in the main post here is total gibberish to me 💀

That said... @z9ctn4fn is it *just* the IED that's causing so much problems? What about the ladder, the DERP/ERP, and the door buster?
EDIT: And were you guys actually running the build based on this PR, or something else? (I just assumed it was about this PR because it was a reply to it...)

Boondorl:
> EDIT: And were you guys actually running the build based on this PR, or something else? (I just assumed it was about this PR because it was a reply to it...)

Cali pointed them towards this PR alongside the artifacts so I assume it's this one. I'm not really sure if I could debug specific items, I took a peek and there's just a lot going on in there for someone who has no idea how the code base works. If it's consistently that item it at least gives a good clue.

The connection issues are strange because at this point the only person guests need to be in contact with is the host. If they're struggling to do that, that sounds like an issue with ISPs simply not allowing connections. It happens and unfortunately there isn't anything we can do about it on the engine side. It's possible CGNAT is at play as that's an infamous problem in general. The only real solution there is to call your ISP and tell them to give you a real IP.

RicardoLuis0:
> HDest community member here. We tried playing Doom 2 with HDest and UaS, and desyncs were still happening reliably. If you use the IED item, it causes a desync- maybe take a look into that? We were also having lots of issues connecting to eachother.

also, this doesn't remove _all_ desyncs, it only removes desyncs that were happening from the previous rollback code (that used memcpy instead of snapshots), if code does something that fundamentally desyncs, it will still desync

(in fact, item usage isn't predicted at all, so this really wouldn't change whether it desyncs or not either way, unless the item changed something in the player pawn that impacted prediction)

Boondorl:
Possibly silly question, but we're certain everyone was using the correct version of Hideous Destructor too, right? Because if someone is running an old version their game will still start, it'll just eventually desync if someone does something that isn't available on their end. I couldn't see anything in the item code that'd break it and I already fixed the bug with net events being dropped on packet resend

Boondorl:
> I would love to try to debug this but pretty much every single noun in the main post here is total gibberish to me 💀

Also, to reply to this, none of the engine specifics here are really of concern on your end. What I need is a 100% reproducible example. One of the hardest parts of debugging multiplayer is that people mess up the files a lot and until I can get proper file verification in, there's no way to determine if it's a true engine-level desync, if the mod is just doing something wrong, or if people have file mismatches and don't realize it because the engine isn't telling them. The latter is far more common than you'd think and is likely a big reason why group A reports no issues at all despite running mods while group B is reporting almost immediate desyncs every time with any mod.

z9ctn4fn:
Okay, second playtest results (this time on vanilla)
- 30-50 ms stable with occasional spikes vs 260 consistently ms with new netcode (even after adjusting net settings) although it dropped to 120 when a different player became host (although I suspect this is due to location and server architecture, P2P vs server/client)
Going to test the new version that just came out
- Note: We managed to get a desync in HDest somehow, don't know exactly how or why- will try to reproduce

RicardoLuis0:
> Okay, second playtest results (this time on vanilla)
> 
> * 30-50 ms stable with occasional spikes vs 260 consistently ms with new netcode (even after adjusting net settings) although it dropped to 120 when a different player became host (although I suspect this is due to location and server architecture, P2P vs server/client)

one thing that should be noted is, with the old netcode, the latency reading is not very straightforward at all (i don't remember the specifics of how it's broken, but i'm sure boon can talk more about it)

z9ctn4fn:
2nd playtest results on new build-
- No desyncs, just lag due to server architecture and regions
Overall, i think it'd be safe to merge if P2P can be implemented to reduce lag.

Boondorl:
> one thing that should be noted is, with the old netcode, the latency reading is not very straightforward at all (i don't remember the specifics of how it's broken, but i'm sure boon can talk more about it)

The latency was weird and unreliable, but essentially, whatever your latency displayed was was the amount of delay in ticks (in ms form) your game had, which is the only metric that actually mattered. This varied highly depending on a given player's distance to other players so the experience could vary from "alright" to absolutely terrible even within the same game. The pacer was also extremely aggressive and if you were at an awkward distance it could cause your game to constantly hitch

TehRealSalt (approved):

PR #966, opened 2026-02-01
Added file verification for net games

Boondorl:
This will account for missing files, added files, file order, and mismatched hashes on top of the engine version. Files loaded via `-optfile` are ignored entirely. This also hooks the idea of optional resources up to the engine, marking them as appropriate.

## Checklist

### DIFF
--- a/src/common/engine/i_net.cpp
+++ b/src/common/engine/i_net.cpp
@@ -655,7 +655,7 @@ static void SendVerificationError(const sockaddr_in& to, const FVerificationErro
 		NetBuffer[5] = (ar->Size() >> 8);
 		NetBuffer[6] = ar->Size();
 		size_t i = 7u;
-		for (const auto& file : *ar)
+		for (auto& file : *ar)
 		{
 			memcpy(&NetBuffer[i], file.GetChars(), file.Len() + 1u);
 			i += file.Len() + 1u;
@@ -1038,7 +1038,10 @@ static FString ReadVerificationError(TArrayView<uint8_t> stream)
 		if (!fileSystem.IsOptionalResource(i))
 		{
 			const FString crc = fileSystem.GetResourceHash(i);
-			files[crc] = fileSystem.GetResourceFileName(i);
+			FString name = fileSystem.GetResourceFileName(i);
+			FixPathSeperator(name);
+			auto a = name.Split('/', FString::TOK_SKIPEMPTY);
+			files[crc] = a.Last();
 		}
 	}
 
--- a/src/common/engine/i_net.cpp
+++ b/src/common/engine/i_net.cpp
@@ -56,6 +56,7 @@
 #include "printf.h"
 #include "version.h"
 #include "widgets/netstartwindow.h"
+#include "filesystem.h"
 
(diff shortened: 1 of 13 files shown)


src/common/engine/i_net.cpp
```diff
@@ -984,6 +1027,74 @@ static bool HostGame(int arg)
 	return true;
 }
 
+static FString ReadVerificationError(TArrayView<uint8_t> stream)
+{
+	if (stream[0] == FVerificationError::VE_ENGINE)
+	{
+		return FStringf("Engine mismatch: host expected %d.%d.%d, got %d.%d.%d",
```
River-Salmon, line 1031: can we also throw this as a warning or something to Log to clearly indicate to the end user that a file mismatch occurred?

src/common/filesystem/source/filesystem.cpp
```diff
@@ -224,10 +224,11 @@ void FileSystem::DeleteAll ()
 bool FileSystem::InitSingleFile(const char* filename, FileSystemMessageFunc Printf)
 {
 	std::vector<std::string> filenames = { filename };
-	return InitMultipleFiles(filenames, nullptr, Printf);
+	std::vector<std::string> empty;
+	return InitMultipleFiles(filenames, empty, nullptr, Printf);
 }
 
-bool FileSystem::InitMultipleFiles (std::vector<std::string>& filenames, LumpFilterInfo* filter, FileSystemMessageFunc Printf, bool allowduplicates)
+bool FileSystem::InitMultipleFiles (std::vector<std::string>& filenames, std::vector<std::string>& optfilenames, LumpFilterInfo* filter, FileSystemMessageFunc Printf, bool allowduplicates)
```
River-Salmon, line 231: nit: If the filenames are immutable, std::vector<std::string_view>& probably preferred here

src/common/engine/i_net.cpp
Boondorl: I thought that's what I_NetError was for; it's just a wrapper for I_Error. If you mean the whole "error window not showing on Linux" situation, I think Salt was working on getting that working so it should hopefully disappear as an issue here soon

TehRealSalt: It routes to I_FatalError, which already handles printing the error to the log. Tested myself to be certain.

TehRealSalt (approved):
Code looks right + tested every possible error that could be thrown & connecting to a proper netgame, LGTM 👍

River-Salmon:
Excellent. LGTM then

CptSledge:
How are the odds of -optfile being finished by the time of 5.0? Not being able to skip file verification is a non-trivial issue for cosmetic mods with things like Hideous Destructor, as many players prefer to have their own sprite and sound replacements.

RicardoLuis0:
> How are the odds of -optfile being finished by the time of 5.0? Not being able to skip file verification is a non-trivial issue for cosmetic mods with things like Hideous Destructor, as many players prefer to have their own sprite and sound replacements.

it had too many issues, but maybe a `-visualfile` can be added, that only loads sprites/sounds/etc and isn't verified? @Boondorl

PR #974, opened 2026-02-04
Add support for Chinese Language

### LINKED ISSUE #886: please add Chinese language

the-phinet:
Discussion here #408
Resolves #886

## Checklist


the-phinet:
Okay, I think this is ready to be looked at. If it works and looks correct to ya'll, I'll clean up the commit history :D

dwing4g:
I have tested this branch. It failed to load WAD.
Rerun with debug, it will break at this line:
```
assert((0 == FindName(PredefinedNames[i], true)) && "Predefined name already inserted");
```
Because there are some conflicted names (case-insensitive) defined:
```
xy(LANG_no, "no")
xy(LANG_pt, "pt")
...
xx(No)
xy(LANG_PT, "pt")
```

the-phinet:
Thanks for catching that. Sorry, I have a bad habit of forgetting to test with mods. 

Edit: Actually why didn't this break when I was testing normally? Maybe I wasn't doing debug builds? In any case, it's fixed now

Edit again: It now checks on build. Should be impossible for that to happen again

TehRealSalt (dismissed):
LGTM. No regressions in behavior, my own mods w/ translations are still functioning

TehRealSalt (dismissed):

src/common/engine/stringtable.cpp
```diff
@@ -22,16 +22,16 @@
 **
 */
 
-#include <string.h>
-
+#include "basics.h"
```
leahanderson1, line 25: alright, just restructuring the headers, seems pretty normal

src/common/engine/stringtable.cpp
```diff
@@ -41,6 +41,216 @@ EXTERN_CVAR(Int, developer);
...
+// this lives somewhere we already have <windows.h>
+#endif
+
+//==========================================================================
+//
+// Map old semi-made-up language codes to IETF language tags
+//
+//==========================================================================
+
+inline void RemapLegacyLanguages(FName &name, FString &lang)
```
leahanderson1, line 71: just a remap function, normal too

src/common/engine/stringtable.cpp
```diff
@@ -296,37 +505,34 @@ void FStringTable::LoadLanguage (int lumpnum, const char* buffer, size_t size)
...
-						activeMaps.Clear();
-						activeMaps.Push(default_table);
-						hasDefaultEntry = true;
-					}
-					else
-					{
-						sc.ScriptError ("The language code must be 2 or 3 characters long.\n'%s' is %zu characters long.",
-							sc.String, len);
-					}
+					sc.ScriptError ("The language code may not be empty.");
```
leahanderson1, line 511: was a little confused at first, but nope, just git being weird, the code got moved down below

src/widgets/widgetresourcedata.cpp
```diff
@@ -122,20 +122,27 @@ std::vector<SingleFontData> LoadWidgetFontData(const std::string& name, bool roo
 	if (!stricmp(name.c_str(), "notosans"))
 	{
 		// to update/add fonts:
```
leahanderson1, line 124: lgtm, just adding the simplified chinese font it looks like

leahanderson1 (approved):
seems all fine and good to me, the PR does what it says on the tin pretty much.

RicardoLuis0 (approved):

PR #983, opened 2026-02-06
Cleanup

the-phinet:
- Fix a bunch of warnings
- De-tangle some imports

## Checklist

### DIFF
--- a/src/common/filesystem/source/md5.hpp
+++ b/src/common/filesystem/source/md5.hpp
@@ -66,6 +66,7 @@
    * efficiently on either one than if ARCH_IS_BIG_ENDIAN is defined.
    */
 
+#include <cstdint>
 #include <stddef.h>
 #include <cstring>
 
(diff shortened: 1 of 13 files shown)


src/common/filesystem/source/md5.hpp
```diff
@@ -196,7 +197,7 @@ namespace FileSys {
                      * On little-endian machines, we can process properly aligned
                      * data without copying it.
                      */
-                    if (!((data - (md5_byte_t const*)0) & 3)) {
```
RicardoLuis0, line 199: this is still missing the subtract after the uintptr_t cast, should be `if (!((reinterpret_cast<std::uintptr_t>(data) - 0) & 3)) {`

RicardoLuis0: wait nvm, i just realized that it's subtracting 0 not subtracting from 0, ignore this 😅

the-phinet: I genuinely love getting comments like this, since it confirms to me that people do double check my work. Sometimes I'm worried that things get added without a second thought. I test the heck out of my stuff, but I'm also notoriously dumb sometimes :sweat_smile:

RicardoLuis0 (approved):

Issue #948, opened 2026-01-24
Restart button fails, trying to load uzdoom as a PWAD(?)


jekyllgrim:
### UZDoom Version

g8591ce9e81

### Game

All

### Operating System

Windows 10

### OS Version

Windows 10 Pro

### Hardware

AMD Ryzen 7 9800X3D, GTX 1080

### Duplicate Confirmation

- [x] I have searched and not found a similar issue.

### Bug Description

If UZDoom closes on startup due to an error (such as a script error), pressing "Restart" in that window produces a `Cannot find wad 'uzdoom'` error. Even if the original issue was fixed in the background (such as, when working from an unpacked folder, you fix the offending script), you can't re-run UZDoom with the Restart button. The latest stable release doesn't have this issue.

### Configuration

```ini

```

RicardoLuis0:
as i said on discord, seems this has always happened yes, because it passes the whole of GetCommandLineW as the args to the new restarted proccess, but GetCommandLineW includes arg[0], the executable itself

https://github.com/UZDoom/UZDoom/blob/17b4f76c7f92c78789fa09467cda3a8eef4e6146/src/common/platform/win32/i_main.cpp#L336-L342

but it was silently ignored, so ever since uzdoom started erroring on unknown files with i_exit_on_not_found, this seems to have started being a problem

Enjay001:
Does this only happen under certain circumstances?  I've been working on a project in an unpacked directory structure all day, starting UZDoom (exact same version of UZDoom and also on Windows 10).  If there is an error, I leave UZDoom sitting there, go to my text file, fix it and press restart and it all works.

RicardoLuis0:
> Does this only happen under certain circumstances? I've been working on a project in an unpacked directory structure all day, starting UZDoom (exact same version of UZDoom and also on Windows 10). If there is an error, I leave UZDoom sitting there, go to my text file, fix it and press restart and it all works.

it only happens in 5.0 if i_exit_on_not_found is enabled, but the uzdoom executable is always passed as a parameter, it's just ignored on versions without i_exit_on_not_found

PR #990, opened 2026-02-07
Strip exe out of Windows restart args

TehRealSalt:
Resolves (potentially?) #948

This fix has been tested using Bottles on Linux. Please do not merge until someone has stated they've tested this on a real machine.

## Checklist

### DIFF
--- a/src/common/platform/win32/i_main.cpp
+++ b/src/common/platform/win32/i_main.cpp
@@ -338,7 +338,59 @@ int DoMain (HINSTANCE hInstance)
 		HMODULE hModule = GetModuleHandleW(NULL);
 		WCHAR path[MAX_PATH];
 		GetModuleFileNameW(hModule, path, MAX_PATH);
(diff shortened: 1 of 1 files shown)


the-phinet:
So, I'm not able to replicate the error Ash gets (no matter what the value of i_exit_on_not_found is), but I can see that the exe file gets loaded after hitting restart

Without your fix:
> Initial:
> image
> After restart:
> image

With your fix:
> Initial:
> image
> After restart:
> image

So things seem to work :D (tested this on actual windows (eww lol))

also `(I am making my distaste for this feature known)` is the best comment I have ever read

the-phinet (approved):

TehRealSalt:
> So, I'm not able to replicate the error Ash gets (no matter what the value of i_exit_on_not_found is), but I can see that the exe file gets loaded after hitting restart

I'll poke @jekyllgrim then. This should still go in because loading the exe as a file is stupid (lol), but I don't want to prematurely close the issue ^^;

Issue #1008, opened 2026-02-11
Frametime not stable


dwaradzyn:
### UZDoom Version

Locally built with this commit: 2026-02-11 09:23:23 -0500 b429b82b27 Strip exe out of Windows restart args

### Game

Doom

### Operating System

Linux x86_64

### OS Version

Linux nobara-pc 6.18.9-201.nobara.fc43.x86_64 #1 SMP PREEMPT_DYNAMIC Sun Feb 8 11:38:21 UTC 2026 x86_64 GNU/Linux

### Hardware

ASUS Zephyrus G16 2024 laptop with CPU: Intel Core 185H and GPU: RTX4090 Mobile

### Duplicate Confirmation

- [x] I have searched and not found a similar issue.

### Bug Description

For a simple level, such as map01 from Doom II and running UZDoom with no mods I would expect my 240Hz monitor to be fully saturated with no frame drops. Unfortunately, this is not the case. Gameplay feels a little bit stuttery and FPS is not stable and jumping between 220 and 241 with UZDoom built from latest sources.

Official UZDoom release (Linux-UZDoom-4.14.3.AppImage) is not affected. I get very stable 240FPS on official release. Also older commits (from Dec 2025) were not affected. The issue seems to be introduced fairly recently. If I manage to find which commit introduced this issue, I will post more information here.

To demonstrate the issue I launched map01 and pressed "Turn right" key. I recorded performance metrics with MangoHud and then charted frametimes with LibreOffice:

<img width="1485" height="857" alt="Image" src="https://github.com/user-attachments/assets/c4897210-ba64-421e-9d44-f55866e66607" />
<img width="1182" height="680" alt="Image" src="https://github.com/user-attachments/assets/9a7292c0-5294-4833-a7c7-b2f952a5f281" />

Here are source MangoHud csv files:

[MangoHud metrics for latest build.csv](https://github.com/user-attachments/files/25242597/MangoHud.metrics.for.latest.build.csv)
[MangoHud metrics for official UZDoom.csv](https://github.com/user-attachments/files/25242596/MangoHud.metrics.for.official.UZDoom.csv)

### Configuration

```ini

```

RicardoLuis0:
could it just be shader caching stutter due to the existing cache not being valid for the new version?

Boondorl:
[...]

PR #1015, opened 2026-02-12
fix up serializer shit

RicardoLuis0:
should help at least a little bit with #1008 but i don't know if it'll fully fix it, i still have some other optimizations in mind though

## Checklist


RicardoLuis0:
(as a bonus this should make save loading a lot faster for actor/object heavy mods)

Boondorl (approved):

drfrag666:
Wenas. But what is this shit for? xD. I've merged it and seems to be applicable.
Well saving and then loading works.

RicardoLuis0:
> Wenas. But what is this shit for? xD. I've merged it and seems to be applicable. Well saving and then loading works.

basically looking up keys for reading serialized data was using a linear array search not a map without that rapidjson define 💀 (because by default it preserves key order for whatever reason, and you need to explicitly tell it that performance is more important than that)

PR #1027, opened 2026-02-14
Keep load order when loading optional files

Boondorl:
This is needed to ensure assets aren't accidentally overwritten. Fixes #1011 and #1028

## Checklist
- [x] I have reviewed [CONTRIBUTING.md](../blob/trunk/CONTRIBUTING.md) and agree to its terms.

PR #1038, opened 2026-02-15
Fix counting iwadnum

dwing4g:
## Checklist

This patch resolved all the left localization issues in `Harmony` and some others.
Incorrect `iwadnum` causes the code to mistakenly process `DeleteLumps = ..."LANGUAGE"...` in the `iwadinfo.txt`.

### DIFF
--- a/src/d_iwad.cpp
+++ b/src/d_iwad.cpp
@@ -870,7 +870,7 @@ int FIWadManager::IdentifyVersion (std::vector<std::string>&wadfiles, std::vecto
 	int iwadnum = 1;
 	if (optional_wad && D_AddFile(optwadfiles, optional_wad, true, -1, GameConfig))
 	{
-		iwadnum++;
+		// iwadnum++; // optional_wad is always "game_support.pk3", that should be not counted
 	}
 
 	fileSystem.SetIwadNum(iwadnum);


RicardoLuis0:
removing it outright will also break the iwad count though

src/d_iwad.cpp
```diff
@@ -870,7 +870,7 @@ int FIWadManager::IdentifyVersion (std::vector<std::string>&wadfiles, std::vecto
 	int iwadnum = 1;
 	if (optional_wad && D_AddFile(optwadfiles, optional_wad, true, -1, GameConfig))
 	{
-		iwadnum++;
+		// iwadnum++; // optional_wad is always "game_support.pk3", that should be not counted
 	}
 
 	fileSystem.SetIwadNum(iwadnum);
```
RicardoLuis0, line 876: the `iwadnum++;` should be kept, but this line (`fileSystem.SetIwadNum(iwadnum);`) moved before the `if`

RicardoLuis0: wait no, hmm, thinking about it more, why is it breaking at all? that iwadnum++ is supposed to make the processing **skip game_support.pk3**, not count it

RicardoLuis0: i think something is rather more broken with the iwadnum logic than what seems in the surface

RicardoLuis0: ah, is it because game_support.pk3 is marked an optional file? if so this needs #1027 to fix that before it can be properly looked over again

dwing4g: You are very perceptive. This PR might not be the most appropriate way. A better fixing would require a thorough review of the relevant code.
In addition, this issue was introduced in this commit:
```
Revision: 66de25698ab4bfecabc9edf8e3b908926348d7d8
Author: Boondorl <boondorl@live.com>
Date: 2026-02-01 15:52:50
Message: Added file verification for net games
```

dwing4g:
https://github.com/UZDoom/UZDoom/pull/1027 has fixed this issue perfectly, so this PR should be canceled‌.

PR #1045, opened 2026-02-17
Remove hard-wrapping

the-phinet:
## Checklist


mc776:
I wonder why such a wide line width had originally been chosen anyway... seems like it would've been the worst of both worlds

RicardoLuis0 (approved):

PR #1046, opened 2026-02-17
Fix casting integer to texture ID not working

nikitalita:
## Checklist

Type cast expressions on ints to TextureIDs don't currently work because that eval branch is missing in the compiler. Adding that fixes the issue.

### DIFF
--- a/src/common/scripting/backend/codegen.cpp
+++ b/src/common/scripting/backend/codegen.cpp
@@ -1934,6 +1934,14 @@ FxExpression *FxTypeCast::Resolve(FCompileContext &ctx)
 		delete this;
 		return x;
 	}
+	else if (ValueType == TypeTextureID && basex->IsInteger())
+	{
+		basex->ValueType = TypeTextureID;
+		auto x = basex;
+		basex = nullptr;
+		delete this;
+		return x;
+	}
 	else if (ValueType->isClassPointer())
 	{
 		FxExpression *x = new FxClassTypeCast(static_cast<PClassPointer*>(ValueType), basex, Explicit);


the-phinet:
It looks like this resolves this comment left by a user on GZDoom, is this correct?

> Also please return the ability of free conversion between int and TextureID: as far as I see, this is safe, because every function with work with textures have a FTextureManager::ResolveTextureIndex() call, and, moreover, TextureID is actually an int. I ask to return it because three of my old flagman projects are stuck and frozen, and I also almost froze my development in GZDoom too already for two years because I always try to make my modifications as compatible as possible, usually from my favorite QZDoom 2.1 (≈ GZDoom 3.4.0) and up to the most modern engine versions; but I just know that no modern engines will work with my mods. This dejects and fights away the desire to Work and Create.

the-phinet (approved):

nikitalita:
I think so

the-phinet:
Hey @MorthimerMcMare,

I noticed you left some feature requests [here](https://github.com/ZDoom/gzdoom/security/advisories/GHSA-prhc-chfw-32jg)

It looks like this fixes one of them. If the other request is still something you'd like to see changed, could you open an issue in this repo?

drfrag666:
Wrong user? I see @madame-rachelle  wrote that message. I've merged this BTW. Y wenas.

RicardoLuis0:
> Wrong user? I see @madame-rachelle wrote that message. I've merged this BTW. Y wenas.

@madame-rachelle is the one who approved/published the security notice but @MorthimerMcMare is the one who reported/wrote it
image

MorthimerMcMare:
Hello everyone!

Yes, I think all my wishes from the inappropriate Security section are already satisfied :) . Sorry that I forgot to open the issue.

> If the other request is still something you'd like to see changed, could you open an issue in this repo?

In future, of course!

PR #1050, opened 2026-02-18
Added MOVINGSPECIAL flag

Boondorl:
This is for items that can be moved around and should be picked up by anything they touch that's capable of doing so.

## Checklist

### DIFF
--- a/src/playsim/actor.h
+++ b/src/playsim/actor.h
@@ -439,6 +439,7 @@ enum ActorFlag9
 	MF9_ISPUFF					= 0x00000040,	// [AA] Set on actors by P_SpawnPuff
 	MF9_FORCESECTORDAMAGE		= 0x00000080,	// [inkoalawetrust] Actor ALWAYS takes hurt floor damage if there's any. Even if the floor doesn't have SECMF_HURTMONSTERS.
 	MF9_NOAUTOOFFSKULLFLY		= 0x00000100,	// Don't automatically disable MF_SKULLFLY if velocity is 0.
+	MF9_MOVINGSPECIAL			= 0x00000200,	// For special items, let them interact with others while moving.
 };
 
 // --- mobj.renderflags ---
--- a/src/playsim/p_map.cpp
+++ b/src/playsim/p_map.cpp
@@ -1721,6 +1721,11 @@ bool PIT_CheckThing(FMultiBlockThingsIterator &it, FMultiBlockThingsIterator::Ch
 	{ // Can be picked up by tmthing
 		P_TouchSpecialThing(thing, tm.thing);	// can remove thing
 	}
+	else if ((tm.thing->flags9 & MF9_MOVINGSPECIAL) && (tm.thing->flags & MF_SPECIAL) && (thing->flags & MF_PICKUP))
+	{
+		// Also allow moving items to situationally get picked up.
+		P_TouchSpecialThing(tm.thing, thing);
+	}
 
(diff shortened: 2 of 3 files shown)


mc776:
Can I suggest that it be called something else though? There are so many things in ZDoom with "special" in it that it gets really confusing.

Boondorl:
I'm not really sure what else to call it tbh. MOVINGPICKUP can't be used here because technically any Actor can have SPECIAL so it's not really bound to items specifically

mc776:
Does this flag actually affect non-pickup behaviour though?

RicardoLuis0:
> Does this flag actually affect non-pickup behaviour though?

well, pickups happen on the touch virtual, so if pickups work anything else would, because it means the touch function is being called properly (+SPECIAL for items doesn't mean action specials, it just means that they trigger touch when they, well, touch, an actor with the +PICKUP flag)

Boondorl:
+SPECIAL works for everything, so technically this can apply to things that aren't items at all, granted it'll only apply to things with +PICKUP. If this sounds too niche to be useful, consider one of Strife's enemies literally relies on this, so it's a feature used outside of items even in vanilla handling.

Maybe something like +SPECIALCANMOVE would be better, that way things that impact SPECIAL always start with SPECIAL while things that _call_ a special have SPECIAL at the end (e.g. BUMPSPECIAL)

mc776:
OK so just to be clear, all this "SPECIAL" stuff is totally separate from the "special" that's called with +USESPECIAL and +BUMPSPECIAL, right? (which, just to throw it in there, have nothing to do whatsoever with +SPECIALFIREDAMAGE and none of these in turn have anything to do with +SPECIALFLOORCLIP...)

If so I think the best option is "+CANBEPICKEDUPWHILEMOVING" and "+SPECIAL" deprecated in favour of "+CANBEPICKEDUP".

But that said... what even does this do? It's already possible to pick up items even if they've been given chase states or velocity, so is this about the item also being able to move in to a non-moving picker-upper and be picked up that way despite not being the passive recipient of the bump?

Boondorl:
> But that said... what even does this do? It's already possible to pick up items even if they've been given chase states or velocity, so is this about the item also being able to move in to a non-moving picker-upper and be picked up that way despite not being the passive recipient of the bump?

Items can only ever be picked up when the other Actor is moving. This means if you have some kind of vacuum effect in your game, the player can stand still and never actually get it. This just allows the inverse case to happen: the item will now check against the other thing on movement to see if it can be picked up

MrRaveYard:
`MOVINGSPECIAL` or `SPECIALMOVE` is not intuitive, especially if you're not familiar with inner workings of the engine, but I suppose that's 95% of all the other flags.

For example, I'd do these changes:

`SPECIAL` -> `TOUCHABLE`
`PICKUP` -> `CANTOUCH`
`USESPECIAL` -> `RUNSPECIALONPLAYERUSE`
`BUMPSPECIAL` -> `RUNSPECIALONPLAYERBUMP`
`TOUCHY` -> `DIEONCOLLISION`

So it would be:
`MOVINGSPECIAL` -> `INSTIGATETOUCH`

But if I had no choice but to pick new name only for this flag with current context of everything...
`INSTIGATEPICKUP` or `INSTIGATESPECIAL`

mc776:
I like these!

RicardoLuis0 (changes requested):
re-reading the comments, probably needs better naming for the flag

PR #1052, opened 2026-02-19
Skip nightly tag when creating version

the-phinet:
Should fix versioning for automated build systems

## Checklist

### DIFF
--- a/.github/workflows/build_core.yml
+++ b/.github/workflows/build_core.yml
@@ -35,6 +35,8 @@ jobs:
     - name: Setup
       id: ghd
       uses: proudust/gh-describe@v2
+      with:
+        exclude: "nightly"
 
     - name: Setup
       uses: ilammy/msvc-dev-cmd@v1
--- a/.github/workflows/upload_nightly.yml
+++ b/.github/workflows/upload_nightly.yml
@@ -18,6 +18,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - id: tag
+        uses: proudust/gh-describe@v2
+        with:
+          exclude: "nightly"
+
       - name: Download Artifacts
         uses: actions/download-artifact@v4
         with:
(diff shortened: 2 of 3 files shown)


RicardoLuis0 (dismissed):

RicardoLuis0:
downloaded from: https://github.com/UZDoom/UZDoom/actions/runs/22165410803
image
image

the-phinet:
I assume that's due to the git-describe action overriding the cmake script

~~I'll look into that in the morning~~ Done

RicardoLuis0 (changes requested):
ci files are missing the file extension 😔

Image

also it's grabbing the wrong tag?

Image
Image

(also just realized i missed the `g` in the 5.0pre tag, but that shouldn't matter for this)

the-phinet:
> Re: missing extensions

Nothing's changed in the code regarding that, and extensions exist just fine on my end

image

> Re: missing tag

Probably a side-effect of not being run on trunk

---

Neither of these are worth blocking this over. If it doesn't work when merged, I'll take a look.

The thing that's really annoying about GitHub actions is that they have different results based on how, when, and where they are run. Plus a 20+ minute iteration time makes debugging slow. Incremental changes in-situ are the best way to get to the correct state of things

RicardoLuis0:
> > Re: missing tag
> 
> Probably a side-effect of not being run on trunk
> 

i thought the whole point of PR continuous integration is that it ran after a simulated merge?

RicardoLuis0 (approved):

RicardoLuis0:
but fair, it can always be fixed later if it doesn't work

RicardoLuis0:
still no extension, and still wrong tag 😔
image
image
https://github.com/UZDoom/UZDoom/actions/runs/22248348556

the-phinet:
Nothing has changed with the file extension. Was this a problem in the past? My browser downloads them just fine.

Do they download as zips when grabbed from the release page?

image

The no tag bit is weird. I'll look into it tomorrow. I wonder If it's the describe action? I'll see what cloning with tags and manually doing a git describe gets us. 

It's actually not just a bit weird, it's really weird. It's skipping quite a few tags...

RicardoLuis0:
> Nothing has changed with the file extension. Was this a problem in the past? My browser downloads them just fine.
> 
> Do they download as zips when grabbed from the release page?

releases page seems to be proper zips, just not directly from the CI

PR #1057, opened 2026-02-21
Add Option to disable Lost Souls conting towards the total monster kill count

### LINKED ISSUE #648: Add compatibility option to make Lost Souls not count toward the monster percentage.
### Game

Doom

### Problem Description

Ever since Doom 2, the Lost souls have not counted as monsters, and for a good reason considering the Pain Elementals can get them stuck in walls screwing over UV-MAX players. For a while GZDoom has had Lost Souls permanently tagged as +COUNTKILL for a while due to... him saying it's due to the programmer's laziness, even though they were counted prior to Doom v1. 666.

### Feature Description

I feel like it would be convenient for us players to have an option to turn off Lost Souls counting towards the percentage, it would also save time for us to not go into the .pk3 file and edit the monster code, maybe you could use A_ChangeFlag.

### Additional Media

Image

a little mockup of how I think it would look, either map or actor behavior is fine.

heiko123abc:
## Checklist

Fixes #648.

I have added a comp flag to the menu that adjust the lost souls to not count when enabled using Zscript.

Disclaimer: I do not have much Zscript experience. I just looked at the wiki and methods to add it in.

I have tested it and it works for single lost souls and those spawned by pain elementals.

grafik


Boondorl:
I like the idea but I think this might be best handled in a compat2 cvar change override rather than on the Actor itself. I might also consider adding an Actor flag, something like COUNTOPTIONAL, and loop through all entities with that, that way people who want their Lost Soul variants to always count can do so.

I'd have to think about what the best way to handle it would be since I could see a few routes here

mc776:
+SOULCOUNTKILL? (by analogy to the old +SOULTRANS)

wadsrc/static/zscript/actors/doom/lostsoul.zs
```diff
@@ -46,6 +46,7 @@ class LostSoul : Actor
 	States
 	{
 	Spawn:
+		SKUL AB 10 BRIGHT compatChange;
```
MrRaveYard, line 49: Altering vanilla monster behavior/states that have been there since 1993 seems like a terrible precedent, even if this slight delay in Spawn likely would have minor significance.

Not to mention changing this flag in the middle of the game would produce slightly weird and inconsistent results. Lost souls that are already active wouldn't fall under it.

More proper solution would be instead call `A_ChangeCountFlags` only when the flag changes. That is: the changing of the flag would run code which would scan all thinkers for Lost Souls and do the thing.

I also honestly don't know if this would break Dehacked compatibility. That would be my concern as well.

RicardoLuis0: it will 100% break dehacked

RicardoLuis0: and yeah, boon's suggestion of a new flag really does seem like the best way to go about it

PR #1058, opened 2026-02-22
Fix calculation of turnheld to use frame time, which improves the feeling of pressing the turn keys and the speed of moving around.

Firebrand-mx:
## Checklist

### DIFF
--- a/src/g_game.cpp
+++ b/src/g_game.cpp
@@ -689,6 +689,7 @@ void G_BuildTiccmd (usercmd_t *cmd)
 	}
 	else
 	{
+		turnheld += TicDup;
 		int tspeed = speed;
 
 		if (turnheld < SLOWTURNTICS) tspeed += 2; // slow turn
--- a/wadsrc/static/shaders/glsl/main.fp
+++ b/wadsrc/static/shaders/glsl/main.fp
@@ -88,7 +88,7 @@ vec3 rgb2hsv(vec3 c)
 
 float grayscale(vec4 color)
 {
-	return dot(color.rgb, vec3(0.3, 0.56, 0.14));
+	return dot(color.rgb, vec3(0.2126, 0.7152, 0.0722));
 }
 
 //===========================================================================
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -262,3 +262,4 @@ Frank Kloster
 Santtu "MFG38" Pesonen
 heiko123abc
 Piotr "stared" Migdał
+Francisco Ortega


the-phinet:
Hey, thanks for your interest in making UZDoom better!

Could you explain in detail what this change is doing, and why? Since `gametic` constantly increments, `turnheld` will now behave differently based on the game's duration.

the-phinet:
Can you please explain the rationale behind these changes? I'm not sure I fully understand. At the very least, changed to player movement like this need to be behind an option.

Firebrand-mx:
> Can you please explain the rationale behind these changes? I'm not sure I fully understand. At the very least, changed to player movement like this need to be behind an option.

Sorry for the delay of response, I have been a bit busy :), the turn speed when using a keyboard is way too slow after the recent changes to joystick support, this is to improve that feeling of slowness when using a keyboard. You are right that using gametic would be incorrect here, the correct fix is to simply use TicDup for turnspeed variable, I already have the proper fix on my end.

the-phinet:
Yeah, you're totally correct. Thanks for catching the regression. 

Unless I'm mistaken, line 678 shouldn't be commented out. The joystick curves should handle any smoothing needed, not the turnspeed var.

Firebrand-mx:
> Yeah, you're totally correct. Thanks for catching the regression.
> 
> Unless I'm mistaken, line 678 shouldn't be commented out. The joystick curves should handle any smoothing needed, not the turnspeed var.

Alright, I have fixed the commented line, I also added a bugfix for the tonemap.fp shader file, to use vec3 type on the uncommented operations, that should be all for this request, thanks!

the-phinet (dismissed):

wadsrc/static/shaders/glsl/main.fp
```diff
@@ -88,7 +88,7 @@ vec3 rgb2hsv(vec3 c)
 
 float grayscale(vec4 color)
 {
-	return dot(color.rgb, vec3(0.3, 0.56, 0.14));
```
the-phinet, line 91: Please add this in a separate PR

the-phinet (changes requested):

src/g_game.cpp
```diff
@@ -689,6 +689,7 @@ void G_BuildTiccmd (usercmd_t *cmd)
 	}
 	else
 	{
+		turnheld += TicDup;
 		int tspeed = speed;
 
 		if (turnheld < SLOWTURNTICS) tspeed += 2; // slow turn
```
RicardoLuis0, line 695: i don't think ticdup is the right thing to use here, if the slow turning lasts too long, wouldn't tweaking SLOWTURNTICS be the proper fix? instead of incrementing turnheld by a random variable so that the slow turn check fails earlier

RicardoLuis0: or is there something else i'm missing?

the-phinet: It's adding a line back for a branch that I didn't consider. https://github.com/UZDoom/UZDoom/blob/125190b42fcf49629d0c1294c7ec424871d97038/src/g_game.cpp#L585

wadsrc/static/shaders/glsl/main.fp
the-phinet: It's switching from a simplified rec601 to a more modern rec709 luminance formula. I'll just approve it as-is.

the-phinet (approved):

PR #1063, opened 2026-02-22
set LTO on by default

Gutawer:
pretty much just a free perf win if the compiler supports it, only downside is the link time being heightened when in use

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,21 @@ endif()
 
 project(UZDoom)
 
+# make all targets use LTO - want to do this early in the cmake to ensure it applies to any targets
+option( FORCE_NO_LTO "Turn off LTO even if it would be possible to use it." NO )
+
+if(NOT FORCE_NO_LTO)
+	include(CheckIPOSupported)
+	check_ipo_supported(RESULT ipo_supported OUTPUT ipo_error)
+	if( ipo_supported )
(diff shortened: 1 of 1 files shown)


RicardoLuis0 (approved):

CMakeLists.txt
```diff
@@ -31,6 +31,21 @@ endif()
 
 project(UZDoom)
 
+# make all targets use LTO - want to do this early in the cmake to ensure it applies to any targets
+option( FORCE_NO_LTO "Turn off LTO even if it would be possible to use it." NO )
```
the-phinet, line 35: Out of curiosity, why would someone want to disable this? I've never been able to wrap my head around giving options like these.

the-phinet (approved):

CMakeLists.txt
Gutawer: fair question - my main reasoning for giving it a switch is just in case the time it takes to run LTO bothers anyone on a lower-core machine who's trying to contribute, since honestly it takes ~10 seconds (from memory) even on my 32-thread one lol - and it's a price you pay on every compile, even incremental ones where only a single object file is compiled

the-phinet: lol fair. I always take compilation time as an excuse to go get a glass of water

TehRealSalt (approved):

PR #1065, opened 2026-02-22
Replace the FName implementation with a not-hand-rolled hashmap

Gutawer:
The FName constructor was measurably slow before and was pretty much a hand-rolled hashmap implementation. This PR brings in the Abseil library (Apache-licensed) which has one of the best hashmap implementations out there (the fastest I know of is in Boost, but Boost is an even heavier library to bring in for just one type...). This also brings in some nice functions for string manipulation that seem weirdly missing from the C++ STL

Let me know if the Abseil bit is unwanted - it should be easy to put it onto `std::unordered_map` instead at some perf loss

Would recommend primarily just reviewing the second commit - the first one just brings in the lib

## Checklist


RicardoLuis0:
> Let me know if the Abseil bit is unwanted - it should be easy to put it onto `std::unordered_map` instead at some perf loss

i'd say that only compile times matter really, if it increases first-compilation times by a few minutes fine, if it increases it by 30+ minutes then an alternative might be better, or else CI will be painfully slow

src/common/utility/name.h
```diff
@@ -45,28 +47,28 @@ class FName
 {
 public:
 	FName() = default;
-	FName (const char *text) { Index = NameData.FindName (text, false); }
-	FName (const char *text, bool noCreate) { Index = NameData.FindName (text, noCreate); }
-	FName (const char *text, size_t textlen, bool noCreate) { Index = NameData.FindName (text, textlen, noCreate); }
-	FName(const FString& text) { Index = NameData.FindName(text.GetChars(), text.Len(), false); }
-	FName(const FString& text, bool noCreate) { Index = NameData.FindName(text.GetChars(), text.Len(), noCreate); }
+	FName (const char *text) { Index = NameManager::Instance().FindName (text, false); }
```
RicardoLuis0, line 50: other than that just, i think turning this into a function call instead of just a static variable will make name lookups in debug builds even slower wouldn't it?

Gutawer: the change to use a function was largely because i was having issues with doing it as a static variable when using more modern data types, though to be honest i can't remember what they were now

regardless, i've tested on debug builds and this doesn't affect performance on a macro level, i think the CPU cache can handle the pattern just fine probably

(i do think that debug build perf should be a secondary priority to code readability anyway though and this singleton pattern is more ideal than an object-level static variable for some other reasons anyway)

Gutawer:
> > Let me know if the Abseil bit is unwanted - it should be easy to put it onto `std::unordered_map` instead at some perf loss
> 
> i'd say that only compile times matter really, if it increases first-compilation times by a few minutes fine, if it increases it by 30+ minutes then an alternative might be better, or else CI will be painfully slow

i do have a good CPU but it's a quite minor change in this aspect - 1393 object files to compile on a clean build before, 1642 after, and the abseil `.cc` files are usually quite small so it's only a minor addition to compile times

the-phinet:
The added library adds ~5min to a single build, ~10min to a full CI run. I think this is acceptable.

> it should be easy to put it onto std::unordered_map instead at some perf loss

I'm fine with adding the library, but in general I prefer using the standard library when possible. What are the differences in performance? If it's minor, it might be worth sticking to the stl

RicardoLuis0 (approved):

RicardoLuis0:
> I'm fine with adding the library, but in general I prefer using the standard library when possible. What are the differences in performance? If it's minor, it might be worth sticking to the stl

at the same time though, string->name conversions are one of the biggest performance hogs in zscript, so any little win there is more than worth it

MajorCooke:
Curious, what about the reverse way, name -> string? Are those just as costly?

RicardoLuis0:
> Curious, what about the reverse way, name -> string? Are those just as costly?

not at all, it just indexes into the name list, instead of needing hashing

PR #1067, opened 2026-02-22
switch RNG over to simpler but better PCG generator

Gutawer:
Removes the SFMT that's been in use for ages in the engine in favor of a more modern algorithm, PCG. Mersenne Twister RNGs are no longer desirable - they have a huge amount of state while still failing more statistical tests anyway. PCG is faster even without being a larger explicitly-SIMD library and has a state of a single 64-bit integer, which is the primary motivation for switching - it makes the rollback state smaller and therefore faster to deal with.

For backwards compatibility with saves, the RNGs are serialized as an object containing an array of two 32-bit ints. This'll allow the engine to take an older save and put its MT state - although truncated - into the PCG state.

I've also cleaned the RNG code up a little bit with regards to floats, so we just have one main float gen implementation. This is optimized for speed, recognizing that game dev use cases don't need the better output that more advanced algorithms for float generation can give.

## Checklist


RicardoLuis0:
looks good but i do still want to test it with a few mods before approving

src/common/engine/m_random.h
```diff
@@ -41,7 +45,24 @@ class FRandom : public SFMTObj
...
 	int Seed() const
 	{
-		return sfmt.u[0] + idx;
+		return s32[0] ^ s32[1];
+	}
+
+	union {
+		std::array<uint32_t, 2> s32;
+		uint64_t s;
+	};
```
kcat, line 54: Not sure how much we care about correctness here, but using a union to type-pun like this is UB in C++, prone to breaking in certain scenarios. It's more correct to use `std::bit_cast` to transform between two different equally-sized types like this (you can just store the `uint64_t s;`, and make a local copy `auto s32 = std::bit_cast<std::array<uint32_t, 2>>(s);` in functions that need it; compilers are very good at optimizing around this and producing the same code). I'm also not sure on the status of big-endian support in the engine, and if the order of the two halves is important.

RicardoLuis0: > I'm also not sure on the status of big-endian support in the engine, and if the order of the two halves is important.

pretty sure big endian support is pretty busted right now, but i don't know if it's an easy fix, or a fundamental design issue, because i do think things did at one point work on big endian systems

MajorCooke:
> looks good but i do still want to test it with a few mods before approving

I'd suggest getting this merged sooner than later so people can start testing it out. If there's actually any problems with it, we can always revert it.

the-phinet (approved):
Code looks good. Played boa for a couple hours using this and found no ill effects.

PR #1069, opened 2026-02-23
ZMusic AlsaMIDIDevice: fix tempo not setting correctly on music loop

charlesthobe:
This fixes a bug that affects playback of midi tracks which have different initial tempo compared to the last tempo changing event such as E4M9 in FreeDoom 0.13.0

## Checklist


charlesthobe:
Contributing upstream only to avoid work duplication, sorry for the bother.

PR #1070, opened 2026-02-24
fix MDL_ROTATING

### LINKED ISSUE #1059: MODELDEF rotation offset issue
### UZDoom Version

nightly-8-ge85bf3f569

### Game

Doom

### Operating System

Windows 11

### OS Version

_No response_

### Hardware

_No response_

### Duplicate Confirmation

### Bug Description

Model rotation as defined through the MODELDEF lump seems to have broken since the last major release.  In 4.14.3, some basic MD3 models with rotations defined through MODELDEF look like this:

https://github.com/user-attachments/assets/ae6e7d6d-8d08-4c03-803c-6c45178bdb9a

But in the latest nightly build (and at least one earlier beta build too) they look like this:

https://github.com/user-attachments/assets/6aff9392-639e-411f-8493-972853655c48

For reference, the MODELDEF definition for these gears looks like variations of this:

```
Model Gear1AntiClockwise
{
	Model 0 "models/gear1.md3"
	Skin 0 "models/skins/gear_diffuse.png"
	Scale 1 0.83333 1
	Offset 0 0 0
	USEACTORPITCH
	USEACTORROLL
	ROTATING
	Rotation-Speed -1.0
	Rotation-Vector 1.0 0.0 0.0
	Rotation-Center 0.0 12.25 0.0

	FrameIndex NO3D A 0 0
}
```

### Replication

Example map available here: [https://drive.google.com/file/d/1bm5d7V0buDZEQYPvaCXwIvgd4NutVWWo/view?usp=sharing](https://drive.google.com/file/d/1bm5d7V0buDZEQYPvaCXwIvgd4NutVWWo/view?usp=sharing)

Run Map01 - the cog should be spinning in place, but rolls through the air in the current build.

### Configuration

```ini
Tested with a clean install
```

### Log

_No response_

RicardoLuis0:
before the refactoring, MDL_USEROTATIONCENTER used Rotation-Center as XZY but MDL_ROTATING used Rotation-Center as XYZ

after the refactoring, both used it as XZY, because i didn't even consider there could possibly be a discrepancy between two flags that use the same field for coordinates

this reverts MDL_ROTATING to XYZ for compatibility, and adds a new MDL_FIXROTATING flag to make it use the same coordinate system as MDL_USEROTATIONCENTER so that both flags can be safely combined

fixes #1059

## Checklist

### DIFF
--- a/src/r_data/models.cpp
+++ b/src/r_data/models.cpp
@@ -170,6 +170,9 @@ VSMatrix FSpriteModelFrame::ObjectToWorldMatrix(FLevelLocals *Level, DVector3 tr
 		objectToWorldMatrix.scale(1, stretch, 1);
 	}
 
+	bool rotating_xzy = (flags & MDL_ROTATING) && (flags & MDL_FIXROTATING);
+	bool rotating_xyz = (flags & MDL_ROTATING) && !(flags & MDL_FIXROTATING);
+
 	// Applying model transformations:
 	// 1) Applying actor angle, pitch and roll to the model
 	if (flags & MDL_USEROTATIONCENTER)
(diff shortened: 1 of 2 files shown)


the-phinet (approved):
Fixes the test map given in the issue. I don't have anything else to test it with, but it looks good to me

PR #1079, opened 2026-02-26
Fix ViewPos ownership cascade and add GetNetworkOwner getter

Alfaifi:
## Summary
- Fix bug in `SetNetworkOwner` where ViewPos cascade used `GetNetworkID()` instead of `GetNetworkOwner()`, causing ViewPos sub-objects to be skipped during ownership transfer
- Add `GetNetworkOwner()` ZScript native getter so modders can query which player owns an object (complementing the existing `SetNetworkOwner` setter)

## Test plan
- [ ] Build on all platforms with zero new warnings
- [ ] In multiplayer with a custom ViewPos actor, verify ownership cascade includes ViewPos when changing owner
- [ ] Verify ZScript `GetNetworkOwner()` returns correct `PlayerInfo` after `SetNetworkOwner()` and returns `null` for world-owned objects

### DIFF
--- a/src/common/engine/i_net.h
+++ b/src/common/engine/i_net.h
@@ -28,7 +28,6 @@
 
 #include <stdint.h>
 #include "m_argv.h"
-#include "tarray.h"
 
 inline constexpr size_t MAXPLAYERS = 64u;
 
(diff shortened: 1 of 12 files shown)


Alfaifi:
Opened against wrong repo by mistake, closing.

PR #1087, opened 2026-02-28
Add extra CI jobs

the-phinet:
- Makes debinfo build generate first, since release builds sometimes fail to compile when debug ones work
- Adds experimental ci builds

## Checklist


TehRealSalt (approved):
(Disclaimer: Not part of CI team)

All the artifacts that apply to my system don't have any regressions that I could notice, and the diff looks OK

Issue #1108, opened 2026-03-07
Separate HRTF setting for background music and in game sound effects


Zue5s:
### Game

All

### Problem Description

The current HRTF [On/Off/Auto] in OpenAL settings toggles HRTF for both music and sound effects, which compromises one or the other. if it's on, the music sound quality gets compromised, if it's off, the sound effects hurts to hear on headphones, reduces immersion and the ability to locate enemies based on sounds.

### Feature Description

Make two HRTF settings for music and in game sound effects, OR, add HRTF [On/Off/Follow Main] in each of the Midi players and Module replayers that are separate from the main HRTF toggle in OpenAL settings

Zue5s:
~~nevermind im stupid, just gotta set it to auto bruh~~
nope, still doesn't work, sorry i mistakenly closed the issue

RicardoLuis0:
i don't think openal-soft allows toggling it per-sound? only globally
but i'm not 100% sure, @kcat might be able to confirm or deny this

kcat:
In the case of background music, it can be toggled using the [`AL_SOFT_direct_channels_remix`](https://openal-soft.org/openal-extensions/SOFT_direct_channels_remix.txt) extension. It allows disabling virtualization/3D processing for individual sources, bypassing HRTF or the surround panning more generally, so it can be used to disable HRTF for background music like this. But it can't be used to disable HRTF for sound effects that need to retain 3D processing (including distance attenuation and panning).

Relatedly, I had [a PR for GZDoom](https://github.com/ZDoom/gzdoom/pull/1652) that added a Super Stereo option for background music, which is a method to "project" the stereo sound onto a continuous surround field with a customizable width (so you could restrict it down to mono, or spread it way out to envelop more of the sides), instead of simply playing it over two virtual speakers. An option like that could relatively easily be extended with a direct/non-virtual output setting alongside it. The primary thing that PR was lacking was the new text lines, as I never got a handle on doing them since they moved out of the code base.

the-phinet:
If you want to re-open that pr for uzd, I can add the strings for you (or walk you through it)

kcat:
Sure, I can do that.

Zue5s:
Let's goooo man

PR #1113, opened 2026-03-08
Music modes

kcat:
This adds options for playing music either with direct mixing (no HRTF or panning virtualization) or a Super Stereo algorithm (which projects the stereo track onto a continuous soundfield which can be widened or narrowed). These are handled with two new CVARs, `snd_musicmode` and `snd_superstereowidth`, accessible in game under the `OpenAL options` menu.

`snd_musicmode` is an integer value that can be set to `0` (Normal), `1` (Direct Mix), or `2` (Super Stereo). Any other value is treated as `0`. `snd_superstereowidth` is a float that can be set to anything between `0` and `1` (inclusive).

The Direct Mix option relies on the `AL_SOFT_direct_channels_remix` extension (added in OpenAL Soft 1.20.1) and Super Stereo relies on the `AL_SOFT_UHJ` extension (added in OpenAL Soft 1.22.0). There's no message if an extension is missing; if a given extension isn't available, the respective setting just won't do anything (it will be as if it's set to Normal).

This is expecting a few new strings for the menus:

```
OPTVAL_DIRECTMIX = "Direct Mix"
OPTVAL_SUPERSTEREO = "Super Stereo"
OPENALMNU_MUSICMODE = "Music Mode"
OPENALMNU_SUPERSTEREOWIDTH = "Super Stereo Width"
```

This is a redux of an older GZDoom PR, slightly extended to address issue #1108.

If there's anything more you need, let me know and I'll do what I can.

## Checklist


the-phinet:
Nice. The easiest way to add the needed strings is to open a pr in https://github.com/UZDoom/Translation, adding them to `engine/zdoom/en_US.po`. Alternatively, I can do that for you when this gets merged.

Zue5s:
This is strange, but the current HRTF makes the original sound effect of the game have too different of a frequency response; i.e. bass, middle, and treble is significantly affected, which is technically kinda realistic but messes up the original intent of the sound effect a little bit (maybe that's just me)

the-phinet (approved):

Issue #1116, opened 2026-03-08
Pink menus when using Vulkan on ARM based Macs


the-phinet:
### UZDoom Version

49ce791ffa033c8eff53e3c368d7a3ca99ed4459

### Operating System

Mac OS

### Hardware

M2

### Duplicate Confirmation

- [x] I have searched and not found a similar issue.

### Bug Description

Title screen (but not gameplay) is bright magenta when using Vulkan apple hardware.

This has been reported before in GZDoom, where the solution was using an old build of MoltenVK: https://github.com/KhronosGroup/MoltenVK/issues/2220

the-phinet:
So far I've noticed a couple minor issues with Vulkan being miss-configured. I've fixed some of these on a local branch, but it has not fixed the issue. I'm still looking into it. Any help would be appreciated.

RicardoLuis0:
if i'm reading correctly a PR in moltenvk added a bad optimization, and according to what https://github.com/KhronosGroup/MoltenVK/issues/2220 said, a dirty hack may be to add `VK_IMAGE_USAGE_SAMPLED_BIT` to this line (only on macos, as it may be a performance regression on other platforms):
https://github.com/UZDoom/UZDoom/blob/60d519677275c2ed9f6099967730e2bdb0afaeaa/src/common/rendering/vulkan/system/vk_renderdevice.cpp#L392

RicardoLuis0:
hmm wait no that'd only work if `needsReinterpretation` is true, and i have no idea what triggers that condition

RicardoLuis0:
but it's worth testing at least i guess

the-phinet:
Yeah, I saw that. Mvk is raising a number of angry messages at how we've configured vulkan, so I'm working towards fixing them before I blame the library. Since the issue seems isolated to zdoom, I assume we're doing something subtly wrong. Once I've got the things I know need fixing fixed, I'll reach out to the mvk folk for help.

RicardoLuis0:
> Yeah, I saw that. Mvk is raising a number of angry messages at how we've configured vulkan, so I'm working towards fixing them before I blame the library. Since the issue seems isolated to zdoom, I assume we're doing something subtly wrong. Once I've got the things I know need fixing fixed, I'll reach out to the mvk folk for help.

also it wasn't zdoom-only:
https://github.com/KhronosGroup/MoltenVK/issues/2220#issuecomment-2785715189

the-phinet:
[...]

PR #1127, opened 2026-03-10
Video mode switching imprivements

the-phinet:
- Video mode selector will now skip unsupported video modes.
- Default video mode on mac is OpenGL until #1116 is resolved
- Added default video mode override for those that may want it

## Checklist

### DIFF
--- a/src/common/rendering/v_video.cpp
+++ b/src/common/rendering/v_video.cpp
@@ -95,7 +95,7 @@ enum {
 #endif
 };
 
-CUSTOM_CVAR(Int, vid_preferbackend, BACKEND_DEFAULT, CVAR_ARCHIVE | CVAR_GLOBALCONFIG | CVAR_NOINITCALL)
+CUSTOM_CVAR(Int, vid_preferbackend, BACKEND_DEFAULT, CVAR_ARCHIVE | CVAR_GLOBALCONFIG)
 {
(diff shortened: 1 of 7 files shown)


RicardoLuis0 (approved):
looks good but might need some extra testing once it's in

src/common/rendering/v_video.cpp
```diff
@@ -93,43 +80,42 @@ CUSTOM_CVAR(Int, vid_maxfps, 500, CVAR_ARCHIVE | CVAR_GLOBALCONFIG)
...
-	default:
+	case BACKEND_OPENGL:
 		Printf("Selecting OpenGL backend...\n");
+		break;
 	}
 
-	Printf("Changing the video backend requires a restart for " GAMENAME ".\n");
+	static bool notice = false;
+	if (notice) Printf("Changing the video backend requires a restart for " GAMENAME ".\n");
+	else notice = true;
```
kcat, line 116: Is this backwards? It skips printing on the first invocation, and will print every time after the first.

the-phinet: That's intentional. This callback gets called during boot now, so I figured it would be best to not tell the user to restart when they just started the game 😅

kcat: I see, I see. Makes sense.

PR #1143, opened 2026-03-15
Add id24's MinRespawnTics and RespawnDice

RageSpark:
The default value of 0 uses the RespawnTime defined in the Skill Definition. Negative values are interpreted as seconds.

## Checklist

### DIFF
--- a/src/playsim/p_mobj.cpp
+++ b/src/playsim/p_mobj.cpp
@@ -402,7 +402,9 @@ void AActor::Serialize(FSerializer &arc)
 		("morphexitflash", MorphExitFlash)
 		("damagesource", damagesource)
 		("behaviors", Behaviors)
-		A("decalgenerator", DecalGenerator);
+		A("decalgenerator", DecalGenerator)
+		A("minrespawntics", MinRespawnTics)
+		A("respawndice", RespawnDice);
 
 		SerializeTerrain(arc, "floorterrain", floorterrain, &def->floorterrain);
 		SerializeArgs(arc, "args", args, def->args, special);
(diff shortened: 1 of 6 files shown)


src/playsim/p_mobj.cpp
```diff
@@ -5113,6 +5115,12 @@ void AActor::Tick ()
 	if (tics == -1 || state->GetCanRaise())
 	{
 		int respawn_monsters = G_SkillProperty(SKILLP_Respawn);
+
+		if (MinRespawnTics > 0)
+			respawn_monsters = MinRespawnTics;
+		else if (MinRespawnTics < 0)
+			respawn_monsters = -MinRespawnTics * 35;
```
the-phinet: Probably use TICRATE here

RageSpark: Forgot about that constant, update coming shortly!

src/playsim/p_mobj.cpp
```diff
@@ -403,6 +403,8 @@ void AActor::Serialize(FSerializer &arc)
 		("damagesource", damagesource)
 		("behaviors", Behaviors)
 		A("decalgenerator", DecalGenerator);
+		A("minrespawntics", MinRespawnTics);
+		A("respawndice", RespawnDice);
```
Boondorl, line 407: The semicolon after DecalGenerator needs to be removed and only the last statement should have one. By itself the `A()` macro doesn't do anything, it's just writing `("respawndice", RespawnDice, def->RespawnDice)`. It's relying on the fact that the `()` operator for `arc` can be chained which is being terminated early by the `;`

Boondorl (dismissed):

RicardoLuis0 (dismissed):

the-phinet (approved):

PR #1145, opened 2026-03-15
Update Keybinds

### LINKED ISSUE #1072: Left-Handed Keybind Preset
### Game

All

### Problem Description

> Funnily enough, as a left handed player, WASD really doesn't suit me at all and I actually found the move to not using the arrow keys as a default quite jarring. Even now, when I do a clean install and press the arrows keys to find them doing nothing, it still feels like something has broken. So, personally, I'd be quite happy if it wasn't the default. I get that people tend to use WASD, so I don't actually think it should be removed as a default, but it's a really awkward one for me. That's what options are for though, and the keybinds are one of the things I change immediately. So, for me WASD is not a good default, but I accept that it probably is for the majority. So, I, agree with you.
>
> Similarly (and I'm speaking more generally about all modern games - so, basically off-topic), keys near WASD being used for use, open, alt-fire, scan, crouch, hack, lock-pick, whatever are, perhaps, even more awkward - they are on the wrong side of the keyboard for me, lost somewhere in the middle of the rest of the keys, involve me taking my hand off the arrows, and are really hard to find in a tight situation. The number of games that do not allow rebinding of such keys is surprisingly high. At least full binding/rebinding of everything is possible in UZDoom.

### Feature Description

Add an optional keybind preset that is suitable for left-handed players. I'm not sure what this would look like, as I am not left-handed.

If some southpaw users could share their setups, it would be very appreciated. Examples of what commercial games offer would also be useful.

### Additional Media

_No response_

the-phinet:
Updates keybinds:

- adds reload bind by default
- adds weapon zoom to middle mouse click
- adds non-toggle crouch alternative bind
- **swaps jump and use on the numpad**
- adds left handed wasd-like binding, which works on tkl layouts down to hhkb
- makes arrow keys slightly more usable

This might make it so left-handed folk can use the game with 0 needed configuration. None of these conflict with current keybinds, afaict.

image

Keyboard picture stolen from here: https://kbs.im/
It's wild to me how many "visualize the mechanical keyboard you want to buy (with sound!)" websites exist. I couldn't find a single one where I could make a quick layout diagram on.

Binds inspired from:
- https://www.reddit.com/r/pcgaming/comments/17nqv1l/what_kind_of_keybinds_do_you_use_for_fps_games_as/
- https://linustechtips.com/topic/37074-key-bindings-for-left-handers/
- https://expertbeacon.com/how-do-left-handed-gamers-play/
- https://www.howtogeek.com/im-a-left-handed-gamer-heres-how-i-customize-my-pc-gaming-experience/
- calling my one left handed friend that plays counter strike

Fixes #1072 

## Checklist

### DIFF
--- a/src/common/console/c_bind.cpp
+++ b/src/common/console/c_bind.cpp
@@ -44,102 +44,95 @@ const char *KeyNames[NUM_KEYS] =
 	// We use the DirectInput codes and assume a qwerty keyboard layout.
 	// See <dinput.h> for the DIK_* codes
 
-	nullptr,	"Escape",	"1",		"2",		"3",		"4",		"5",		"6",		//00
-	"7",		"8",		"9",		"0",		"-",		"=",		"Backspace","Tab",		//08
-	"Q",		"W",		"E",		"R",		"T",		"Y",		"U",		"I",		//10
-	"O",		"P",		"[",		"]",		"Enter",	"Ctrl",		"A",		"S",	//18
-	"D",		"F",		"G",		"H",		"J",		"K",		"L",		";",		//20
-	"'",		"`",		"Shift",	"\\",		"Z",		"X",		"C",		"V",		//28
-	"B",		"N",		"M",		",",		".",		"/",		"RShift",	"KP*",		//30
-	"Alt",		"Space",	"CapsLock",	"F1",		"F2",		"F3",		"F4",		"F5",		//38
-	"F6",		"F7",		"F8",		"F9",		"F10",		"NumLock",	"Scroll",	"KP7",		//40
-	"KP8",		"KP9",		"KP-",		"KP4",		"KP5",		"KP6",		"KP+",		"KP1",		//48
-	"KP2",		"KP3",		"KP0",		"KP.",		nullptr,	nullptr,	"OEM102",	"F11",		//50
-	"F12",		nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	//58
-	nullptr,	nullptr,	nullptr,	nullptr,	"F13",		"F14",		"F15",		"F16",		//60
-	"F17",	  "F18",  	"F19",  	"F20",	  "F21",	  "F22",	  "F23",	  "F24",  	//68
-	"Kana",		nullptr,	nullptr,	"Abnt_C1",	nullptr,	nullptr,	nullptr,	nullptr,	//70
-	nullptr,	"Convert",	nullptr,	"NoConvert",nullptr,	"Yen",		"Abnt_C2",	nullptr,	//78
-	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	//80
-	nullptr,	nullptr,	nullptr,	nullptr,	nullptr,	"KP=",		nullptr,	nullptr,	//88
-	"Circumflex","@",		":",		"_",		"Kanji",	"Stop",		"Ax",		"Unlabeled",//90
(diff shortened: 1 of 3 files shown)


the-phinet:
@Enjay001, would you be willing to test if these binds are good enough for the average southpaw to be 90% happy with the defaults? I'm trying to minimize the amount of re-binding (or binding in general) that is needed.

Online I saw `IJKL`, `OKL;`, and `PL;'`. `OKL;` makes the most sense to me, since it's not on the bumped `j`, but next to it, like `WASD` is next the bumped `f`. I do notice I need to stretch my pinky a bit far to reach `shift`. Would using `PL;'` (or `'` intead of `shift`) make more sense?

mc776:
Shouldn't down on the numpad be 2?

RicardoLuis0:
> Shouldn't down on the numpad be 2?

i guess that's the "down" arrow on numpad, but it's really awkward to use because of the positioning compared to 5

the-phinet:
> Shouldn't down on the numpad be 2?

Based on the legends that would makes sense. I didn't want to change more then required. If the binds were slit into "right handed", "left handed", "keypad" sets, I think changing it would make sense. If Enjay (or another lefty that uses the keypad) thinks it should default to 2, I'm happy to change it.

mc776:
I don't use the numpad for FPS gaming so I'll let someone else comment, but I did use it all the time for platformers and having down be 5 instead of 2 would have badly thrown me off...

nashmuhandes (dismissed):

Enjay001:
> @Enjay001, would you be willing to test if these binds are good enough for the average southpaw to be 90% happy with the defaults? I'm trying to minimize the amount of re-binding (or binding in general) that is needed.
> 
> Online I saw `IJKL`, `OKL;`, and `PL;'`. `OKL;` makes the most sense to me, since it's not on the bumped `j`, but next to it, like `WASD` is next the bumped `f`. I do notice I need to stretch my pinky a bit far to reach `shift`. Would using `PL;'` (or `'` intead of `shift`) make more sense?

Oh, sorry, I totally missed this somehow (possibly because I was quite ill a couple of weeks ago).

I'm not sure I can speak for the average southpaw, but anything that shifts the keys across to the right side of the keyboard is an improvement.  I understand the attraction of having something in the "body" of the main keys (proximity of other keys for other minor functions), but one of the reason that I like the arrows is because they have no keys immediately around them (making accidental hitting of other things less likely, and making the arrows easy to find).  Something on the numpad would be my second choice (and the layout in your diagram is sensible enough).

Of the possibilities on the main keyboard that you listed, `PL;'` feels the most comfortable to me, mostly because it is further across to the right, and it also puts the nice big right-shift key in easy reach of the pinkie finger (in fact, my pinkie naturally rests right on it (on my keyboard at least - a pretty old UK layout Logitech G11).

The other options you mentioned `IJKL` in particular, might be towards the right-side of the main key area but on a keyboard with a number pad, it's still fairly central overall, requiring the right hand to move slightly too far left to be comfortable (IMO), or requiring the user to push their keyboard to the right.  At least, that's how it felt to me.

Enjay001:
> Shouldn't down on the numpad be 2?

By what's written on the keys, yes, absolutely - but by feel, 5 is actually more comfortable and seems to be more common in games that offer default bindings on the numpad. (Even Eye of the Beholder in 1991 had backwards bound to numpad5 (but also to numpad2 (and arrow down)).
I have also seen layouts more like the attached in some games:
Image1


So, the although 5 is not marked with a down arrow, and 2 is, 5 is commonly used for "backwards" and is in a pretty comfortable position for that (better than 2).

the-phinet:
Shifted the left binds to `PL;'`, as per Enjay's recommendation

TehRealSalt (approved):

PR #1149, opened 2026-03-16
Set SHARE_DIR to a subdirectory of CMAKE_INSTALL_PREFIX by default

fpiesche:
## Checklist

Currently on Linux, the `uzdoom.pk3` (and other supplementary files) is installed to a subdirectory of the `CMAKE_INSTALL_PREFIX`, which is correct behaviour. However, the default lookup paths added to a freshly-generated `uzdoom.ini` only include the hardcoded `DEFAULT_SHARE_DIR` of `/usr/local/share`.

As such, if the `CMAKE_INSTALL_PREFIX` is not `/usr/`, the correct path will not be added to a new default `uzdoom.ini` and uzdoom will fail to start as it isn't checking the correct path for `uzdoom.pk3`.

This is e.g. the case when building uzdoom into a Flatpak, where the `CMAKE_INSTALL_PREFIX` is `/app/` - so `uzdoom.pk3` correctly ends up in `/app/share/games/uzdoom/`, but a newly-generated config tries to look for it in `/usr/local/share/games/uzdoom/`.

This fixes this by simply setting the `SHARE_PATH` macro to the `share` directory below the `CMAKE_INSTALL_PREFIX`.

The only thing I am unsure about here is how this will behave on non-Linux Unixes. It *should* be fine since it just uses a standard CMake path and will leave `SHARE_DIR` untouched if not compiling via CMake but I don't really have much of a way to test it.

This will retain the default hardcoded `SHARE_DIR` of `/usr/local/share/` when generating new `uzdoom.ini` files but also add the one inside the `CMAKE_INSTALL_PREFIX` to the defaults. Fully replacing this with the CMake prefix would be cleaner; however doing so would need some more changes to how the default lookup paths are generated in `src/gameconfigfile.cpp` which currently forces `/usr/local/share/` to be included via the `DEFAULT_SHARE_DIR` macro.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1310,6 +1310,13 @@ if(BUILD_FLATPAK)
 	target_compile_definitions(zdoom PRIVATE IS_FLATPAK=1)
 endif()
 
+# Set SHARE_DIR to a subdirectory of the cmake install prefix, rather than hardcoding
+if(UNIX AND NOT APPLE)
+  set(SHARE_DIR "\"${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DATAROOTDIR}\"")
+  target_compile_definitions(zdoom PRIVATE SHARE_DIR=${SHARE_DIR})
+  message(STATUS "Set SHARE_DIR to ${SHARE_DIR}")
+endif()
+
 if(${CMAKE_SYSTEM_NAME} STREQUAL "SunOS")
 	# [BL] Solaris requires these to be explicitly linked.
 	set( PROJECT_LIBRARIES ${PROJECT_LIBRARIES} nsl socket)


the-phinet:
I don't think this approach will work. Since `SHARE_DIR` is always defined now, the haiku code path will never trigger. If this were to be added, the stuff in `i_system.h` should be removed, and haiku should be accounted for.

Would it not be not be better to instead of `if(UNIX AND NOT APPLE)` doing `if(BUILD_FLATPAK)`. That way flatpaks are corrected, and if the user wants to override `SHARE_DIR`, they still have that option through build params (`-DSHARE_DIR=whatever`)?

fpiesche:
Would `-DSHARE_DIR` actually work here? It's a C++ macro and not a cmake build option. I could swear I'd tried `-DSHARE_DIR` and it didn't do anything in the Flatpak build but will try again.

However: if it doesn't do anything as such, I could just expose the macro for overriding via cmake as an option, which would solve the issue - leave it unset by default, but allow overriding it if installing to a custom path...

the-phinet:
> Would `-DSHARE_DIR` actually work here? It's a C++ macro and not a cmake build option. I could swear I'd tried `-DSHARE_DIR` and it didn't do anything in the Flatpak build but will try again.

Sorry, I really should check what I send before I send it. The correct thing would be `-DCMAKE_CXX_FLAGS=-DSHARE_DIR='\"whatever\"'`

fpiesche:
Here I was trying to figure out how to set a C++ macro with cmake and could only find the `target_compile_definitions` path. Shows how much I know about C. :rofl:

PR #1153, opened 2026-03-20
Updated metainfo

the-phinet:
- Add some backlinks for use on flathub
- Reordered some keys to group things a bit better
- Fix some minor capitalization issues

Christ on a bike, flathub's metainfo checker is annoyingly pedantic and also doesn't follow the freedesktop spec

## Checklist

### DIFF
--- a/src/common/platform/win32/i_crash.cpp
+++ b/src/common/platform/win32/i_crash.cpp
@@ -1788,7 +1788,7 @@ static INT_PTR CALLBACK OverviewDlgProc (HWND hDlg, UINT message, WPARAM wParam,
 		SetWindowTextW(edit, L"Please tell us about this problem.\n"
 			"The information will NOT be sent to Microsoft.\n\n"
 			"An error report has been created that you can submit to help improve " GAMENAME ". "
-			"You can either save it to disk and make a report in the bugs forum at " FORUM_URL ", "
+			"You can either save it to disk and make a report in the issue tracker at " BUGS_URL ", "
 			"or you can send it directly without letting other people know about it.");
 		SendMessageW(edit, EM_SETSEL, 0, 81);
 		SendMessageW(edit, EM_SETCHARFORMAT, SCF_SELECTION, (LPARAM)&charFormat);
@@ -1814,7 +1814,7 @@ static INT_PTR CALLBACK OverviewDlgProc (HWND hDlg, UINT message, WPARAM wParam,
 		{
 			if (link->msg == WM_LBUTTONDOWN)
 			{
-				ShellExecuteA (NULL, "open", BUGS_FORUM_URL, NULL, NULL, 0);
+				ShellExecuteA (NULL, "open", BUGS_URL, NULL, NULL, 0);
 				SetWindowLongPtrW (hDlg, DWLP_MSGRESULT, 1);
 				return TRUE;
 			}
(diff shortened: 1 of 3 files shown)


RicardoLuis0:
feels like FAQ should still be https://zdoom.org/about, if that page needs any updating we can just update it

the-phinet:
> feels like FAQ should still be https://zdoom.org/about, if that page needs any updating we can just update it

Sure, but I'd like to have a link to the gh wiki somewhere on the flathub store page. Also with the connection issues on the website's side, I'm unsure if sending even more people that way is a great idea.

It's for these links at the bottom of the page:
image

Here's a list of available types (unfortunately no duplicates allowed): https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-url

the-phinet:
Actually, reading through the about page on the zdoom.org, it's not an FAQ at all. It's a feature list. Unless we want to add faq stuff there, I think moving to the github wiki is the right move.

RicardoLuis0:
> Actually, reading through the about page on the zdoom.org, it's not an FAQ at all. It's a feature list. Unless we want to add faq stuff there, I think moving to the github wiki is the right move.

that's what i mean yeah, seems that page hasn't seen any updates in quite a while, so making it into a proper faq (or adding a faq section to it), and possibly adding the wiki link there might be good

the-phinet:
I still don't think that's a good idea due to the connection issues. ZDoom's about page is would need such an overhaul that it feels like destroying history. The left hand pages on the wiki's front page (The stuff under "Using ZDoom" and "Room 101") are outdated to the point of being useless and contain a wild amount of irrelevant information (and deadnaming).

We're already using the github wiki for the "how to compile UZDoom" stuff, using it for the "how to use UZDoom" stuff makes so much sense.

RicardoLuis0:
the pointed page is just full of TODOs anyhow, so i really don't see how it's better, but fair 😅

the-phinet:
There are now significantly less TODOs lol. Hopefully should be good to go?

RicardoLuis0:
ah, just noticed

RicardoLuis0:
the bug report link in version.h is still pointing to the forums, if that could be updated too, it'd be great (otherwise, it's all good)

src/common/platform/win32/i_crash.cpp
```diff
@@ -1788,7 +1788,7 @@ static INT_PTR CALLBACK OverviewDlgProc (HWND hDlg, UINT message, WPARAM wParam,
 		SetWindowTextW(edit, L"Please tell us about this problem.\n"
 			"The information will NOT be sent to Microsoft.\n\n"
 			"An error report has been created that you can submit to help improve " GAMENAME ". "
-			"You can either save it to disk and make a report in the bugs forum at " FORUM_URL ", "
+			"You can either save it to disk and make a report in the bugs forum at " BUGS_URL ", "
```
RicardoLuis0: just an even more minor nitpick but the text should probably be changed from 'bugs forum' to 'issue tracker' here

RicardoLuis0 (dismissed):

src/common/platform/win32/i_crash.cpp
the-phinet: On it! I did notice this was the only place the url was used. Should we add it to the crash stuff on other platforms?

RicardoLuis0: probably yeah

RicardoLuis0 (approved):

Issue #1080, opened 2026-02-27
Instant crash entering a map in Hideous Destructor


mc776:
### UZDoom Version

4.14pre-1227-gba14a60e

### Game

All

### Operating System

Linux x86_64

### OS Version

Debian Trixie

### Hardware

Nvidia

### Duplicate Confirmation

- [x] I have searched and not found a similar issue.

### Bug Description

Last working CI: https://github.com/UZDoom/UZDoom/commit/1fedfa60de9b280238b548769b682e600bfda1c1

Issue posted on HD: https://codeberg.org/mc776/HideousDestructor/issues/379

### Replication

1. Download HD https://codeberg.org/mc776/HideousDestructor/ and run it
2. Warp or start a new game
3. Expected: enter map with chosen loadout
4. Actual: CTD right after the first gametic

### Configuration

(already tried with config renamed)

### Log

in the terminal:

```string expected for 'name'
'Attempt to load pointer to inexisted CVar '''
 while restoring HDP_L6


*** Fatal Error ***
Address not mapped to object (signal 11)
Address: 0x561e4e470bf0

Generating uzdoom-crash.log and killing process 4183042, please wait... sh: 1: gdb: not found
Killed
```

saved crashlog:
```
*** Fatal Error ***
Address not mapped to object (signal 11)
Address: 0x55798529bd10

System: Linux Frankenbeast-1 6.12.63+deb13-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.12.63-1 (2025-12-30) x86_64 GNU/Linux

UZDoom version g4.14pre-1239-g01ab1b22 (01ab1b22)
Command line: /tmp/.mount_Linux-8UJlbX/./usr/bin/uzdoom -nomusic -noautoload -iwad /home/m/doom/LotansTomb/wads/lt2.wad -file .
Wad 0: uzdoom.pk3
Wad 1: game_support.pk3
Wad 2: lt2.wad
Wad 3: game_widescreen_gfx.pk3
Wad 4: /home/m/doom/hd/_git/

Current map: RANGE

viewx = -29.000000
viewy = 32.000000
viewz = 49.680000
viewangle = 0.000000

Executing: gdb --quiet --batch --command=gdb-respfile-3qCTRB
```

the-phinet:
[...]

PR #1154, opened 2026-03-21
Partially revert ba14a60

mc776:
Addresses #1080

## Checklist


Boondorl:
I think @Gutawer would have better odds reviewing this one because this seems to be related to the optimization she made in the serializer and I can't quite say why it was done the way it was

RicardoLuis0:
#1167 should fix the issue without needing the removal of the new optimization

PR #1156, opened 2026-03-22
PP Expansion - Weapon & UI

TheIndiv:
Adds "weapon" and "ui" postprocess targets to GLDEFS. Weapon sprites render into an isolated Rgba16f texture, run through any enabled weapon shaders, then get alpha-composited onto the scene. The in-game HUD does the same through the ui chain, compositing onto the fully-rendered scene afterward. Effects can bleed past element boundaries and interact across layers.

Menus and the console bypass the ui pipeline entirely.

All GLDEFS shaders now get six auto-declared samplers: InputTexture, LastInputTexture, WeaponTexture, WeaponLastTexture, UITexture, and UILastTexture. WeaponTexture No overhead when neither weapon nor ui target is in use.

Implemented for both GL and Vulkan.

- [✔] I have reviewed [CONTRIBUTING.md] and agree to its terms.

### DIFF
--- a/src/common/rendering/gl/gl_framebuffer.cpp
+++ b/src/common/rendering/gl/gl_framebuffer.cpp
@@ -49,6 +49,7 @@
 
 #include "flatvertices.h"
 #include "hw_cvars.h"
+#include "hwrenderer/postprocessing/hw_postprocess.h"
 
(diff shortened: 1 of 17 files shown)


RicardoLuis0:
will need to wait until after the render refactor for merging

PR #1158, opened 2026-03-23
fix: guard against null wiper and out-of-range wipetype

TheIndiv:
PerformWipe would crash with a null deref if Wiper::Create returned null
(e.g. an unrecognised wipe type). Add an early return that cleans up and
bails instead.

wipetype was a plain CVAR with no bounds check, so a stale value of 4
(wipe_NUMWIPES) persisted from old configs and triggered the above crash
on the first level transition. Changed to CUSTOM_CVAR with a validator
that resets any out-of-range value to wipe_Melt on load.

- [✔] I have reviewed [CONTRIBUTING.md](../blob/trunk/CONTRIBUTING.md) and agree to its terms.

### DIFF
--- a/src/common/2d/wipe.cpp
+++ b/src/common/2d/wipe.cpp
@@ -517,6 +517,12 @@ void PerformWipe(FTexture* startimg, FTexture* endimg, int wipe_type, bool stops
 	auto starttex = MakeGameTexture(startimg, nullptr, ETextureType::SWCanvas);
 	auto endtex = MakeGameTexture(endimg, nullptr, ETextureType::SWCanvas);
 	auto wiper = Wiper::Create(wipe_type);
+	if (!wiper)
+	{
+		I_FreezeTime(false);
+		GSnd->SetSfxPaused(false, 1);
+		return;
+	}
 	wiper->SetTextures(starttex, endtex);
 
 	wipestart = I_msTime();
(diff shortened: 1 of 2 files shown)


src/common/2d/wipe.cpp
```diff
@@ -517,6 +517,12 @@ void PerformWipe(FTexture* startimg, FTexture* endimg, int wipe_type, bool stops
 	auto starttex = MakeGameTexture(startimg, nullptr, ETextureType::SWCanvas);
 	auto endtex = MakeGameTexture(endimg, nullptr, ETextureType::SWCanvas);
 	auto wiper = Wiper::Create(wipe_type);
+	if (!wiper)
+	{
+		I_FreezeTime(false);
+		GSnd->SetSfxPaused(false, 1);
+		return;
+	}
```
RicardoLuis0: it's probably better to just move those calls to after the `!wiper` check instead of undoing them, because they may have side-effects

RicardoLuis0 (changes requested):

src/common/2d/wipe.cpp
TheIndiv: Updated wipe.cpp please take a look!

RicardoLuis0 (approved):

PR #1178, opened 2026-03-26
Add colors parameter to NameKeys

DyNaM1Kk:
## Checklist

### DIFF
--- a/wadsrc/static/zscript/engine/ui/menu/menu.zs
+++ b/wadsrc/static/zscript/engine/ui/menu/menu.zs
@@ -23,7 +23,7 @@
 
 struct KeyBindings native version("2.4")
 {
-	native static String NameKeys(int k1, int k2);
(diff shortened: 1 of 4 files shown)


wadsrc/static/zscript/engine/ui/menu/menu.zs
```diff
@@ -23,7 +23,7 @@
 
 struct KeyBindings native version("2.4")
 {
-	native static String NameKeys(int k1, int k2);
+	native static String NameKeys(int k1, int k2, bool colors = true);
```
the-phinet, line 26: Is this supposed to be different than the default on the native function?

DyNaM1Kk: Oh I didn't even notice that, but that's the way it is with `NameAllkeys`, too.

RicardoLuis0 (approved):

PR #1182, opened 2026-03-27
Added a self-reference check to P_CheckSight

### LINKED ISSUE #1147: CheckSight crashes when called to/from actors crossing through stacked portals
### UZDoom Version

UZDoom 5.0.0-pre-80-gfd3fcf897

### Game

All

### Operating System

Windows 11

### OS Version

Windows 11 Pro 10.0.26100 Build 26100

### Hardware

13th Gen Intel(R) Core(TM) i9-13900HX, GeForce RTX 4060 Ti, 32GB memory

### Duplicate Confirmation

### Bug Description

Calling CheckSight will crash UZDoom if the calling actor is self-referencing while passing through a stacked portal (eg., `self.CheckSight(self)`). This is almost certainly related to previous issues with vertical portals (see #1037 #1051).

### Replication

See attached .zip. The custom TestDoomPlayer actor will constantly run CheckSight on itself. Console messages should show that TestDoomPlayer is constantly checking its sightline to TestDoomPlayer.

As soon as the player starts to enter the stacked portal (by moving up the steps), UZDoom will freeze. This issue is not specific to player actors, but this was the easiest way to demonstrate it.

[StackedSectorCheckSightBug.zip](https://github.com/user-attachments/files/26010727/StackedSectorCheckSightBug.zip)

theshadsy:
If an actor calls CheckSight on themselves and they can see themselves, return true. Fixes #1147.

This is meant as a fix for a crash that occurs when an actor self-references with CheckSight while moving through a stacked portal. Boondorl suggested in Discord this should be the standard behavior for CheckSight anyway; there are no circumstances where a player is visible to themselves and a self-referenced CheckSight should return false.

~~Untested, but works in theory!~~ [See comment below](https://github.com/UZDoom/UZDoom/pull/1182#issuecomment-4140135490); these commits have been tested and they resolve the crash in #1147 without impacting other uses of CheckSight.

(If this does get added, I'd recommend squashing to one commit, since the first one had an error!)

## Checklist

### DIFF
--- a/src/playsim/p_sight.cpp
+++ b/src/playsim/p_sight.cpp
@@ -881,6 +881,14 @@ sightcounts[0]++;
 		}
 	}
 
+	// Return true if the actor is calling CheckSight
+	// on themselves and they can see themselves
+	if (t1 == t2)
+	{
+		res = true;
+		goto done;
+	}
+	
 	// killough 4/19/98: make fake floors and ceilings block monster view
 
 	if (!(flags & SF_IGNOREWATERBOUNDARY))
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -266,3 +266,4 @@ Francisco Ortega
 Adam Pliszka
 Vortex (vortexdesign)
 RageSpark
+Phil "Shadsy" Salvador


src/playsim/p_sight.cpp
```diff
@@ -881,6 +881,14 @@ sightcounts[0]++;
 		}
 	}
 
+	// Return true if the actor is calling CheckSight
+	// on themselves and they can see themselves
+	if (t1 == t2)
+	{
+		res = true
```
RicardoLuis0: wouldn't it be best to return false? returning true may lead to some functions immediately returning self instead of searching for further actors

theshadsy: 1. Oh my god I forgot to add a semicolon, I'm a doofus. See the next commit.
2. I think this is the right behavior? P_CheckSight checks if actor t1 can see actor t2. If they're the same actor, it should always return true since they can always see themselves, right? There's no iteration through actors in this script, it's just looking for a straight line from t1 to t2.

theshadsy: Okay I see, I think you might be confusing this with A_CheckSight or CheckIfSeen, which iterate through camera actors. This is the underlying function behind those actions; it just checks for a LOS between two specific actors, so that shouldn't be an issue.

RicardoLuis0: well yes, but also for things like aiming with AimTraverse/etc, those functions don't have a self check -- if P_CheckSight reports that something is 'visible' they'll just try and work with that information

RicardoLuis0: and an actor is just fundamentally not visible from itself

theshadsy:
Whoops, apologies for breaking protocol on untested pull requests. I grabbed the artifact build, and I can confirm that this fixes the crash identified in #1147, without impacting the other uses of CheckSight. Will check _before_ making a pull request in the future!

src/playsim/p_sight.cpp
theshadsy: The behavior in UZDoom right now is that actors _can_ see themselves when using CheckSight, so that is not being changed here (and probably shouldn't be at the risk of breaking backcompat). This just skips over the LOS/portal-testing steps and immediately assumes that, if an actor is visible to itself, it can see itself.

RicardoLuis0: > The behavior in UZDoom right now is that actors _can_ see themselves when using CheckSight, so that is not being changed here (and probably shouldn't be at the risk of breaking backcompat). This just skips over the LOS/portal-testing steps and immediately assumes that, if an actor is visible to itself, it can see itself.

ah, i didn't realize that, fair

RicardoLuis0 (dismissed):

src/playsim/p_sight.cpp
theshadsy: Oh yeah, not making a judgment one way or another on whether that's the right call or makes sense! That's a discussion outside my grade hahaha.

Boondorl:
I'd still like to double check what exactly is crashing here in case it's something that needs to be fixed, but this check should be added regardless to avoid any possible pointless busy work as naturally something will always be able to see itself

the-phinet:
Unless you're still modifying the code, there's no need to pull in changes. We'll do that when we merge it into the codebase. (also use base instead of merge, if you do need to pull in changes)

Also also, Can you add yourself to the bottom of the CONTRIBUTORS file?

theshadsy:
> Unless you're still modifying the code, there's no need to pull in changes. We'll do that when we merge it into the codebase. (also use base instead of merge, if you do need to pull in changes)
> 
> Also also, Can you add yourself to the bottom of the CONTRIBUTORS file?

Sorry, still new to this whole process! (I did that on my other pull request too.)

Will add myself to the list!

RicardoLuis0 (approved):
ah, just realized that the CONTRIBUTORS commit dismissed the review xD

PR #1186, opened 2026-03-27
Fix Tooltip Scroll Speed being the other way around

### LINKED ISSUE #862: Tooltip Scroll Speed wrong way around
### UZDoom Version

UZDoom Begin 321

### Game

All

### Operating System

Windows 11

### OS Version

N/A

### Hardware

N/A

### Duplicate Confirmation

### Bug Description

Tooltip scroll speed has the range 1 - 5. To me, 5 should be fastest, and 1 should be slowest.
Currently the reverse is happening. 5 is slowest and 1 is fastest.

### Replication

_No response_

### Configuration

```ini

```

### Log

_No response_

RageSpark:
Fixes #862

## Checklist

### DIFF
--- a/src/common/menu/menu.cpp
+++ b/src/common/menu/menu.cpp
@@ -72,7 +72,7 @@ CUSTOM_CVAR(Float, m_tooltip_delay, 9.0f, CVAR_ARCHIVE | CVAR_GLOBALCONFIG)
 	if (self <= 0.0f)
 		self = 0.1f;
 }
-CUSTOM_CVAR(Float, m_tooltip_speed, 3.0f, CVAR_ARCHIVE | CVAR_GLOBALCONFIG)
+CUSTOM_CVAR(Float, m_tooltip_speed, 0.5f, CVAR_ARCHIVE | CVAR_GLOBALCONFIG)
 {
 	if (self <= 0.0f)
 		self = 0.1f;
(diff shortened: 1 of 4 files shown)


Boondorl:
I'm still not entirely sure how I feel about this, the main reason being the value is supposed to be the number of seconds per line rather than the number of lines per second. IMO it's much more confusing to work between 0 and 1 than it is whole numbers, especially because this won't scale linearly which the slider forces, and perhaps the description just needs further clarifying

Edit: To further expound on the "lines per second" thing, to get the same whole numbers we have now, it'd go 1.0 -> 0.5 -> 0.3333 -> 0.25 -> 0.2 -> 0.16666 -> etc. which naturally can't be captured correctly with a set value. The result is 0.5 - 1.0 will only see marginal changes in line speed and it'll then start to skyrocket with the only real solution being the slider having extremely small increments, something I'm against because it makes interacting with it on keyboard/controller painful

RageSpark:
> I'm still not entirely sure how I feel about this, the main reason being the value is supposed to be the number of seconds per line rather than the number of lines per second. IMO it's much more confusing to work between 0 and 1 than it is whole numbers, especially because this won't scale linearly which the slider forces, and perhaps the description just needs further clarifying
> 
> Edit: To further expound on the "lines per second" thing, to get the same whole numbers we have now, it'd go 1.0 -> 0.5 -> 0.3333 -> 0.25 -> 0.2 -> 0.16666 -> etc. which naturally can't be captured correctly with a set value. The result is 0.5 - 1.0 will only see marginal changes in line speed and it'll then start to skyrocket with the only real solution being the slider having extremely small increments, something I'm against because it makes interacting with it on keyboard/controller painful

I see! Makes sense now, thanks for the explanation, should I close this?

PR #1187, opened 2026-03-27
Correctly check Player's inventory if it's a Voodoo Doll activating a locked Line Special

### LINKED ISSUE #301: Voodoo dolls don't check players for keys
Windows 10

In [Junkfood 2](https://www.doomworld.com/forum/topic/138114-junkfood-2-slaughter-boom/) on MAP65, there's a voodoo doll that repeatedly runs over a line which triggers something if it has three keys. However, UZDoom apparently does not check the actual player, only the voodoo doll's inventory, so it cannot actually trigger, and progression is blocked due to this.

RageSpark:
Fixes #301

## Checklist

### DIFF
--- a/src/gamedata/a_keys.cpp
+++ b/src/gamedata/a_keys.cpp
@@ -113,6 +113,10 @@ struct Lock
 
 	bool check(AActor * owner)
 	{
+		// Check if it's a Voodoo doll, if so, grabs the actual player.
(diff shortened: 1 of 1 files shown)


MajorCooke:
Thanks buddy!

Boondorl (approved):

PR #1191, opened 2026-03-28
Emulate Mikoportals Compatibility flag

### LINKED ISSUE #642: Mikoportals physics incompatibility with vanilla/boom/mbf21
### UZDoom Version

UZDoom 4.14.3

### Game

Doom

### Operating System

Windows 11

### OS Version

_No response_

### Hardware

_No response_

### Duplicate Confirmation

### Bug Description

Mikoportal is a behaviour compatible with vanilla, boom and mbf21 described here: https://doomwiki.org/wiki/Mikoportal
This behaviour relies on a Z-coordinate underflow and allows to create, for example, vanilla conveyors, because this results in essentially frictionless movement.

Here is a minimal reproducer wad:
[mikoportal.zip](https://github.com/user-attachments/files/23840203/mikoportal.zip)

The crusher blows up the barrel that pushes the voodoo doll into a mikoportal setup floor. In dsda-doom with -complevel 2 or 9 or 21 the voodoo doll will move without friction in it's closet and after about 8 seconds will cross the W1 exit line.
In UZDoom the voodoo doll will stop shortly after being initially pushed.

In many cases that I know to work around that in zdoom mappers had to create a TERRAIN lump and use special flat for vanilla conveyors that will be marked as having no friction.
One such example is in [NERF](https://www.doomworld.com/idgames/levels/doom2/Ports/megawads/nerf).
```
//ZDoom Mikoportals fix
//Frictionless floor to fix behaviour of Mikoportals in ZDoom based ports
terrain mikoportal
{
	friction 255
}

floor MIKOPORT mikoportal
```

Another effect based on a similar mechanic is a snap-to-ceiling effect used in a few maps. The easiest one to demonstrate is in [revenant hallway](https://www.doomworld.com/idgames/levels/doom2/Ports/megawads/revhallway) map 25. In dsda-doom with -complevel 9 the player will snap to the ceiling after passing the opening corridor and be able to proceed to the map. In UZDoom you cannot leave that opening corridor because you don't get snapped to the ceiling.

However, workaround above doesn't really help in this case because the lack of friction is not the only thing that is relied upon here.

I assume since UZDoom stores coordinates differently from vanilla internally this has to be specifically emulated to make it work.

### Replication

_No response_

### Configuration

```ini

```

### Log

_No response_

RageSpark:
Fixes #642, unsure if this is the best way but tested and fully working with the examples at the mentioned issue.

May require a bit more testing in some maps and a better name?

## Checklist


RicardoLuis0 (dismissed):

Gutawer:
There's probably some reason this hasn't been done but I'd like to understand it - I'd expect this to be implemented by adding some code after this line: https://github.com/UZDoom/UZDoom/blob/trunk/src/playsim/p_mobj.cpp#L2951

Effectively the original bug here is a fixed-point overflow so I'd expect we could emulate it directly by adding something like `Z = positive_fmod(Z + 32768.0, 65536.0) - 32768.0` after the velocity is added to the Z position. This would preserve some of the finer details of Mikoportals, for instance the fact that it actually *is* possible to land on one by running off a floor at -32766 - the engine will apply 2 units of gravity on the first tick, and then the post-addition position will actually be precisely -32768, which is not *quite* enough to trigger the underflow.

And my feeling is that doing it like this would also mean not needing to change XY movement code as the engine would just never see the actor on the ground. Though some care would still need to be taken to avoid interpolation on the tick.

RageSpark:
> There's probably some reason this hasn't been done but I'd like to understand it - I'd expect this to be implemented by adding some code after this line: https://github.com/UZDoom/UZDoom/blob/trunk/src/playsim/p_mobj.cpp#L2951
> 
> Effectively the original bug here is a fixed-point overflow so I'd expect we could emulate it directly by adding something like `Z = positive_fmod(Z + 32768.0, 65536.0) - 32768.0` after the velocity is added to the Z position. This would preserve some of the finer details of Mikoportals, for instance the fact that it actually _is_ possible to land on one by running off a floor at -32766 - the engine will apply 2 units of gravity on the first tick, and then the post-addition position will actually be precisely -32768, which is not _quite_ enough to trigger the underflow.
> 
> And my feeling is that doing it like this would also mean not needing to change XY movement code as the engine would just never see the actor on the ground. Though some care would still need to be taken to avoid interpolation on the tick.

Just by adding a simple Z < -32768 after AddZ seems to works perfectly, should have tried this sooner. Tried your code and it also seems to work perfectly (but I am not quite knowledge on fmod)

RicardoLuis0 (approved):

Gutawer:
Seems good now. I think I prefer what you've written to the `fmod` version honestly - it's a bit less strictly "accurate" but should account for 100% of real-world use cases especially since it's probably impossible in a vanilla map to trigger a Mikoportal while moving _upwards_ (since you'd have to clear the actor's height in a single tick).

leahanderson1 (approved):
+1

PR #1202, opened 2026-03-31
Limit translation warning length

### LINKED ISSUE #1200: Developer mode 2+ causes a crash with Blade of Agony (and presumably some other things)
### UZDoom Version

5.0.0-pre-145-g08344e151d

### Game

All

### Operating System

Windows 10

### OS Version

_No response_

### Hardware

_No response_

### Duplicate Confirmation

### Bug Description

If you are playing Blade of Agony, the game will crash if develper mode is set to 2, 3 or 4 (0 (default) and 1 seem OK).

The game can start in as much as it will boot up, but if you try to start a game from the menu with developer 2 already active, the game will crash as you enter the map.  If you start a game with developer mode set to 0, go to a map and then type `developer 2` at the console, the game will crash.

While I have mostly just tried this by simply starting a game from the in-game menu, it is also possible to get the crash when going to different maps via the `map` console command or adding `+map [insert map name here]` to the command line.  i.e. it seems to affect most, if not all, maps in the game.

I normally have developer mode set to 2 but I have only noticed this with BoA.

I have tried this with an otherwise totally default ini.  Simply changing the developer cvar and nothing else is enough to get the crash.

### Replication

Use the build of BoA provided by Hawkwind in #1173 
https://github.com/Realm667/WolfenDoom/archive/refs/heads/master.zip
That's the entire project and it can be renamed to BoA.ipk3 and placed in the UZDoom folder (or another IWAD search location) to get it to work.

**With Developer mode already at 2**
1. With developer mode already set to 2 (or higher) in your ini, start UZDoom and pick Blade of Agony as your IWAD.
2. When the menu appears, start a game
3. As the first map loads, the game will crash

[CrashReport1.zip](https://github.com/user-attachments/files/26347821/CrashReport1.zip)

**With developer mode at 0**
1. With developer mode at the default of 0 in your ini, start UZDoom and pick Blade of Agony as your IWAD.
2. When the menu appears, start a game
3. After the map has started (an intro cutscene) open the console and type `developer 2` and the game will crash

[CrashReport2.zip](https://github.com/user-attachments/files/26347866/CrashReport2.zip)

### Configuration

```ini
developer=2
```

### Log

_No response_

the-phinet:
Closes #1200 

Does not fix the actual issue (I'll open an issue in BOA's github soon), but it does prevent a crash

## Checklist

### DIFF
--- a/src/common/audio/sound/oalsound.cpp
+++ b/src/common/audio/sound/oalsound.cpp
@@ -1116,7 +1116,7 @@ SoundHandle OpenALSoundRenderer::LoadSoundRaw(uint8_t *sfxdata, int length, int 
 	{
 		static bool warned = false;
 		if(!warned)
-			Printf(DMSG_WARNING, "Loop points not supported!\n");
+			DPrintf(DMSG_WARNING, "Loop points not supported!\n");
 		warned = true;
 	}
 
--- a/src/common/engine/m_haptics.cpp
+++ b/src/common/engine/m_haptics.cpp
@@ -249,7 +249,7 @@ const FName * Joy_GetMapping(const FName identifier)
 	{
 		if (RumbleMissed.Contains(identifier)) return;
 		RumbleMissed.Push(identifier);
-		Printf(PRINT_HIGH|PRINT_NONOTIFY, "Unknown rumble mapping '%s'\n", identifier.GetChars());
+		Printf(PRINT_NONOTIFY, "Unknown rumble mapping '%s'\n", identifier.GetChars());
 	};
 
 	if (!mapping && identifier != "")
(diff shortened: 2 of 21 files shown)


src/common/audio/sound/oalsound.cpp
```diff
@@ -1116,7 +1116,7 @@ SoundHandle OpenALSoundRenderer::LoadSoundRaw(uint8_t *sfxdata, int length, int
 	{
 		static bool warned = false;
 		if(!warned)
-			Printf(DMSG_WARNING, "Loop points not supported!\n");
+			Printf(PRINT_HIGH, "Loop points not supported!\n");
```
RicardoLuis0: this should probably be turned into a DPrintf, that seems to have been the original intention

src/common/engine/m_haptics.cpp
```diff
@@ -299,7 +299,7 @@ const struct Haptics * Joy_GetRumble(FName identifier)
 	if (!rumble && !RumbleMissed.Contains(identifier))
 	{
 		RumbleMissed.Push(identifier);
-		Printf(DMSG_ERROR, TEXTCOLOR_RED "Rumble mapping not found! '%s'\n", identifier.GetChars());
+		Printf(PRINT_MEDIUM, TEXTCOLOR_RED "Rumble mapping not found! '%s'\n", identifier.GetChars());
```
RicardoLuis0, line 302: same for this? (and the others on this file)

the-phinet: Maybe... I'll fix this one when I mute the haptics stuff later this week

RicardoLuis0 (approved):

PR #1204, opened 2026-03-31
Use DMG for packaging macOS builds

### LINKED ISSUE #599: Package MacOS builds as DMG
This will allow for cleaner downloads, and a more user-friendly "install" experience

Basic steps: https://apple.stackexchange.com/questions/404954/how-to-create-drag-to-applications-folder-in-dmg

But this needs to be done from ci somehow

leahanderson1:
## Checklist

this PR is very WIP, as of right now, it packages the .dmg file into a zip, which is unnecessary. The outputted DMG should also be observed to see if it needs any changes to the built-in config file. As far as I can see, the example file is licensed under MIT, but we [should check in with them](https://github.com/dmgbuild/dmgbuild/issues) just in case. we should also change the license to GPL3 in the settings file or completely remove that part. If someone could open up the resulting DMG in Finder and show me the resulting window, that would be great.

closes #599


the-phinet:
Nice, this looks good and works well! Is there a way to disable the licence agreement bit at the beginning?

leahanderson1:
yes, i can just remove that bit of code afaik

the-phinet (dismissed):

the-phinet:
> If someone could open up the resulting DMG in Finder and show me the resulting window, that would be great.

Oh shit, I totally missed this request. I'm happy merging it as-is, if you are. Obviously it needs some art, and probably a way to open licenses.zip, but that can be added later.

image

leahanderson1:
> Oh shit, I totally missed this request. I'm happy merging it as-is, if you are.
> image

we should probably make it UZDoom.app instead of uzdoom.app

the-phinet:
Are you okay with me pushing the change directly to your pr?

leahanderson1:
sure

the-phinet (approved):

leahanderson1:
another thing that we could ideally do: skip the zip process entirely for the DMG file

the-phinet:
It's required (afaik) for the build step. The release step will unpack it. That's how the appimages are set up anyways

leahanderson1:
oh, then i suppose this is just about ready to merge, then.

PR #1212, opened 2026-04-01
Fix some release issues

the-phinet:
- I forgot to filter the pr list
- If a release runs when the branch has been updated, the notes generated were incorrect (temporarily)

## Checklist

### DIFF
--- a/.github/workflows/release_core.yml
+++ b/.github/workflows/release_core.yml
@@ -10,6 +10,9 @@ on:
       run-url:
         required: true
         type: string
+      commit:
+        required: true
+        type: string
       branch:
         required: true
         type: string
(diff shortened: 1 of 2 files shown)


RicardoLuis0 (approved):

RicardoLuis0 (changes requested):
ah wait just realized, this is still missing the commit message (though probably minus description, for brevity) for the base commit

the-phinet:
Oh, I figured that wasn't really required. I'll add it back :)

RicardoLuis0:
> Oh, I figured that wasn't really required. I'll add it back :)

the reason i wanted the base commit message there in the first place is to make it easy for people to tell when the experimental branch updates only for synching upstream, having only the commit sha makes it hard to tell for non-technical people

the-phinet:
Ahh that makes sense. I have it as a clicky link now, but I'll switch it over to text

Edit: Done. ~~I did not test the current iteration, but it should be fine. If you want I can do that~~ Running a test now lol. See you in 45 min

RicardoLuis0 (dismissed):

the-phinet:
There was a minor issue. It's fixed and confirmed working now

RicardoLuis0 (approved):

PR #1215, opened 2026-04-02
Fix specific fonts selection in launcher since stringtable refactor

dwing4g:
Include some related bugs fixed:
- `zh-*-HK` (HongKong) uses `zh-Hant` (Tranditional Chinese).
- lang names in the specific font table should be exactly the same as the language.csv header.
- add `GetCanvas()->setLanguage` for error window.

## Checklist

### DIFF
--- a/src/common/engine/stringtable.cpp
+++ b/src/common/engine/stringtable.cpp
@@ -167,7 +167,7 @@ inline FName GetFallback(FName name)
 		{"en-*-CA", "en-GB"},
 		{"en-*-SD", "en-GB"}, // TODO: support Subdivision
 		{"zh-*-CN", "zh-Hans-CN"},
-		{"zh-*-HK", "zh-Hans-HK"},
+		{"zh-*-HK", "zh-Hant-HK"},
 		{"zh-*-MO", "zh-Hant-MO"},
 		{"zh-*-SG", "zh-Hans-SG"},
 		{"zh-*-TW", "zh-Hant-TW"},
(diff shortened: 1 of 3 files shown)


the-phinet (changes requested):
Points 1 and 3 are valid fixes, but 2 is not. I do see the bug you're seeing, but this is not the way to fix this. I'll open a pr soon that should address it correctly.

Can you take a look at #1219, to see if it fixes the issue you were trying to fix here?

dwing4g:
> Points 1 and 3 are valid fixes, but 2 is not. I do see the bug you're seeing, but this is not the way to fix this. I'll open a pr soon that should address it correctly.

Yes, the fix for point 2 might not be very appropriate, but the recent stringtable refactor and fallback method have made font selection quite **tricky**. This is the smallest fix I can think of. If the stringtable requires further refactoring in the future, we can address it then, but we should keep this issue in mind.

I think it might be difficult to find a perfect solution to this problem. The fallback method can lead to the UI mixing text in different languages, however ZWidget does not support specifying a font for each widget. It's not necessary to be so precise, fallback fonts are enough. Therefore, once a language in stringtable is selected based on the Locale, it should be sufficient to prioritize one font for it, and most games just do it this way.

dwing4g:
> Can you take a look at #1219, to see if it fixes the issue you were trying to fix here?

Great! That PR works correctly on my locale. Although I'm not yet sure if it works properly in other language environments and operating systems.
I also considered this modification method, but for me, matching `lang-script-region` is too complicated. Almost all games don't take these details, except for large-scale and widely popular worldwide software like operating systems.

Anyway, this PR can be closed.

PR #1239, opened 2026-04-07
Add support for scrolling TextEdit by mouse wheel in ZWidget

dwing4g:
Affect multi-line text area such as `About` and `Release Notes` in Launcher.

## Checklist

### DIFF
--- a/libraries/ZWidget/include/zwidget/widgets/textedit/textedit.h
+++ b/libraries/ZWidget/include/zwidget/widgets/textedit/textedit.h
@@ -61,6 +61,7 @@ protected:
 	void OnKeyChar(std::string chars) override;
 	void OnKeyDown(InputKey key) override;
 	void OnKeyUp(InputKey key) override;
+	bool OnMouseWheel(const Point &pos, InputKey key) override;
 	void OnGeometryChanged() override;
 	void OnEnableChanged() override;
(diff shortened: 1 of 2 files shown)


RicardoLuis0:
this should probably be sent to https://github.com/UZDoom/ZWidget instead

dwing4g:
> this should probably be sent to https://github.com/UZDoom/ZWidget instead

You are right.
It seems that `UZDoom/ZWidget` is keeping track of the upstream repo `dpjudas/ZWidget`. So I just added the feature PR on the upstream.
And this PR can be closed.

the-phinet:
> It seems that UZDoom/ZWidget is keeping track of the upstream repo dpjudas/ZWidget. So I just added the feature PR on the upstream.

We don't grab anything from `dpjudas/ZWidget` anymore. The current state of the upstream is very broken

Can you send a pr to the `legacy` branch of https://github.com/UZDoom/ZWidget

dwing4g:
> We don't grab anything from `dpjudas/ZWidget` anymore. The current state of the upstream is very broken
> 
> Can you send a pr to the `legacy` branch of https://github.com/UZDoom/ZWidget

OK!
https://github.com/UZDoom/ZWidget/pull/5

PR #1242, opened 2026-04-08
Share node cache between ports

the-phinet:
As requested by @elf-alchemist

## Checklist


src/common/rendering/gl/gl_shader.cpp
```diff
@@ -100,7 +100,7 @@ static FString CreateProgramCacheName(bool create)
 {
 	FString path = M_GetCachePath(create);
```
RicardoLuis0, line 101: shader caches should not be shared

RicardoLuis0: same for vulkan caches/etc

the-phinet: They're not, they use the default uzdoom namespace

image

RicardoLuis0 (approved):

elf-alchemist:
Very nice :D

Issue #1083, opened 2026-02-27
Launcher ListViews are not scrolled to the selected items initially


dwing4g:
### UZDoom Version

git HEAD

### Game

All

### Operating System

Windows 10

### Duplicate Confirmation

- [x] I have searched and not found a similar issue.

### Bug Description

`GamesList` and `LangList` in launcher should be scrolled to the selected items at the beginning.
This issue happens when a list is too long to show in the list window and the last selected item is listed below the window.

In `playgamepage.cpp`, there is one `ScrollToItem` that does not work, because `ContentGeometry` is not updated.
```
	if (info.DefaultIWAD >= 0 && info.DefaultIWAD < info.Wads->SSize())
	{
		GamesList->SetSelectedItem(info.DefaultIWAD);
		GamesList->ScrollToItem(info.DefaultIWAD); // <== not works
	}
```

This issue can be solved by adding `GamesList->ScrollToItem(GamesList->GetSelectedItem());` at the end of `PlayGamePage::OnGeometryChanged()`, so does the `LangList` in `settingspage.cpp`.
However, there maybe better ways.

### Configuration

```ini

```

PR #1245, opened 2026-04-08
Fix scrolling to the selected item for GamesList and LangList (#1083)

dwing4g:
I've decided to release a PR to solve the issue (#1083) I raised so many days before.
Calling `ScrollToItem` in `OnGeometryChanged` seems to be a decent and simple solution. The only side effect is that when switching tabs or changing the window size, it will also scroll to the selected item, but I don't think it's bad.

## Checklist


the-phinet:
The fix looks good. Since you've already done the work I’m happy to merge this. To be transparent, I've been trying to save `good first issue` tasks for newcomers. Especially with the summer break coming up, I expect more students will be looking to enter into open source.

the-phinet (approved):
Oops, forgot to approve this

PR #1256, opened 2026-04-09
Crosshair improvements

the-phinet:
- Added contrast inverted crosshair
- Exposed crosshair colours to menu

[Screencast_20260409_195038.webm](https://github.com/user-attachments/assets/e9c3d66f-9ef1-48c0-b507-989f43f1cdfa)

## Checklist

### DIFF
--- a/src/common/console/c_cvars.cpp
+++ b/src/common/console/c_cvars.cpp
@@ -169,7 +169,7 @@ FBaseCVar::FBaseCVar (const char *var_name, uint32_t flags, void *callback, cons
 	if (var_name)
 	{
 		var = FindCVar(var_name, NULL);
-		C_AddTabCommand (var_name);
+		if (!(flags & CVAR_HIDDEN)) C_AddTabCommand (var_name);
 		VarName = var_name;
 		VarFName = var_name;
 		cvarMap.Insert(var_name, this);
--- a/src/common/console/c_cvars.h
+++ b/src/common/console/c_cvars.h
@@ -68,6 +68,7 @@ enum
 	CVAR_ZS_CUSTOM_CLONE	= 1 << 20, // Clone of a Custom ZScript CVar
 	
 	CVAR_SYSTEM_ONLY		= 1 << 21, // System-related cvar that should only ever be changed by the user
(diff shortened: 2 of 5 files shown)


mc776:
oh hell yeah

RicardoLuis0:
requiring alpha in crosshair images would break mods with custom crosshairs

RicardoLuis0:
it should be added as a new optional parameter in XHAIRS, default off for compatibility, and explicitly enabled for the default crosshairs (in https://github.com/UZDoom/UZDoom/blob/cd734a548d26a718dee79d67f356fb7dd4f11f37/wadsrc/static/xhairs.txt)

the-phinet:
Instead of that, is there a way to translate the monochrome pixel data to alpha in a way that can be inverted? Only reason I added the alpha channel is because I couldn't get the alpha blend to work with the monochrome bits. It's literally not doing anything otherwise

RicardoLuis0:
> Instead of that, is there a way to translate the monochrome pixel data to alpha in a way that can be inverted? Only reason I added the alpha channel is because I couldn't get the alpha blend to work with the monochrome bits. It's literally not doing anything otherwise

should just need `STYLEALPHA_SrcCol`/`STYLEALPHA_InvSrcCol` for non-alpha textures instead of `STYLEALPHA_Src`/`STYLEALPHA_InvSrc` IIRC

the-phinet:
Alpha stuff has been fixed :)

MajorCooke:
I hope there's an option to render the crosshair normally as it was before, because I noticed in the latest git versions, hovering a crosshair over subtractive renderstyles causes it to [disappear...](https://puu.sh/KLv0D/c052f1db9e.png)

the-phinet:
> I hope there's an option to render the crosshair normally as it was before, because I noticed in the latest git versions, hovering a crosshair over subtractive renderstyles causes it to [disappear...](https://puu.sh/KLv0D/c052f1db9e.png)

This PR makes no changes to the method uses to render the default crosshair, so it should make no difference for that use case

TehRealSalt (approved):

PR #1257, opened 2026-04-10
Silence a bunch of warnings and a leftover print statement

the-phinet:
## Checklist


.github/workflows/pr_bot.yml
RicardoLuis0, line 1: is there a way to just disable a workflow by default? i think doing that and having people who want the bot just enable it would be better than hardcoding repositories

RicardoLuis0:
also seems there's a more up to date fork of DUMB (though it has also stopped development like the original one)
https://github.com/kode54/dumb
https://github.com/kode54/dumb/blob/master/CHANGELOG.md

image
image

libraries/ZMusic/thirdparty/libxmp/src/loaders/vorbis.c
```diff
@@ -1451,7 +1451,7 @@ static int set_file_offset(stb_vorbis *f, unsigned int loc)
    #endif
    f->eof = 0;
    if (USE_MEMORY(f)) {
-      if (f->stream_start + loc >= f->stream_end || f->stream_start + loc < f->stream_start) {
```
RicardoLuis0, line 1454: this doesn't seem to check the `f->stream_start + loc < f->stream_start` case?

RicardoLuis0:
also GME also has newer versions https://github.com/libgme/game-music-emu/releases

RicardoLuis0:
same for libxmp https://github.com/libxmp/libxmp/releases

RicardoLuis0:
(and any fixes to still-maintained that are still needed after updating should probably be sent upstream first)

RicardoLuis0:
same for timidity 😅 
https://github.com/sezero/libtimidity
(timidity++ on the other hand, is indeed out of support)

RicardoLuis0:
https://github.com/Mindwerks/wildmidi also has a ton of updates that need to be pulled from upstream

RicardoLuis0:
other than those minor nitpicks with zmusic, all the changes to the main code seem absolutely fine

.github/workflows/pr_bot.yml
the-phinet: Not that I know of. If it makes you feel any better, this is also the approach used in the pages trigger action

libraries/ZMusic/thirdparty/libxmp/src/loaders/vorbis.c
the-phinet: I believe they are equivalent, but I might be wrong...

```
(A + loc >= B || A + loc < A)
(loc >= B - A)

loc=10
A=240
B=250
(250 >= 250 || 250 < 250) T
(10 >= 10) T

loc=9
A=240
B=250
(249 >= 250 || 249 < 240) F
(9 >= 10) F

loc=10
A=250
B=255
(5 >= 255 || 5 < 250) T
(10 >= 5) T

loc=10
A=240
B=245
(250 >= 245 || 250 < 240) T
(10 >= 5) T
```

the-phinet:
Okay, I'm slowly going through to see what libs I can update. (These really should be submodules, too)

- libxmp: ours diverges from upstream enough that I have no idea how to update it
- DUMB: ours diverges from upstream enough that I don't have the mental bandwidth for. Last commit tracked is [566c9876c74f170ee0fdb1e0eb33147581f3d490](https://github.com/kode54/dumb/commit/566c9876c74f170ee0fdb1e0eb33147581f3d490)
- GME: ~~Updated~~ nvm, it was grabbing a system library. Ours diverges from upstream enough that I don't have the mental bandwidth for.
- Timidity: ours diverges from upstream enough that I have no idea how to update it
- Timidity++: unsupported
- oplsynth: unsupported
- WildMIDI: ours diverges from upstream enough that I have no idea how to update it

libraries/ZMusic/thirdparty/libxmp/src/loaders/vorbis.c
RicardoLuis0: > I believe they are equivalent, but I might be wrong...
> 
> ```
> (A + loc >= B || A + loc < A)
> (loc >= B - A)
> 
> loc=10
> A=240
> B=250
> (250 >= 250 || 250 < 250) T
> (10 >= 10) T
> 
> loc=9
> A=240
> B=250
> (249 >= 250 || 249 < 240) F
> (9 >= 10) F
> 
> loc=10
> A=250
> B=255
> (5 >= 255 || 5 < 250) T
> (10 >= 5) T
> 
> loc=10
> A=240
> B=245
> (250 >= 245 || 250 < 240) T
> (10 >= 5) T
> ```

the only time `f->stream_start + loc < f->stream_start` is relevant is if loc is negative, which would never pass in `loc >= f->stream_end - f->stream_start`

(`loc >= f->stream_end - f->stream_start` is only equivalent to the `f->stream_start + loc >= f->stream_end` part of the original test)

RicardoLuis0: and since there's a check that would only pass if loc is negative, i can only assume that it _can_ be negative

the-phinet: `loc` is unsigned, so it's only an overflow check

RicardoLuis0: though it can be simplified to `loc < 0` instead of `f->stream_start + loc < f->stream_start`, unless it's a really shitty way of checking for integer overflow? (which wouldn't work, the compiler assumes integers will not overflow, so it just optimizes it to `loc < 0`)

RicardoLuis0: (and it's such a simple optimzation that it does it even in O0 btw https://gcc.godbolt.org/z/7eahYTn7b)

RicardoLuis0: oh wait no, it's pointer math, so it's only optimised away in O1, not in O0
https://gcc.godbolt.org/z/Wq3b44dq7

RicardoLuis0: but now that i see that it's pointer math, i can see that it cannot in any way be an overflow check, because a 32-bit int can't overflow a pointer, it is just a convoluted way to check if `loc < 0`

the-phinet: > but now that i see that it's pointer math, i can see that it cannot in any way be an overflow check, because a 32-bit int can't overflow a pointer, it is just a convoluted way to check if loc < 0

Right, becasue of virtual memory addresses, correct? Given that, and the fact that loc cannot be negative, this check can be entirely omitted, correct?

RicardoLuis0: > Right, becasue of virtual memory addresses, correct?

no, just that, adding a 32-bit value to a 64-bit value cannot overflow it

RicardoLuis0: though i guess it was probably written with 32-bit pointers in mind, in which case it would be an overflow check

RicardoLuis0: > though i guess it was probably written with 32-bit pointers in mind, in which case it would be an overflow check

or maybe loc was signed at some point in the past? and it was changed to unsigned but the check not omitted

the-phinet: > adding a 32-bit value to a 64-bit value cannot overflow it

Wait, what? That's not true.
```
uint64_t a = 0xfffffffffffffff7;
uint32_t b = 12;
uint64_t c = a + b;
printf("%llu\n", c); // 3
```

> or maybe loc was signed at some point in the past? and it was changed to unsigned but the check not omitted

I assume that's what happened, yeah

RicardoLuis0: > > adding a 32-bit value to a 64-bit value cannot overflow it
> 
> Wait, what? That's not true.

ah, true 🤦‍♀️ (my brain must've been thinking about the usual address range for pointers xD)

the-phinet:
I've removed the fixes in zmusic from here so they're not blockers. They should be pr'd to zmusic anyways

RicardoLuis0 (approved):

PR #1268, opened 2026-04-13
Added vsync and dynlight toggle to launcher

the-phinet:
The option was added to the launcher after I asked to change the default. An actual onboarding process would be better, but this should be fine until then

## Checklist

### DIFF
--- a/src/g_cvars.cpp
+++ b/src/g_cvars.cpp
@@ -60,9 +60,8 @@ CUSTOM_CVAR (Int, turnspeedsprintslow, 320, CVAR_ARCHIVE | CVAR_GLOBALCONFIG)
 	if (self <= 0) self = 1;
 }
 
-
-
-CUSTOM_CVAR (Bool, gl_lights, true, CVAR_ARCHIVE | CVAR_GLOBALCONFIG | CVAR_NOINITCALL)
+DEPR_CVAR (Bool, gl_lights, true, "Use r_dynlights");
+CUSTOM_CVAR (Bool, r_dynlights, true, CVAR_ARCHIVE | CVAR_GLOBALCONFIG | CVAR_NOINITCALL)
 {
 	for (auto Level : AllLevels())
 	{
--- a/src/common/engine/i_interface.cpp
+++ b/src/common/engine/i_interface.cpp
@@ -81,6 +81,8 @@ EXTERN_CVAR(Bool, ui_generic)
 EXTERN_CVAR(Int, vid_preferbackend)
 EXTERN_CVAR(Bool, vid_fullscreen)
 EXTERN_CVAR(Bool, vid_vsync)
+EXTERN_CVAR(Bool, r_dynlights)
+EXTERN_CVAR(Bool, gl_light_shadowmap)
 
 #ifdef HAS_UPDATER
 EXTERN_CVAR(Int, updater_update_interval)
@@ -106,6 +108,8 @@ FStartupSelectionInfo::FStartupSelectionInfo(const TArray<WadStuff>& wads, FArgs
 	DefaultBackend = vid_preferbackend;
 	DefaultFullscreen = vid_fullscreen;
 	DefaultVsync = vid_vsync;
+	DefaultDynLights = r_dynlights;
+	DefaultShadowmaps = gl_light_shadowmap;
 
 	if (defaultiwad[0] != '\0')
 	{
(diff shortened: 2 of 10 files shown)


Boondorl:
Not the biggest fan of some of the random cleanup in here since it seems to be a more than just formatting, but I think I'll have to leave it to consensus if it's fine

the-phinet:
> Not the biggest fan of some of the random cleanup in here since it seems to be a more than just formatting, but I think I'll have to leave it to consensus if it's fine

Do you mean the removal of the Raze hack? (I've removed it and added it to its own PR #1291)

Edit: yes that was the concern

leahanderson1:
@the-phinet you should prolly resolve the merge conflicts

the-phinet:
> you should prolly resolve the merge conflicts

Lol forgot about this. I'll fix those up today

leahanderson1 (approved):
LGTM

src/g_cvars.cpp
```diff
@@ -60,9 +60,7 @@ CUSTOM_CVAR (Int, turnspeedsprintslow, 320, CVAR_ARCHIVE | CVAR_GLOBALCONFIG)
 	if (self <= 0) self = 1;
 }
 
-
-
-CUSTOM_CVAR (Bool, gl_lights, true, CVAR_ARCHIVE | CVAR_GLOBALCONFIG | CVAR_NOINITCALL)
```
Boondorl, line 65: I believe this was accessible via ZScript so it needs to be deprecated rather than deleted

RicardoLuis0: it should also stay as gl_lights

RicardoLuis0: (and they should be toggleable independently, someone may want lights in hw mode but not in sw mode)

the-phinet: > I believe this was accessible via ZScript so it needs to be deprecated rather than deleted

Good call. Fixed

> it should also stay as gl_lights
> (and they should be toggleable independently, someone may want lights in hw mode but not in sw mode)

I strongly disagree. Our options are cluttered enough already. The only difference between these is an implementation detail. If someone is particular enough to need a feature on in one context but not another, they can easily maintain multiple config files.

RicardoLuis0: > > it should also stay as gl_lights
> > (and they should be toggleable independently, someone may want lights in hw mode but not in sw mode)
> 
> I strongly disagree. Our options are cluttered enough already. The only difference between these is an implementation detail. If someone is particular enough to need a feature on in one context but not another, they can easily maintain multiple config files.

lights in software mode are fundamentally different from hardware mode, ex. they look horrible with paletted renderer, perform poorly in both it and the truecolor one, wanting to have dynlights on for hw, and off for sw is something that should be possible without needing to clunkily switch config files

the-phinet: You're only arguing to keep them separated since they've been separated for a long time. We don't need a permutation for every setting with every backend used. I could argue that mouselook needs a separate toggle for software rendering for exactly the same reasons you've given -- You would rightfully reject that.

Keep in mind this change is being made for technical reasons. At this late stage, I'd rather not go back and rework how things are

Boondorl: We can always come back to this after 5.0 if it ends up being meaningful. If not, I'm fine with just consolidating it to one cvar because tbh I doubt someone is regularly switching between software and hardware and needs separate settings for both.

Boondorl (dismissed):

Boondorl:
I have some more thoughts on this PR, but I think it's harmless enough at this point. I don't want to sound like a bureaucrat, but I think combined with some of the private discussion around this, it's obvious we're lacking needed professional standards on submitting PRs and organizing them and that's causing a lot of unnecessary friction. We can't keep throwing in random reworks and preference changes for features, especially when the slated PR does not mention any of these things. Post 5.0 we're going to need to start enforcing some policy here. I think it'll result in a smoother pipeline for everyone involved and should result in less arguing over irrelevant details by simply not allowing irrelevant details to slip in.

the-phinet:
Trying to fix the conflict, I really screwed up the history. Thankfully git remembers all :sweat_smile:

leahanderson1:
"changes can be cleanly merged"
its new reviewin' time

Boondorl (approved):

PR #1278, opened 2026-04-15
Fixed double-clicking behavior

DyNaM1Kk:
Fixes double bindings not keeping normal bindings on the same key from executing. New version of the code now checks if a key has a double binding before doing anything. If it does, it delays the execution of the normal binding by a value given by the cl_doubleclickthreshold user CVar (250ms by default vs previous 571ms). If another press arrives in the time window, only the double binding gets executed. Keys that don't have double bindings fall back to standard behavior so there's no input latency when it's not necessary.

## Checklist

### DIFF
--- a/src/common/console/c_bind.cpp
+++ b/src/common/console/c_bind.cpp
@@ -24,6 +24,7 @@
 
 #include <stdint.h>
 
+#include "doomdef.h"
 #include "cmdlib.h"
 #include "keydef.h"
 #include "c_commandline.h"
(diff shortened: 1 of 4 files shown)


mc776:
This is going to be awful for everyone who's set up their doublebinds based on the current behaviour (so that the primary bind always executes but in a context where it doesn't even matter).

RicardoLuis0:
> This is going to be awful for everyone who's set up their doublebinds based on the current behaviour (so that the primary bind always executes but in a context where it doesn't even matter).

that can be fixed by making the doublebinds in question an alias that executes both actions though

the-phinet:
> This is going to be awful for everyone who's set up their doublebinds based on the current behaviour (so that the primary bind always executes but in a context where it doesn't even matter).

In this case, I'm pretty okay with this. I'd rather the engine do things correctly than support old workarounds.

Also: https://xkcd.com/1172/

Boondorl:
Seconding this feature as 5.0 is exposing it in menus proper and unfortunately its current implementation is extremely lacking. As an example, I wanted to set it so double tapping our ability key made it fire a secondary ability instead but with the existing behavior (always fire the single tap bind as well), that makes this functionally impossible. This is especially huge for controllers since you don't have much room there and need to take advantage of every button you can. It leaves me wondering what the actual use case of this feature truly was, which is why I think it was never properly exposed. It's extremely niche at the moment.

A possible solution for existing binds is to put in a parameter that marks whether or not the bind gets the delayed execution treatment. Any existing double tap binds will assume false while new ones will assume true (with the option to pass false from the console). I think the new behavior is just far more intuitive for its use cases so should at the very least be the default for mods looking to use it.

Edit: the easiest way to add a parameter might honestly be just using a new command for double tap binds that use the delayed execution behavior. Since the current menu stuff isn't in live yet we can freely adjust that as needed.

DyNaM1Kk:
I thought about this before doing the rework, but I couldn't think of cases when someone would want the old behavior. It seemed like an oversight to me, hence the PR name.

Still, after some experimenting and thinking about this for a while from both perspectives again, I don't think backwards compatibility is worth it for two reasons.

Main reason is that it adds complexity for a very niche use case IMO. I would like to see input handling be expanded in the future as it's very primitive at the moment. We don't even have a proper abstraction layer for chorded or contextual actions, which would elevate user interaction so much for mods and games, especially on controllers. I fear those would be harder and messier to implement and maintain later if we were to support the old behavior too.

Second is that it's trivial to fix from the user/mod side, but requires good effort from the engine side. Not to disregard users who have had their binds set with the old behavior in mind, but I don't think it'd be possible to support both in a clean and elegant way.

Boondorl:
> This is going to be awful for everyone who's set up their doublebinds based on the current behaviour (so that the primary bind always executes but in a context where it doesn't even matter).

Looking at this again, I think I misread it slightly and assumed there was a contextual behavior here where someone might want to one tap vs two tap for order of execution, but on reread, yeah, this doesn't really make sense. If the execution of the first bind is in a context where it doesn't matter, then skipping it entirely shouldn't matter either? And as Jay mentioned, you can bind multiple commands via `;` separating them.

This feature looks good to go to me. If it becomes problematic enough we can look into adding a solution, but only if it doesn't hinder future expansions.

the-phinet (dismissed):

src/common/console/c_bind.cpp
```diff
@@ -135,12 +136,93 @@ const char *KeyNames[NUM_KEYS] =
 	"Guide",       "Pad_Misc",   "Pad_Touchpad",               //
 };
 
+CVAR(Int, cl_doubleclickthreshold, 250, CVAR_ARCHIVE | CVAR_GLOBALCONFIG)
```
the-phinet, line 139: This should be exposed to the menus in the accessibility section. The delay is too long for my taste, but will certainly be to short for someone with motor control issues.

DyNaM1Kk: I was actually going to do that in a separate PR. I'll include it here now that you mention it. Do you think 100 to 1000 is a good enough range?

the-phinet: Nice! And yeah, that sounds perfect. Probably adjustable in steps of 50?

the-phinet (approved):

PR #1287, opened 2026-04-18
Optmenu reset keybind

### LINKED ISSUE #756: A way to reset individual options in the menu
I'm pretty sure there is no way to reset an option to the default value from the option menu. Retroarch binds (iirc) Backspace/Y on keyboard/gamepad respectively.

the-phinet:
Backspace will reset most options now. Fields and binds have not been implemented, since those are weirdly complicated (at least the ones I was testing)

Fixes #756

## Checklist

### DIFF
--- a/wadsrc/static/zscript/engine/ui/menu/optionmenuitems.zs
+++ b/wadsrc/static/zscript/engine/ui/menu/optionmenuitems.zs
@@ -502,6 +502,17 @@ class OptionMenuItemOption : OptionMenuItemOptionBase
 			}
 		}
 	}
+
+	override bool MenuEvent (int mkey, bool fromcontroller)
+	{
+		if (mkey == Menu.MKEY_Clear && mCVar != null)
+		{
+			mCVar.ResetToDefault();
+			Menu.MenuSound("menu/change");
+			return true;
+		}
+		return Super.MenuEvent(mkey, fromcontroller);
+	}
 }
 
 //=============================================================================
--- a/wadsrc/static/zscript/engine/ui/menu/optionmenuitems.zs
+++ b/wadsrc/static/zscript/engine/ui/menu/optionmenuitems.zs
@@ -839,6 +839,7 @@ class OptionMenuSliderBase : OptionMenuItem
 	CONST HELD_RESET_TICS = 6;
 
 	// command is a CVAR
+	CVar mCVar;
 	double mMin, mMax, mStep;
 	int mShowValue;
 	int mDrawX;
@@ -872,6 +873,7 @@ class OptionMenuSliderBase : OptionMenuItem
 		mShowValue = showval;
 		mDrawX = 0;
 		mSliderShort = 0;
+		mCVar = CVar.FindCVar(command);
 		mDisplayScale = displayScale;
 		mValueFormat = valueFormat;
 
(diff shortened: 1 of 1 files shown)


RicardoLuis0:
> Backspace will reset most options now. Fields and binds have not been implemented, since those are weirdly complicated (at least the ones I was testing)
> 
> Fixes #756
> 
> ## Checklist
> * [x]  I have reviewed [CONTRIBUTING.md](../blob/trunk/CONTRIBUTING.md) and agree to its terms.

a keybind seems like an odd way for reseting, wouldn't a reset button next to the option work better, and be easier for people to know it exists?

the-phinet:
It's the same key used to reset keybinds, so there's a precedent in the engine. I've seen the reset keybind in some engines before (I think retroarch uses select on a gamepad?). A graphical button would be good in the future (especially during the menu redesign), but a keybind will be required in either case for gamepad users.

RicardoLuis0:
> It's the same key used to reset keybinds, so there's a precedent in the engine. I've seen the reset keybind in some engines before (I think retroarch uses select on a gamepad?). A graphical button would be good in the future (especially during the menu redesign), but a keybind will be required in either case for gamepad users.

it's used to _clear_ keybinds, not reset them to default, that's why it might be confusing to have it _reset_ settings

the-phinet:
It's really not that confusing. Anyways, I would consider "clearing" a cvar to be resetting it to default in most cases. The only exception to that is binds

the-phinet:
What are your thoughts on adding a keybind? That way binds can be cleared AND reset, not just cleared? Maybe the delete key?

RicardoLuis0:
> What are your thoughts on adding a keybind? That way binds can be cleared AND reset, not just cleared? Maybe the delete key?

might be better as just a dialog, press backspace, get a "do you want to clear or reset or cancel" choice

RicardoLuis0:
that way accidental presses of backspace don't get rid of binds either

the-phinet:
> might be better as just a dialog, press backspace, get a "do you want to clear or reset or cancel" choice

Specifically for binds, sure. But not for other options. Getting a popup for every thing you're trying to reset is wildly annoying.

So that's a no to adding a second key? And keeping backspace as the reset bind for options?

RicardoLuis0:
> So that's a no to adding a second key? And keeping backspace as the reset bind for options?

i just think a dialog would be more useful (doesn't need anyone to learn a new key, and keeps del's ability to be bound as a regular key as it can now)

RicardoLuis0:
> Specifically for binds, sure. But not for other options. Getting a popup for every thing you're trying to reset is wildly annoying.

just for binds yeah, other options don't even have a concept of "clear" to need a dialog

the-phinet:
> i just think a dialog would be more useful (doesn't need anyone to learn a new key, and keeps del's ability to be bound as a regular key as it can now)
> 
> just for binds yeah, other options don't even have a concept of "clear" to need a dialog

I feel that I'm having trouble following lol. Is this correct:
- backspace to reset options is fine for now (not binds)
- eventually (in 5.1) we can add a button to reset (during the menu rework)
- eventually (in 5.1) we can add a dialogue to rebind/clear/reset binds (during the menu rework)

RicardoLuis0:
> > i just think a dialog would be more useful (doesn't need anyone to learn a new key, and keeps del's ability to be bound as a regular key as it can now)
> > just for binds yeah, other options don't even have a concept of "clear" to need a dialog
> 
> I feel that I'm having trouble following lol. Is this correct:
> 
> * backspace to reset options is fine for now (not binds)
> * eventually (in 5.1) we can add a button to reset (during the menu rework)
> * eventually (in 5.1) we can add a dialogue to rebind/clear/reset binds (during the menu rework)

the dialogue doesn't _need_ the menu rework, it can just use the existing messagebox menu (but if you feel it's best for 5.1 then it can be left for 5.1)

the-phinet:
> the dialogue doesn't _need_ the menu rework, it can just use the existing messagebox menu (but if you feel it's best for 5.1 then it can be left for 5.1)

That's fair. I'd like 5.0 out the door soon, so I'm gonna say no to implementing the dialogue now. (I do agree it should exist in the future)

Boondorl (approved):

PR #1289, opened 2026-04-18
Fixed missing params in FillSelectMenu

DyNaM1Kk:
Caused a VM Abort when opening the menu where you select the reverb environments

## Checklist

### DIFF
--- a/src/common/audio/sound/s_reverbedit.cpp
+++ b/src/common/audio/sound/s_reverbedit.cpp
@@ -365,8 +365,8 @@ DEFINE_ACTION_FUNCTION(DReverbEdit, FillSelectMenu)
 			if (func != nullptr)
 			{
 				DMenuItemBase *item = (DMenuItemBase*)cls->CreateNew();
-				VMValue params[] = { item, &text, FName(cmd).GetIndex(), false, true };
-				VMCall(func->Variants[0].Implementation, params, 5, nullptr, 0);
+				VMValue params[] = { item, &text, FName(cmd).GetIndex(), false, true, (FIntCVar*)nullptr, 0, FName("Hide").GetIndex() };
+				VMCall(func->Variants[0].Implementation, params, 8, nullptr, 0);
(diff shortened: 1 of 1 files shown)


Boondorl:
That's very strange, it seems to call `OptionMenuItemCommand`'s Init which only has 4 parameters. Something else has to be wrong here if it's grabbing the base Init function seeing as these aren't virtuals

src/common/audio/sound/s_reverbedit.cpp
```diff
@@ -365,8 +365,8 @@ DEFINE_ACTION_FUNCTION(DReverbEdit, FillSelectMenu)
 			if (func != nullptr)
 			{
 				DMenuItemBase *item = (DMenuItemBase*)cls->CreateNew();
-				VMValue params[] = { item, &text, FName(cmd).GetIndex(), false, true };
-				VMCall(func->Variants[0].Implementation, params, 5, nullptr, 0);
+				VMValue params[] = { item, &text, FName(cmd).GetIndex(), false, true, (FIntCVar*)nullptr, 0, FName("Hide").GetIndex() };
+				VMCall(func->Variants[0].Implementation, params, 8, nullptr, 0);
```
Boondorl, line 369: I suspect the `Variants[0].Implementation` might be the culprit. I'm not really sure why this is being done this way

DyNaM1Kk: Maybe you're looking at an older version? It actually has 7 as I see [here](https://github.com/UZDoom/UZDoom/blob/8eb9e8c4a8394a09b3d2f7a73755c560106501eb/wadsrc/static/zscript/engine/ui/menu/optionmenuitems.zs#L229)

Boondorl: Huh, yeah, had no idea it was updated recently. I haven't updated my UZDoom in a bit but I didn't expect these changes lol

Boondorl (approved):

PR #1285, opened 2026-04-17
Allow for tooltip width limit to be changed

the-phinet:
## Checklist
- [x] I have reviewed [CONTRIBUTING.md](../blob/trunk/CONTRIBUTING.md) and agree to its terms.

Issue #1034, opened 2026-02-15
Localize language plural forms


the-phinet:
English (for the most part lol) is simple with how it handles plurals. If you have exactly 1 thing, it's just `1 thing`; if it's not exactly 1 thing, it's `%d things`. Often when writing out a sentence where you can have a variable number of things, you'll write `%d thing(s)`. Other times, you'll structure your sentence in a way where you don't have to worry about different plural forms.

Other languages can have different numbers of plural forms, with different rules. We now have the ability to encode and parse this information, so it would be really cool to properly handle and display them.

Here's a sample of some language's plural rules as we currently store them:
- Japanese: `nplurals=1; plural=0;`
- English: `nplurals=2; plural=n != 1;`
- Russian: `nplurals=3; plural=(n%10==1 && n%100!=11 ? 0 : n%10>=2 && n%10<=4 && (n%100<10 || n%100>=20) ? 1 : 2);`
- Hebrew: `nplurals=4; plural=(n == 1) ? 0 : ((n == 2) ? 1 : ((n > 10 && n % 10 == 0) ? 2 : 3));`

Some documentation:
- https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html
- https://www.gnu.org/software/gettext/manual/html_node/Translating-plural-forms.html

## Questions

This issue is to discuss a plan to implement this, answering the following questions:
- How should we represent plural forms in the translation files?
- How do we retrieve the current plural forms from the translations table in the engine?

Other questions that might be worth considering:
- How do we handle fallbacks if there is a plural, but no translation for it?
- Should we handle certain ranges in a special way? Often it is better to write "We ate no apples" instead of "We ate 0 apples". This may be a thing also in other languages.

Feel free to suggest any other things we should consider, and I can add them above.

## Strings found with plurals. 
These don't actually make sense in some languages. So if we were to add a plural system, they should be replaced. There are likely more in the language files, these are just the ones I was able to find from a quick search.
- [TXT_BADDIE_KILLED](https://hosted.weblate.org/translate/uzdoom/zdoom-engine-strings/en/?checksum=2f2775478c67e5ec) / [TXT_BADDIES_KILLED](https://hosted.weblate.org/translate/uzdoom/zdoom-engine-strings/en/?checksum=22229b480483a2d1)
[...]

Issue #757, opened 2025-12-16
Slider granularity shift mode


the-phinet:
Right now sliders can be defined to show more (or less) digits than the increment can touch. Using a mouse to control the slider can lead to partial values that are annoying to look at.

Example: 
> slider shows `0.86`, but the left/right keys change the value by `0.1`, so `1.0` can never be achieved

Example: 
> slider shows `0.3`, but the left/right keys change the value by `0.01`, making to you need to hit the key up to 10 times for the displayed value to change.

I'd like to add a shift mode that rounds the underlying value to the nearest displayed digit, then increments the value by the smallest displayable value

Example: 
> slider shows `0.86`, but the left/right keys change the value by `0.1`. Holding shift will change the the value by `0.01`

Example:
> slider shows `0.3`, the underlying value is `0.32`, but the left/right keys change the value by `0.01`. Holding shift round the underlying value to `0.3` and change it by `0.1`

Enjay001:
Good ideas.  This would be a good QoL improvement.

Allied to this, sometimes certain sliders will first appear at a default value, but because of the way the increments work, if you move the slider up/down to see how whateveritis looks at different values, you can never get back to the default value.  I'm not sure if that's a problem inherent in the slider logic itself, or simply how people have set up their sliders in mods, but it is frustrating.

mc776:
IIRC when I ran into this problem it was because the range was like -64 to 64 or whatever while the increment was in 5s or some other value that wasn't a factor of the range in either direction.

I don't think there's a way to fix this to catch all edge cases in UZDoom besides letting the user reset to default at any time, but I may be biased because I had just read #756.

the-phinet:
> IIRC when I ran into this problem it was because the range was like -64 to 64 or whatever while the increment was in 5s or some other value that wasn't a factor of the range in either direction.
> 
> I don't think there's a way to fix this to catch all edge cases in UZDoom besides letting the user reset to default at any time, but I may be biased because I had just read [#756](https://github.com/UZDoom/UZDoom/issues/756).

A shift mode would fix this. In this case, it would be: round to nearest integer and then force increment by 1

mc776:
Could this be made into round/increment 0.1 for sliders that have a range that cover fewer than 3 total possible whole-number values? Or is this more of just making sure things are reset to 0?

(I'm imagining 3 would be a range of -1.00 to +1.00 while 2 would typically be 0.00 to +1.00 with the general intention that some value in between would be what most people want)

PR #1290, opened 2026-04-19
Improved menu sliders

DyNaM1Kk:
Added following features to sliders:
- Direct text input of values.
- Hold-to-accelerate sliding.
- Display scale parameter that multiplies the internal value on the menu to show more user-friendly numbers.
- Suffix parameter that appends the specified string to the end of the displayed value.

Other notable changes:
- SliderBase now has 4 extra protected virtuals that allow control of the new features: `CanInputSliderValue`, `FormatSliderValue`, `ParseSliderValue`, `GetHoldStepSize`.
- Updated some slider definitions in MENUDEF to use display scale, suffixes, and adjusted step sizes, prioritizing user-friendliness.
- General cleanup of slider code, typo fixes and etc.

Notes:
- While I did go through every slider definition, I don't have the necessary knowledge for every option and how they behave to update them all. It would be great if others could take a look and rework the ones I skipped where necessary.
- Direct text input doesn't block non-numerical values. Therefore, letters can be typed in, but the value will just be set to the minimum in most cases. I personally think this is fine.

## Checklist


wadsrc/static/menudef.txt
```diff
@@ -1116,10 +1116,10 @@ OptionMenu VideoModeMenu protected
 	Option "$VIDMNU_FORCEASPECT",    vid_aspect,        ForceRatios
 	Option "$VIDMNU_CROPASPECT",     vid_cropaspect,    CropAspect
 	Option "$VIDMNU_SCALEMODE",      vid_scalemode,     ScaleModes
-	Slider "$VIDMNU_SCALEFACTOR",    vid_scalefactor,   0.25, 2.0, 0.25, 2
+	Slider "$VIDMNU_SCALEFACTOR",    vid_scalefactor,   0.25, 2.0, 0.25, 2, "", 0, Gray, 1.0, "x"
 	StaticText " "
 	Option "$DSPLYMNU_VSYNC",        vid_vsync,         OnOff
-	Slider "$VIDMNU_MAXFPS",         vid_maxfps,        35, 500, 1
+	Slider "$VIDMNU_MAXFPS",         vid_maxfps,        35, 500, 1, 0, "", 0, Gray, 1.0, " FPS"
```
the-phinet: This should use a l10n key. Maybe `UNIT_FPS`? That way the others can get `UNIT_*` keys if needed

the-phinet:
I like most of these changes, especially the units/display multiplier! 

I'm unsure about the accelerating sliders:
- If they were to be added, I would prefer the repeat delay change, not scaling the numbers
- I'm unsure how this will affect #1285 and #757
- In general I'm not a fan of accelerating UI elements

To be clear: This isn't me telling you to change it -- just a point of discussion

wadsrc/static/menudef.txt
```diff
@@ -2839,34 +2839,34 @@ OptionMenu "ReverbSettings" protected
...
+	SliderReverbEditOption "$REVMNU_Reverb_Pan_X",           -2000,   2000,   1,     3, 16, 1.0,    ""
+	SliderReverbEditOption "$REVMNU_Reverb_Pan_Y",           -2000,   2000,   1,     3, 17, 1.0,    ""
+	SliderReverbEditOption "$REVMNU_Reverb_Pan_Z",           -2000,   2000,   1,     3, 18, 1.0,    ""
+	SliderReverbEditOption "$REVMNU_Echo_Time",               0.075,  0.25,   0.005, 0, 19, 1000.0, " ms"
+	SliderReverbEditOption "$REVMNU_Echo_Depth",              0,      1,      0.01,  0, 20, 100.0,  "%"
+	SliderReverbEditOption "$REVMNU_Modulation_Time",         0.04,   4,      0.01,  0, 21, 1000.0, " ms"
+	SliderReverbEditOption "$REVMNU_Modulation_Depth",        0,      1,      0.01,  0, 22, 100.0,  "%"
+	SliderReverbEditOption "$REVMNU_Air_Absorption_HF",      -100,    0,      0.01,  2, 23, 1.0,    " dB/m"
+	SliderReverbEditOption "$REVMNU_HF_Reference",            1000,   20000,  1,     0, 24, 1.0,    " Hz"
+	SliderReverbEditOption "$REVMNU_LF_Reference",            20,     1000,   0.1,   1, 25, 1.0,    " Hz"
```
the-phinet: Were these ranges supposed to change?

wadsrc/static/menudef.txt
DyNaM1Kk: Woops, forgot to do that. Would these names be ok?
**UNIT_FPS
UNIT_MSECOND
UNIT_SECOND
UNIT_METER
UNIT_DB
UNIT_DB_P_METER
UNIT_HERTZ**

the-phinet: Yeah, those sound perfect to me

wadsrc/static/menudef.txt
DyNaM1Kk: Yes. They're clamped in the backend and the old ranges did not match those limits, which resulted in the slider cursor getting stuck towards the middle instead of moving fully from one end to the other. That simply fixes the UI and does not affect the final sound at all.

For example, the backend min/max is 1000000/20000000 as seen [here](https://github.com/UZDoom/UZDoom/blob/8ade5692386b815c7fdb017ca4a425942139a3c8/src/common/audio/sound/s_environment.cpp#L57), which then get divided by 1000 as seen [here](https://github.com/UZDoom/UZDoom/blob/8ade5692386b815c7fdb017ca4a425942139a3c8/src/common/audio/sound/s_environment.cpp#L542), so the actual range ends up being 1000/20000. The decay values also had the same issue.

RicardoLuis0:
it's probably best to defer this to 5.1, since it'll also be getting a menu rework

the-phinet:
> it's probably best to defer this to 5.1, since it'll also be getting a menu rework

I disagree. This is a minor enough change that it's perfect for 5.0. Holding on to it for 5.1 would be pointless due to the aforementioned menu rework.

DyNaM1Kk:
@the-phinet 
I looked at https://github.com/UZDoom/UZDoom/pull/1285. I don't see any possible conflicts. None of the changes should overlap. Only problem is that I think your PR makes the rounding math incorrect for sliders where `mDisplayScale` is not 1. Feel free to merge that one first. I'll rebase and fix the math here after if needed.

As for acceleration, the problem is that current sliders are very slow with large ranges, especially with small step sizes. While you can use a mouse for that, controllers don't have that option. Personally, I prefer keys anyway since they give you more control and convenience IMO.

As for how acceleration is done, I thought about it and decided your idea did sound better. I reimplemented it the way you suggested, but it didn't feel quite right. Input is per tic and the current repeat rate is 3/35. I made it so it increased to 2/35 and then to 1/35 when held for long. The problem with that was that large ranges were still extremely slow and unfortunately, I can't make it go any faster without a major rewrite. Besides, I think the current percentage-based implementation is great since it will work nicely with all existing sliders both in the engine and in mods no matter the range or step size. If the values aren't satisfactory, I can tweak them.

wadsrc/static/menudef.txt
the-phinet: Oof, thinking this through even more, `x` and `%` should be localized, too. (and if you're feeling spicy, piping them through a through a stringformat). While those are pretty common, I do know that some languages use different symbols or placement of the symbol

the-phinet:
> I looked at https://github.com/UZDoom/UZDoom/pull/1285. I don't see any possible conflicts. None of the changes should overlap. Only problem is that I think your PR makes the rounding math incorrect for sliders where mDisplayScale is not 1. Feel free to merge that one first. I'll rebase and fix the math here after if needed.

Sounds good!

> As for acceleration, the problem is that current sliders are very slow with large ranges, especially with small step sizes. While you can use a mouse for that, controllers don't have that option. Personally, I prefer keys anyway since they give you more control and convenience IMO.

Yeah, I never use the mouse in menus. I do wonder if this is a bit of an AB problem. Maybe those ranges/steps need to be changed?

> As for how acceleration is done, I thought about it and decided your idea did sound better. I reimplemented it the way you suggested, but it didn't feel quite right. Input is per tic and the current repeat rate is 3/35. I made it so it increased to 2/35 and then to 1/35 when held for long. The problem with that was that large ranges were still extremely slow and unfortunately, I can't make it go any faster without a major rewrite. Besides, I think the current percentage-based implementation is great since it will work nicely with all existing sliders both in the engine and in mods no matter the range or step size. If the values aren't satisfactory, I can tweak them.

Sounds good to me.

wadsrc/static/menudef.txt
DyNaM1Kk: I've been thinking about this. I can do this in two ways. First would be to use a single key like `UNITVAL_PERCENT = "{VALUE}%";` and modify the unit placement per language. Second would be to have two keys, unit key - `UNIT_PERCENT = "%";` and layout key - `UNIT_PERCENT_FMT = "{VALUE}{UNIT}";`. You'd only use the unit key in MENUDEF and modify the unit and layout keys per language where needed. The relevant ZScript code of course replaces the VALUE and UNIT patterns with the correct data.

I can't decide which to go with. While the clear advantage to the second way is the reusability of the units for any use case, it is not as clean and requires many more keys.

the-phinet: I think method 1 makes more sense for 2 reasons:
- with the specific example given, I don't think I see any advantage. If it was `UNIT_FMT`, I could see the reusability argument, but since it's `UNIT_PERCENT_FMT`, I don't see any benefit.
- if it was `UNIT_FMT`, I would be worried about a language having the unit on different sides of the number. In English, `$` and `%` appear on opposite sides, but in other languages they do not.

I think is standard string format would be fine? `%dhz` / `%d%%` / `$%d`

DyNaM1Kk: I think my wording wasn't the best. I'll give a better example:
```
[enu default]
UNIT_PERCENT = "%";
UNIT_PERCENT_FMT = "{VALUE}{UNIT}";

[Arabic]
UNIT_PERCENT = "٪";
UNIT_PERCENT_FMT = "{UNIT}{VALUE}";
```
This would make it so in English you'd see `50%` and in Arabic - `٪50`. Then anyone would be able to reuse those keys anywhere in the engine or modders could draw percentages on the screen without having to do any localization themselves. Code-wise, you'd just call `Replace` on the localized FMT string to replace "{VALUE}" and "{UNIT}" patterns.

Thinking about it more, I could add a key like `UNIT_GENERIC = "{VALUE}{UNIT}";` to handle most cases, which would reduce the amount of keys needed for this. This makes me lean more toward the second method, but I'd still like to hear what you or others think. I know there was an issue with the amount of strings on Weblate in the past so I'm trying to be careful.

the-phinet: Ahh, that makes more sense. If I understand what you're suggesting, you're envisioning that translators can do something like:
```
[enu default]
MECH_BATTERY = "Mech is $UNIT_PERCENT_FMT charged";

[Arabic]
MECH_BATTERY = "abc $UNIT_PERCENT_FMT xyz";
```
and then those macros will be filled in by us, correct?

I don't think that would any simpler or clearer than having the translator explicitly filling out the whole string:
```
[enu default]
MECH_BATTERY = "Mech is {value}% charged";

[Arabic]
MECH_BATTERY = "abc ٪{value} xyz";
```

The only benefit I do see is that the modder can request `$UNIT_PERCENT` by itself, but I'm not sure what use-case would require this. Maybe a fancy menu with units out of line? In that case the placement of those would still require the positioning information.

--- 

Now I may be wildly misunderstanding you again. I think in this case simpler is better, basically only worrying about our menu use-case. Strings in the Weblate is a minor concern, but my main concern is the layering of logic here.

I'm by no means an expert on this topic, so I'd like to as @jekyllgrim for some input here.

jekyllgrim: Well, let's try determining our use cases clearly.

If we're only talking about sliders that are supposed to show the text `"{value} unit"`:
- create separate string keys for every `"{value} unit"` combination, where the unit part is written out explicitly, such as `"{value}ms"`, `"{value} FPS"`, `"{value}%"`, etc.;
- do not create separate keys for the units themselves; let the translators edit the formatted combined entries above however they want.

If there's *also* ever a need to just display units alone (I can't imagine what that would be, though), it's fine to create string keys specifically for units. However, they should not be referenced in the `"{value} unit"` keys — that's unnecessary overconstruction.

Examples like Marcus's MECH_BATTERY seem to be purely mod-specific cases, so that would be up to the mod author to handle properly. But the reasonable way would again be to use `"battery is {value}% charged"`, and let the translator adapt and place % as needed. (Besides, we don't have an easy way to reference string keys inside other strings; to do this the author of the project would have to explicitly use string formatting.)

Additionally:
- using `x` as a shorthand for a multiplier is by all means not universal; in some languages it's unknown or considered very colloquial (edit: at this point, it's *probably* known practically everywhere, but it's not commonly used everywhere);
- spacing also varies—English may be fine with both "{value}ms" and "{value} ms", for instance, but in some other languages only one version is recognized as correct.

Hope that helps! Feel free to ask for clarifications.

jekyllgrim: P.S. Keep in mind that the literal "%" needs to be escaped as `"%%"`, since a single "%" is  treated as a formatting tag.

kcat:
Out of curiosity, are there any languages that format differently for different numbers? Like how in English, we do `1 unit` and `2 units`. Or maybe some language that does something like `<prefix> 0` and `1 <suffix>`. Is this a potential issue worth considering?

the-phinet:
> Out of curiosity, are there any languages that format differently for different numbers?

Yes, certainly. 

> Like how in English, we do `1 unit` and `2 units`. 

Yeah, there are many different ways languages handle this. I've seen languagea with up to 6 different plural forms (Arabic) and as little as 1 (Japanese)

> Or maybe some language that does something like `<prefix> 0` and `1 <suffix>`.

There are no languages that I'm aware of that do this. I believe there are some locales that place the currency sign between the whole and fractional part (10$50), but that feels out of scope here?

Another thing worth maybe considering is decimal and group separators.

> Is this a potential issue worth considering?

Yes I think so, but not right now. While none of the units used in the menu currently would benefit, I think it makes sense to pipe sliders through the plural form function ([when that gets implemented](https://github.com/UZDoom/UZDoom/issues/1034))

Boondorl:
Yeah, plurality is the bane of a lot of localization which is why it's often better to use fixed values in strings rather than try and dynamically build them. From talks, most game companies end up making their own custom solutions to handle it because you'd need a way to define rules for each language. Even using rule count is misleading because the rules themselves can be very esoteric and way more complex than the "1 or non-1" of English

jekyllgrim:
Worth noting, while the problem of plurality is extremely complex, for standardized units it's largely solved with abbreviations. So, with "FPS", "ms", "MB", "sec" and such this shouldn't be a concern. A custom system is still needed, but it'll mainly be useful for custom projects where authors may want to create strings with dynamic values and proper localization support.

River-Salmon (dismissed):

DyNaM1Kk:
I've rebased the branch. The only change I've made since the last commit is the scale math in `OptionMenuItemSlider`'s `Init`.
From:
```
scale = 10 ** mShowValue;
```
To:
```
scale = (10 ** mShowValue) * displayScale;
```
This fixes the values rounding down to 0 when `displayScale` is used. I won't be making any more changes so if the code looks good, I believe it's ready to be merged.

the-phinet (approved):

PR #1310, opened 2026-04-22
Add optional pride colors to launcher

the-phinet:
Adds a `-pride` flag (pun intended), which adds pride colours to the launcher banner. Currently there are 21 flags, but more can easily be added. Which flag is displayed is chosen randomly at boot. It can also be configured to be a static flag, or be a random one without needing to add the commandline flag. Also you can ask it to respect the theme (but color mixing can be weird looking, so that's off by default).

pride

## Checklist

### DIFF
--- a/src/common/engine/i_interface.cpp
+++ b/src/common/engine/i_interface.cpp
@@ -70,6 +70,8 @@ CVAR(Int, defaultnethostteam, 255, CVAR_ARCHIVE | CVAR_GLOBALCONFIG)
 CVAR(Int, defaultnetjointeam, 255, CVAR_ARCHIVE | CVAR_GLOBALCONFIG)
 CVAR(Bool, defaultnetextratic, false, CVAR_ARCHIVE | CVAR_GLOBALCONFIG)
 CVAR(String, defaultnetsavefile, "", CVAR_ARCHIVE | CVAR_GLOBALCONFIG)
+CVAR(String, ui_colors, "", CVAR_ARCHIVE | CVAR_GLOBALCONFIG);
+CVAR(Float, ui_color_mix, .35, CVAR_ARCHIVE | CVAR_GLOBALCONFIG);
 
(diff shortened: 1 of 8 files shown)


RicardoLuis0:
why not have it as a cvar? that way people can choose the flag of their preference to be displayed always (and that way it'd be persistent rather than needing a launch arg)

the-phinet:
> why not have it as a cvar? that way people can choose the flag of their preference to be displayed always (and that way it'd be persistent rather than needing a launch arg)

Lol great question. Mainly because I think the `-pride` flag pun is incredibly funny

I'll add some cvars to allow for persistence and customization (but I will be keeping the flag)

RicardoLuis0:
> > why not have it as a cvar? that way people can choose the flag of their preference to be displayed always (and that way it'd be persistent rather than needing a launch arg)
> 
> Lol great question. Mainly because I think the `-pride` flag pun is incredibly funny
> 
> I'll add some cvars to allow for persistence and customization (but I will be keeping the flag)

yeah, no harm at all in keeping the flag :3

RicardoLuis0:
actually, is it easy to have some flags draw an svg on top of the stripes? that way we can get the progress pride flag in the launcher (if it isn't easy it can just be added in a later PR)

the-phinet:
Great idea, but not easy. I would wait to do that until we have an actual ui toolkit.

Svgs can be drawn (I think), but anchoring them would be a nightmare. In addition to that, zwidget doesn't support translucency, so keeping the logo legible is a challenge.

RicardoLuis0:
> Great idea, but not easy. I would wait to do that until we have an actual ui toolkit.
> 
> Svgs can be drawn (I think), but anchoring them would be a nightmare. In addition to that, zwidget doesn't support translucency, so keeping the logo legible is a challenge.

i'll try and work it out myself for 5.1 then

RicardoLuis0 (approved):

PR #1311, opened 2026-04-22
Support a per-map noambientocclusion option in MAPINFO

samwiddowson:
## Checklist

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -270,3 +270,4 @@ Phil "Shadsy" Salvador
 Charles the Thobe
 Leah Anderson
 wallabra
+Sam "LordEntr0py" Widdowson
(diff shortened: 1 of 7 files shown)


RicardoLuis0 (changes requested):
looks good, but there are random whitespace changes through the files

samwiddowson:
> looks good, but there are random whitespace changes through the files

They are all trailing spaces. If they get removed now in this straightforward change future devs won't need to pick through their changes when their IDE tries to clean it up for them.

That said, I'd prefer to follow team policy than make a fuss so I'll remove the whitespace changes if preferred.

Edit: you might well be aware of it, but if you want to hide those charges temporarily for purpose of reviewing this change you can do so through the settings option next to the green "Review" button on the changes tab, then selecting "Hide whitespace".

RicardoLuis0:
> Edit: you might well be aware of it, but if you want to hide those charges temporarily for purpose of reviewing this change you can do so through the settings option next to the green "Review" button on the changes tab, then selecting "Hide whitespace".

they still show on git blame, so at the very least they should be on a separate formatting-only commit

samwiddowson:
> > Edit: you might well be aware of it, but if you want to hide those charges temporarily for purpose of reviewing this change you can do so through the settings option next to the green "Review" button on the changes tab, then selecting "Hide whitespace".
> 
> they still show on git blame, so at the very least they should be on a separate formatting-only commit

Makes sense! I separated the whitespace changes into a separate commit. Much cleaner now :)

RicardoLuis0 (approved):

River-Salmon (approved):

PR #1322, opened 2026-04-24
Fix whitespace formatting

the-phinet:
This project is a mess of mixed indentation modes and random trailing whitespace. Instead of every new PR having some formatting stuff in it, this rips that band-aid off. Once this gets merged, I'll add the `Fixup` commit to the blame ignore, so it doesn't appear in blames be default.

This PR only changes whitespace (you can verify that by `git diff upstream/trunk -w --ignore-blank-lines`). It trims trailing whitespace, ensures a trailing newline, and makes indentation consistent. It's a script that can be added to CI (the script should be removed when we start enforcing a format through clang-format).

The indentation heuristic may mess up on some files, but that can be resolved later. Indentation is only enforced when indentation is already inconsistent, or when it's a cpp/zs file.

## Checklist


TehRealSalt (dismissed):

tools/format-spaces.sh
```diff
@@ -0,0 +1,165 @@
+#!/bin/bash
```
Gutawer: `#!/usr/bin/env bash` would be better as not every linux distro has `/bin/bash`. for instance NixOS doesn't

nikitalita:
Why isn't this being controlled by something like `clang-format.yml`?

the-phinet:
> Why isn't this being controlled by something like `clang-format.yml`?

It will be. It's been deemed that we don't want to do a clang format until later. This is a stop-gap until then

nikitalita (approved):

River-Salmon (dismissed):

Boondorl (approved):

PR #1324, opened 2026-04-25
Account for signal mixing when spatializing all channels to the same 3D point source.

### LINKED ISSUE #1312: Stereo sounds played in the world are louder than mono sounds
### UZDoom Version

pre-237-g9f8c07c05e

### Game

All

### Operating System

Windows 10

### OS Version

_No response_

### Hardware

_No response_

### Duplicate Confirmation

### Bug Description

When stereo sounds (i.e, the sound file asset is in stereo; has two channels) is played in the world as a locational sound, the volume seems to be louder than when the sound file is in mono.

### What this is NOT about
- Stereo sounds being played in mono in the 3D world. Yes, it has been established that stereo sounds will be downmixed to mono first, before being played from a 3D location in the world. This is **NOT** what this report is about.

### What this IS about
- Stereo sounds being played in the world seems to be at a much louder volume compared to mono sounds.

### Replication

# Example PK3

[StereoSoundVolumeTest.zip](https://github.com/user-attachments/files/27003931/StereoSoundVolumeTest.zip)

- Load the example attached into UZDoom and play MAP01.
- Press the switch to play a locational sound emitting from the eye object. It will flip-flop between a mono sound and a stereo sound.

**Both sounds are literally the same sound, just one is in mono, while the other was the exact sound converted to stereo with proper volumes (i.e, the volume is NOT doubled just because it's the same mono audio duplicated left and right - I actually made sure the volume between both files are 100% similar)**

It should be rather obvious that one sound is louder than the other. If you have some kind of audio metering (either software or hardware), you can confirm this by just watching your meters, too.

You can also double-confirm that both sound assets are indeed of identical volume by opening up the PK3 in SLADE or whatever, and auditoning the two sounds in the `sounds` directory with your favorite audio app.

### Configuration

```ini

```

### Log

_No response_

thecatontheceiling:
Fixes #1312 

Some notes:
- I didn't force a call to the new GetSourceScale function in StartSound3D as the source isn't fully configured yet when that code is hit, instead I just calculated the scale manually there.
- I didn't call alGetError() in the new function as ``source`` and ``buffer`` are basically guaranteed to be valid there and shouldn't cause OpenAL errors.

## Checklist

### DIFF
--- a/src/common/audio/sound/oalsound.cpp
+++ b/src/common/audio/sound/oalsound.cpp
@@ -556,6 +556,26 @@ static float GetRolloff(const FRolloffInfo *rolloff, float distance)
 	return soundEngine->GetRolloff(rolloff, distance);
 }
 
+static float GetSourceScale(bool isSpatializeSupported, ALuint source)
+{
+	if (!isSpatializeSupported)
+		return 1.0f;
+
+	ALint spatialize = AL_AUTO_SOFT;
+	alGetSourcei(source, AL_SOURCE_SPATIALIZE_SOFT, &spatialize);
+	if (spatialize != AL_TRUE)
+		return 1.0f;
+
+	ALint buffer = 0;
+	alGetSourcei(source, AL_BUFFER, &buffer);
+	if (buffer == 0)
+		return 1.0f;
+
+	ALint chans = 1;
+	alGetBufferi(buffer, AL_CHANNELS, &chans);
+	return (chans > 1) ? (1.0f / chans) : 1.0f;
+}
+
 ALCdevice *OpenALSoundRenderer::InitDevice()
 {
 	ALCdevice *device = NULL;
@@ -989,9 +1009,11 @@ void OpenALSoundRenderer::SetSfxVolume(float volume)
 			ALuint source = GET_PTRID(schan->SysChannel);
 			volume = SfxVolume;
 
+			float scale = GetSourceScale(AL.SOFT_source_spatialize, source);
+
 			alDeferUpdatesSOFT();
-			alSourcef(source, AL_MAX_GAIN, volume);
-			alSourcef(source, AL_GAIN, volume * schan->Volume);
+			alSourcef(source, AL_MAX_GAIN, volume * scale);
+			alSourcef(source, AL_GAIN, volume * schan->Volume * scale);
 		}
 		schan = schan->NextChan;
 	}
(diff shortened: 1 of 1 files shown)


RicardoLuis0:
@kcat would you mind reviewing this one? i don't know enough about the openal code to be able to verify that this is (or isn't) the correct solution 😅

thecatontheceiling:
I'll add myself to CONTRIBUTORS and remove that command soon (probably) as I think it's pretty self-explanatory anyway

kcat:
I'm not fond of this adjustment. IMO, it makes stereo sounds played as spatialized mono too quiet, with a noticeable jump in volume as it gets close enough to not spatialize and a noticeable drop as it moves away to spatialize. Imagine a stereo sound that only has audio in the left channel. You'll want that to remain as-is when spatialized since that's all the energy of the sound, not inexplicably drop by -6dB. Or that same sound played on a single channel in a 5.1 mix, should also stay as-is rather than drop by -14 or -15dB just because of the number of channels carrying it.

Sounds in a proper multi-channel mix will have their energy spread across channels, not simply duplicated.

The spatialize extension originally did scale multi-channel sounds like this automatically, but I found it too disruptive to the original sound volume and took it out. This also makes it consistent with formats like B-Format which don't use discrete channels (it essentially has infinite channels as a sound can be placed anywhere on a spherical grid; spatializing to mono just takes the sum of the mix since the energy is spread out, not reduced by `1 / inf` or -inf dB).

thecatontheceiling:
I did realise this could be a problem, but I wasn't really sure what could be done about it. 

Maybe UZDoom could, before passing anything to OpenAL, check whether the loaded audio file has two channels, compare them, and discard one channel if they are identical (or almost identical), and then hand OpenAL a mono buffer so that all this hackery is avoided. Not sure if it's that simple in practice and how much effort it would take to analyze audio files like that. 

I learned how stereo panning works while I was working on this PR so you probably have better ideas.

kcat:
> Maybe UZDoom could, before passing anything to OpenAL, check whether the loaded audio file has two channels, compare them, and discard one channel if they are identical (or almost identical)

I wager most stereo sounds would be stereo for a reason and not be detected as nearly identical. And of those that are needlessly stereo and could have one channel dropped, the issue is with the intended volume relative to other sounds. Is it supposed to be relatively loud (where dropping half the energy makes it too quiet), or relatively quiet (where keeping all the energy may or may not be too loud)? That's not something UZDoom (or OpenAL) can make an assumption about, so it's up to the sound designer to make sure the sounds are properly balanced both individually (ensuring a sound's volume doesn't inadvertently change as a result of its pan in a stereo mix), and for the soundscape (ensuring a sound's relative volume is correct when played in the scene with other sounds). UZDoom can really only play what it's told to play and how its told to play it.

thecatontheceiling:
So is the original issue not really a bug with UZDoom then?

kcat:
Yeah, I don't think I'd call it a bug. The example given in that issue has the stereo sound authored with more total energy than the mono sound, so it should be expected to be louder. The stereo sound is effectively playing the mono sound twice a small distance apart.

nashmuhandes:
> So is the original issue not really a bug with UZDoom then?

There was an honest mistake when I prepared the example files in the OG thread. [Explained here](https://github.com/UZDoom/UZDoom/issues/1312#issuecomment-4321674894). I'll have to redo the example file and then revisit the issue

PR #1330, opened 2026-04-26
Remove discord rpc

RicardoLuis0:
if it's readded later, it'll need to use a different library, ex. [cemu's DiscordRPCLite](https://github.com/cemu-project/Cemu/tree/main/src/Cemu/DiscordPresence)

## Checklist


the-phinet:
Wouldn't it make more sense to comment out the discord stuff in `src` / `wadsrc`? I would assume when we swap to a different lib, the structure would be identical.

RicardoLuis0:
> Wouldn't it make more sense to comment out the discord stuff in `src` / `wadsrc`? I would assume when we swap to a different lib, the structure would be identical.

ah, good catch

the-phinet (approved):

PR #1345, opened 2026-04-30
Licensing Fixes

### LINKED ISSUE #1338: Missing licenses for vendored libraries
UZDoom potentially vendors some libraries without also exporting the license files.

I've made a guess below at which project you are vendoring under `libraries/` but the authors need to verify these are the correct licenses.

**asmjit**
This could be copied over:
https://github.com/asmjit/asmjit/blob/0bd5787b54b575ed94bf32ac452153b34385c514/LICENSE.md

**miniz**
This appears in the source which is probably fine though it may be good to copy this over:
https://github.com/richgel999/miniz/blob/224d207ce8fffb908e156d27478be3afb5d83e6a/LICENSE

**range_map**
There is this comment in the commit history:
> I have e-mailed the author for clarifcation on the license, will update this when they respond

https://github.com/UZDoom/UZDoom/tree/0e3cf203856a8759ccb7044f44c47f8530cc3734/libraries/range_map/include/range_map

It seems to only be used in one small part of the repository so it may be easier to simply replace it with something a simple inhouse implementation using a `std::vector` if the licensing is not clear. 

**superfasthash**
The source bundled only points to a website. It looks like it may be bundled with the [LGPL 2.1 license](http://www.gnu.org/licenses/lgpl-2.1.txt). This could potentially be added.

**xs_Float**
There appears to be no clear license here.

https://github.com/UZDoom/UZDoom/blob/0e3cf203856a8759ccb7044f44c47f8530cc3734/src/common/thirdparty/xs_Float.h#L1-L14

Again, perhaps a small in house implementation may remove the need for this.

the-phinet:
Resolves #1338

- Adds license text for miniz, asmjit, and range_map to licenses.zip
- Swaps SuperFastHast to the BSD license, adding it to licenses.zip. This is explicitly allowed in the text of the release of SFH. While the old license (and the LGPL version) are both compatible with us, BSD is easier.
- Removes xs_Float, replacing with our own code

The first 2 points are straightforward, but the xs_Float on is less so. Here's a why:
- No license for xs_Float is given in text, or in the [original post](https://stereopsis.com/sree/fpu2006.html)
- The [author](http://sree.kotay.com/) is likely not able to be contacted due to legal reasons.
- There are a number of technical issues with it:
  - It relies in UB
  - "We cannot do that with compiler provided conversions because they differ too much", while the library produces different (and sometimes very incorrect) results between Windows and Linux
  - The bit-magic is likely not faster on modern computers anymore

The replacement has been tested a bunch for regressions, and none have been found. (Kept the testing commit, since that shows how the functions were renamed, and it allows for further testing if issues come up)

When swapping the xs_Float function calls to our own, I noticed a couple questionable rounding methods used in some places. I'm wondering if those were used because of the bad function names making it unclear what they did?

## Checklist


src/gamedata/r_defs.h
```diff
@@ -591,12 +591,12 @@ struct FTransform
...
 	DAngle Angle, baseAngle;
 
-	finline bool operator == (const FTransform &other) const
+	inline bool operator == (const FTransform &other) const
 	{
 		return xOffs == other.xOffs && yOffs + baseyOffs == other.yOffs + other.baseyOffs &&
 			xScale == other.xScale && yScale == other.yScale && Angle + baseAngle == other.Angle + other.baseAngle;
 	}
-	finline bool operator != (const FTransform &other) const
+	inline bool operator != (const FTransform &other) const
```
River-Salmon, line 599: Is the intent for the inline here to be a compiler hint, or a linking hint? Is this demonstrating the same intent as the original finline def?

River-Salmon (approved):

src/gamedata/r_defs.h
RicardoLuis0: > Is the intent for the inline here to be a compiler hint, or a linking hint? Is this demonstrating the same intent as the original finline def?

it's not the same, that's there to force the function to be inlined, while regular `inline` is only guaranteed to allow duplicate definitions (and while gcc does try its best to inline inline functions, msvc does not IIRC)

the-phinet: It's mainly changed to not rely on a macro in a random 3rd party header file. We have LTO enabled now, so even without a `__forceinline` for msvc, I think we're fine. If profiling shows this to be an issue, a targeted fix would make sense. In general, I think it makes a lot of sense to move towards standard c++ stuff, avoiding compiler-specific fixes.

River-Salmon: That's fine with me, as long as the difference has been considered. Sounds like it has

TehRealSalt (approved):

finn-ball:
At the moment you have third party libraries in two different locations (`libraries` and `src/common/thirdparty`). This PR would introduce a third location for these artifacts. 

Would it be better to move some of the license files such as `docs/licenses/asmjit.txt` and `docs/licenses/range_map.txt` to be in the same folder as where the sourcecode is defined? They could be simply renamed `LICENCE.txt` as they are defined in their own folders.

If multiple libraries occupy the same folder, perhaps you could namespace them like: `superfasthash.LICENCE.txt`. 

Ideally in the future, you may want one top level folder such as `thirdparty/`

the-phinet:
> At the moment you have third party libraries in two different locations (libraries and src/common/thirdparty). This PR would introduce a third location for these artifacts.

The purpose of the `docs/licenses` folder is to generate the licenses.zip file for distribution, so having the files in a centralized location is unfortunately required. I do plan on eventually restructuring the folder (and probably generating a single text file for the purpose of embedding the licenses in the about section of the launcher), so this is just a temporary solution.

> Ideally in the future, you may want one top level folder such as thirdparty/

Yeah, agreed. I do plan on untangling (and un-modifying) libraries we have been using, in order to make updating them easier.

PR #1349, opened 2026-05-03
Don't use haptics while on keyboard

### LINKED ISSUE #548: Limit haptic feedback to only actively used gamepads
### Game

All

### Problem Description

_No response_

### Feature Description

When playing mkb, controller still rumbles

### Additional Media

Related to #142

TehRealSalt:
Resolves #548 

WIP, I was struggling with Windows support because it has quite a few different branches for the different gamepad types.

## TODO

- [ ] Finish platform support
  - [ ] Windows (full support)
  - [ ] Mac (stub, it does not support haptics)
- [ ] On the haptics' code's end, measure how long it has been since the non-keyboard input
- [ ] Don't blast a call to `I_Rumble(0, 0, 0, 0)` every frame while using keyboard; only do it when necessary

## Checklist

### DIFF
--- a/src/common/console/keydef.h
+++ b/src/common/console/keydef.h
@@ -243,3 +243,11 @@ static const int KeyAxisMapping[NUM_AXIS_CODES] = {
 	KEY_PAD_RTRIGGER,
 };
 
+
+enum EKeyDeviceTypes
+{
+	DEVICE_KBM, // Keyboard / mouse
+	DEVICE_JOY, // DInput Joystick
+	DEVICE_PAD, // XInput Gamepad
+	DEVICE__MAX
+};
--- a/src/common/engine/m_haptics.cpp
+++ b/src/common/engine/m_haptics.cpp
@@ -22,6 +22,7 @@
 #include "c_dispatch.h"
 #include "doomdef.h"
 #include "doomstat.h"
+#include "i_system.h"
 #include "m_haptics.h"
 #include "name.h"
 #include "printf.h"
@@ -441,6 +442,17 @@ void Joy_ReadyRumbleMapping()
 void Joy_RumbleTick()
 {
 	if (!Haptics.enabled) return;
+
+	if (I_GetLastUsedDevice() == DEVICE_KBM)
+	{
+		// no haptics while using keyboard + mouse
+
+		// TODO: probably save the device and only
+		// do this on value change, instead of
+		// blasting every frame
+		I_Rumble(0, 0, 0, 0);
+		return;
+	}
 
 	// pause detection
 	if (AppActive != Haptics.active)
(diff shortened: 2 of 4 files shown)


arrowgent:
but what if you have a keyboard... that has haptic feedback on the keys?!

although i would agree, no need to process a Rumble feature if there is no haptic connected device (or its disabled in menus)


or crazy idea, force-feedback keys
when you run into a wall, the keyboard button PUSHES BACK

kcat:
How would this work with analog keyboards, which simulate analog keys as a gamepad device (like analog triggers on modern console controllers), while the digital keys are normal keyboard device keys? Changing behavior based on the selected or last used input device type can easily run into problems with devices that mix input types.

RicardoLuis0:
> How would this work with analog keyboards, which simulate analog keys as a gamepad device (like analog triggers on modern console controllers), while the digital keys are normal keyboard device keys? Changing behavior based on the selected or last used input device type can easily run into problems with devices that mix input types.

haptics could be changed to only be enabled on actively used controllers, that way only the emulated keyboard device would get them, and not any real controllers

the-phinet:
> haptics could be changed to only be enabled on actively used controllers, that way only the emulated keyboard device would get them, and not any real controllers

I think this is reasonable, and also what is planned anyways.

Unfortunatly with obscure hardware, there's only so much we can do to support them. If someone wants to donate an analogue keyboard, I'm sure we can figure out a good solution.

RicardoLuis0:
> > haptics could be changed to only be enabled on actively used controllers, that way only the emulated keyboard device would get them, and not any real controllers
> 
> I think this is reasonable, and also what is planned anyways.
> 
> Unfortunatly with obscure hardware, there's only so much we can do to support them. If someone wants to donate an analogue keyboard, I'm sure we can figure out a good solution.

if they're just emulating a controller it shouldn't act any different from just testing with two separate (regular) controllers connected

kcat:
> if they're just emulating a controller it shouldn't act any different from just testing with two separate (regular) controllers connected

Yes, it's basically just like playing with a (mouse +) keyboard + gamepad simultaneously. Some keys correspond to a gamepad device's axis events (the ones feeding analog presses) and others may correspond to gamepad buttons or traditional keyboard keys (what does what is configurable via profiles for the device). And of course it can be used alongside a mouse, like a normal keyboard+mouse setup just with some keyboard keys acting like a gamepad. There's probably better videos to explain it, but this is the one I most clearly remember that does a fine job:

https://www.youtube.com/watch?v=k9f5bOnxBtk

The biggest compatibility issues it has is with games that disable mouse input after detecting gamepad input (it can be used as a fully normal keyboard without gamepad events of course, but that kinda defeats the purpose of using an analog keyboard). Though I have also seen videos by him where he's obviously using the keyboard because the game would constantly flicker between showing gamepad UI elements and kb+mouse UI elements on screen depending on the keys pressed or if the mouse was used for a given frame; playable, but visually distracting. So it would seem the best way to be compatible would be to not think in terms of a gamepad *or* kb+mouse, and instead account for a kb+mouse+gamepad type setup.

PR #1350, opened 2026-05-04
Automap SectorFillColor + SectorFillAlpha

TehRealSalt:
Continued from @XaserAcheron's work on #456

image
image

## Changes from #456

- Allow the fill color in overlay mode
- Added the options to the menus
- Added sliders to determine the fill color's alpha
  - Needed to make fill color reasonable in overlay mode
  - But also, just looks nice on textured backgrounds (like the Raven games)
  - Removes the need for the sentinel color for undefined color, turn it off simply by using alpha 0.

Alpha can be expanded to use for textured mode, but I didn't want to reuse the cvar because it might be confusing, and I wasn't sure about complicating this branch any further by adding more cvars.

## Checklist


src/am_map.cpp
```diff
@@ -311,6 +313,8 @@ CVAR (Color, am_ovthingcolor_monster,	0xe88800,	CVAR_ARCHIVE);
 CVAR (Color, am_ovthingcolor_ncmonster,	0xe88800,	CVAR_ARCHIVE);
 CVAR (Color, am_ovthingcolor_item,		0xe88800,	CVAR_ARCHIVE);
 CVAR (Color, am_ovthingcolor_citem,		0xe88800,	CVAR_ARCHIVE);
+CVAR (Color, am_ovsectorfillcolor,		0x000000,	CVAR_ARCHIVE);
+CVAR (Float, am_ovsectorfillalpha,		0.0f,		CVAR_ARCHIVE);
```
the-phinet: This is awesome! My only feedback is that setting this to 0.2 would be a better default

the-phinet (dismissed):

src/am_map.cpp
TehRealSalt: I was thinking to not rock the boat, but fair! Made a commit with defaults for both normal+overlay mode

RicardoLuis0 (approved):

Issue #311, opened 2025-10-31
Add more spectre fuzz effects (REFRACTION)


Sol1vaN:
Hi there! 😃
Well, taking advantage of the historic momentum:

Can you add this spectacular effect?
REFRACTION for Fuzz transparency
<img width="2556" height="1440" alt="Image" src="https://github.com/user-attachments/assets/c393d470-c9be-485f-b173-33e1a66fbb81" />

hope you can 🙏

### Feature Description

Only want to add this to the actual options, I mean:
- Pixel fuzz
- Smooth fuzz
- Swirly fuzz
- Translucent fuzz
- Noise
- Smooth noise
- Software
- Shadow
- Refraction fuzz <-- new one

### Additional Media

Nugget example
<img width="2556" height="1440" alt="Image" src="https://github.com/user-attachments/assets/82dc0d20-e2e9-4eb4-ae94-516cb3de5e47" />

GZDoom example
<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/b8c2fd97-8e4f-44f5-8ba4-93da5a92d07e" />

Cacodemon345:
My understanding is that doing it with UZDoom's renderer is not really possible without significantly raising GPU requirements and causing stability issues on AMD hardware, as it needs Vulkan's interlocking extensions whose support isn't quite present. Even the alternate extensions just for this on Qualcomm and ARM GPUs are hit-or-miss.

Sol1vaN:
> My understanding is that doing it with UZDoom's renderer is not really possible without significantly raising GPU requirements and causing stability issues on AMD hardware, as it needs Vulkan's interlocking extensions whose support isn't quite present. Even the alternate extensions just for this on Qualcomm and ARM GPUs are hit-or-miss.

man, we talking about a little effect, for today, 2025, where theres a enough gpu power, even for old gpus, this effect will not impact considerably i think.

and i pointed to add this option on uzdoom, not for replace the current ones.
if someone have performance problems, simply they can change the effect for other one.

hope devs can consider this, is a nice effect really

PR #1355, opened 2026-05-05
Expose SceneColor as a bindable input for material HardwareShaders with per-draw on-demand grab

vortexdesign:
I'm building a weapons gameplay mod and I wanted a shockwave for my grenade and didn't want to go down the sprite route or use the CameraTextures workaround since I'm trying to mirror the look of a game in the unreal 4 engine. So after some research it appears I've stumbled onto an issue that's been talked about for decades and mentioned by Nash several times in the forums which is that currently the engine doesn't expose the SceneColor to create effects like true refraction from the player's perspective.

This implementation does exactly that and though it does touch a bunch of files it doesn't actually even add that much code, I think it was just a case of folks not wanting to do the work since the workarounds were serviceable because as it turns out the GrabPass pattern I went with was first proposed by Xabis all the way back in 2017.

**Screenshot Examples:**
Screenshot 2026-05-05 141953
Screenshot 2026-05-05 142547

**Some Technical Details:**
- **`SceneColor`** gets auto-bound to every material HardwareShader, exposing a sampleable copy of the scene framebuffer for spatial effects like refraction.
- **Per-draw on-demand grab** triggers when a material with the `samplesscene` flag draws and something non-refractive has been drawn since the last grab (dedup). This correctly captures sprites and actors into the grab so refractive surfaces can see them. Same explicit declaration pattern as Unity GrabPass and Unreal SceneColor node.
- **New GLDEFS keyword `samplesscene`** on HardwareShader bindings tells the engine which materials should trigger grabs.
- Both backends grab scene color into a dedicated background texture (see diff for backend-specific details: `glBlitFramebuffer` on GL, `vkCmdBlitImage` with layout transitions on Vulkan). Vulkan side bumps `SHADER_MIN_REQUIRED_TEXTURE_LAYERS` from 11 to 13 to add SceneColor at descriptor binding 12.
- Tested locally on NVIDIA RTX 5080 with Windows 11 on both Vulkan and OpenGL backends.

**GLDEFS Example:**
```
HardwareShader "MYTEXTURE"
{
Shader "shaders/refraction.fp" 330
samplesscene
}
```

**Some Caveats:**
- The GLES backend is out of scope. It uses renderbuffers for the scene buffers and would need a much bigger refactor.
- No refraction-of-refraction. Refractive surfaces read scene state from before they draw, so they can't sample other refractive surfaces drawn after them.
- Materials with `samplesscene` applied to opaque-pass items (e.g. 3D models marked solid) would see stale grab data since the per-draw check only fires during the translucent pass. Same architectural restriction as Unreal's SceneColor node which is disallowed in opaque materials.

**Historical References:**
- [Topic #64103](https://forum.zdoom.org/viewtopic.php?t=64103) - Nash - March 2019 - "Things that are not available based on what I see from that COD video: Depth-based water fade (requires depth buffer on material shaders); Localized distortion (rotating submarine blades): requires access to scene buffer in material shaders"
- [Topic #1122947](https://forum.zdoom.org/viewtopic.php?p=1122947) - Nash - October 2019 - "Material textures do not have access to the current screen colour and depth, so no that's currently not possible. Those two things are needed to do any effects that seemingly affect what is 'behind' a texture — for example, refraction behind a glass or water, or 'heat haze'."
- [Topic #57385](https://forum.zdoom.org/viewtopic.php?t=57385) - Xabis - October 2017 - "Seems a simple approach to allow distorting the scene would be by doing something like unity's GrabPass, which simply copies the underlying pixels to a separate texture that can be accessed by the shader. Since it can be expensive to capture, it should be opt-in. For textures/flats, it could be another config param in GLDEFS."

**Demo PK3 ZIP:**
- I've attached a demo zip file that adds a shockwave to the ordinary rocket launcher using a refraction shader. The bubble correctly refracts walls, sprites, and gore behind it.
- Just do uzdoom.exe -iwad doom2.wad -file SceneColorDemo and fire a rocket
- Download [SceneColorDemo.zip](https://github.com/user-attachments/files/27609468/SceneColorDemo.zip)

**Follow up PR:**
I was also thinking about doing a follow up PR once this one is merged that would directly address [Issue #311](https://github.com/UZDoom/UZDoom/issues/311) and create a Refraction fuzz option.

**AI Disclaimer:**
No code was generated or copy-pasted for this PR, however I did use Claude for initial discovery and diagnosis, code review and suggestions and helping me with some of the math for getting the refraction shader to look right.

## Checklist

### DIFF
--- a/src/common/rendering/gl/gl_framebuffer.cpp
+++ b/src/common/rendering/gl/gl_framebuffer.cpp
@@ -405,6 +405,15 @@ void OpenGLFrameBuffer::AmbientOccludeScene(float m5)
 	gl_RenderState.Apply();
 }
 
+// Blit current scene color to the sampleable background texture and bind to unit 18.
+void OpenGLFrameBuffer::GrabSceneColor()
+{
+	if (!GLRenderer || !GLRenderer->mBuffers)
+		return;
+	GLRenderer->mBuffers->GrabSceneColor();
+	GLRenderer->mBuffers->BindSceneColorBackgroundTexture(18);
+}
+
 void OpenGLFrameBuffer::FirstEye()
 {
 	GLRenderer->mBuffers->CurrentEye() = 0;  // always begin at zero, in case eye count changed
--- a/src/common/rendering/gl/gl_framebuffer.h
+++ b/src/common/rendering/gl/gl_framebuffer.h
@@ -51,6 +51,7 @@ public:
 	void Update() override;
 
 	void AmbientOccludeScene(float m5) override;
+	void GrabSceneColor() override;
 	void FirstEye() override;
 	void NextEye(int eyecount) override;
 	void SetSceneRenderTarget(bool useSSAO) override;
--- a/src/common/rendering/gl/gl_renderbuffers.cpp
+++ b/src/common/rendering/gl/gl_renderbuffers.cpp
@@ -70,6 +70,8 @@ void FGLRenderBuffers::ClearScene()
 {
 	DeleteFrameBuffer(mSceneFB);
 	DeleteFrameBuffer(mSceneDataFB);
+	DeleteFrameBuffer(mSceneColorBackgroundFB);
+	DeleteTexture(mSceneColorBackgroundTex);
 	if (mSceneUsesTextures)
 	{
 		DeleteTexture(mSceneMultisampleTex);
@@ -228,6 +230,10 @@ void FGLRenderBuffers::CreateScene(int width, int height, int samples, bool need
 			mSceneDataFB = CreateFrameBuffer("SceneGBufferFB", mPipelineTexture[0], mSceneDepthStencilBuf);
 		}
 	}
+
+	// SceneColor grab-pass background, RGBA16F single-sample.
+	mSceneColorBackgroundTex = Create2DTexture("SceneColorBackground", GL_RGBA16F, width, height);
+	mSceneColorBackgroundFB = CreateFrameBuffer("SceneColorBackgroundFB", mSceneColorBackgroundTex);
 }
 
 //==========================================================================
@@ -523,6 +529,34 @@ void FGLRenderBuffers::BlitSceneToTexture()
 
 	glBindFramebuffer(GL_READ_FRAMEBUFFER, 0);
 	glBindFramebuffer(GL_DRAW_FRAMEBUFFER, 0);
+}
+
+//==========================================================================
+//
+// Grab pass: blit scene color to the sampleable background texture (bindings restored to mSceneFB).
+//
+//==========================================================================
+
+void FGLRenderBuffers::GrabSceneColor()
+{
+	if (!mSceneColorBackgroundFB || !mSceneFB)
+		return;
+
+	glBindFramebuffer(GL_READ_FRAMEBUFFER, mSceneFB.handle);
+	glBindFramebuffer(GL_DRAW_FRAMEBUFFER, mSceneColorBackgroundFB.handle);
+	glBlitFramebuffer(0, 0, mWidth, mHeight, 0, 0, mWidth, mHeight, GL_COLOR_BUFFER_BIT, mSamples > 1 ? GL_LINEAR : GL_NEAREST);
+
+	glBindFramebuffer(GL_READ_FRAMEBUFFER, mSceneFB.handle);
+	glBindFramebuffer(GL_DRAW_FRAMEBUFFER, mSceneFB.handle);
+}
+
+void FGLRenderBuffers::BindSceneColorBackgroundTexture(int index)
+{
+	if (!mSceneColorBackgroundTex)
+		return;
+	glActiveTexture(GL_TEXTURE0 + index);
+	glBindTexture(GL_TEXTURE_2D, mSceneColorBackgroundTex.handle);
+	glActiveTexture(GL_TEXTURE0);
 }
 
 //==========================================================================
--- a/src/common/rendering/gl/gl_renderbuffers.h
+++ b/src/common/rendering/gl/gl_renderbuffers.h
@@ -138,6 +138,10 @@ public:
 	void BindSceneDepthTexture(int index);
 	void BlitSceneToTexture();
 
+	// Grab pass: copy scene color into a sampleable texture before translucent.
+	void GrabSceneColor();
+	void BindSceneColorBackgroundTexture(int index);
+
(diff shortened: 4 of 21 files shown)


RicardoLuis0:
exposing depth won't work with portals, and to make it work with portals would lead to users being able to make shaders that can be way too expensive depending entirely on which map is being played (because it'd need to run once per portal), so this is a no from me unfortunately

RicardoLuis0:
also grabbing the scene color has already been implemented by https://github.com/UZDoom/UZDoom/pull/704

vortexdesign:
@RicardoLuis0 On PR #704 I checked and #704 (LastInputTexture) is a different mechanism as it exposes the previous frame's scene output to PostProcess shaders only for temporal effects (TAA, auto-exposure, motion blur per the PR body) while this PR exposes the current frame's scene framebuffer (captured before the translucent pass) to material HardwareShaders for spatial effects like refraction. Material shaders cannot read LastInputTexture or any other PP output so #704 doesn't help with material-shader refraction which is why I think this PR is still needed even though #704 is already merged.

On SceneDepth + portals yah fair point, looking at the existing SSAO implementation though in hw_drawinfo.cpp the engine does have an established pattern for managing expensive depth-based effects in portals: 
- gl_ssao_portals CVAR
- per-portal AllowSSAO() opt-in
- a counter that decrements per portal

So SceneDepth could adopt the same pattern and the opt-in r_scene_depth_sampler CVAR already addresses the default case (zero cost when off), and per-portal gating could be added in this PR or as a follow-up if needed. 

Let me know your thoughts.

RicardoLuis0:
> @RicardoLuis0 On PR #704 I checked and #704 (LastInputTexture) is a different mechanism as it exposes the previous frame's scene output to PostProcess shaders only for temporal effects (TAA, auto-exposure, motion blur per the PR body) while this PR exposes the current frame's scene framebuffer (captured before the translucent pass) to material HardwareShaders for spatial effects like refraction. Material shaders cannot read LastInputTexture or any other PP output so #704 doesn't help with material-shader refraction which is why I think this PR is still needed even though #704 is already merged.

i see, SceneColor would be mergeable then, though opt-in is also a no-go entirely for anything shader related, because it'd break shader compilation

> * gl_ssao_portals CVAR
> * per-portal AllowSSAO() opt-in
> * a counter that decrements per portal

SSAO in portals gets _disabled_ for portals past the limit (and AllowSSAO has nothing to do with opt-in, it's used to disable SSAO in skyboxes), but SSAO _as an effect_ doesn't break if it's only applied partially, other effects like DOF, and many others that users would want to implement would entirely break visually if they're only applied to closer parts of the scene

vortexdesign:
> i see, SceneColor would be mergeable then, though opt-in is also a no-go entirely for anything shader related, because it'd break shader compilation

On SceneColor side the uniform is declared unconditionally in the prefix so only the buffer allocation and grab pass are CVAR-gated, shaders compile fine but for SceneDepth though yah the prefix declaration is gated which breaks compilation so opt-in doesn't work there, damn.

> SSAO in portals gets disabled for portals past the limit (and AllowSSAO has nothing to do with opt-in, it's used to disable SSAO in skyboxes), but SSAO as an effect doesn't break if it's only applied partially, other effects like DOF, and many others that users would want to implement would entirely break visually if they're only applied to closer parts of the scene

Ah ok fair points here my bad on the characterization, so it looks like then the easiest path is probably just to drop SceneDepth from this PR and just ship SceneColor, I'll have to think through SceneDepth much more before coming up with a solution there.

vortexdesign:
Updated per review feedback:
- Dropped SceneDepth from this PR as the design needs way more thought
- Removed the opt-in CVAR (`r_scene_color_sampler`) so SceneColor allocation and grab pass are now always-on
- Updated PR title and description to match
- Re-uploaded the demo PK3 (no longer needs the CVAR set)

Please let me know if there's anything else you guys need for this one, thanks.

RicardoLuis0:
wait, why is scene color being passed to post-process shaders? they _already_ have a scene color input

RicardoLuis0:
(also, don't keep the old stuff as a commit, that just makes reviewing (and in the future git bisecting) more complicated, each commit should ideally only be relevant to the features actually being merged, not to any that were removed)

vortexdesign:
> wait, why is scene color being passed to post-process shaders? they already have a scene color input

It's not, PP shaders don't get it only material shaders do. PP shader compile doesn't touch either of the places SceneColor is declared, vk_postprocess.cpp is just where GrabSceneColor lives next to the other blit functions and is called from hw_drawinfo.cpp not from a PP pass.

> (also, don't keep the old stuff as a commit, that just makes reviewing (and in the future git bisecting) more complicated, each commit should ideally only be relevant to the features actually being merged, not to any that were removed)

Understood. I also just realized I forgot to remove some leftover includes from the SceneDepth stuff so I'll roll all that into one clean squashed commit on the next push.

vortexdesign:
The squash is in and the leftover includes have been removed so everything is ready for review in a single commit now, let me know if there's anything else needed on my end.

vortexdesign:
Updated to address the actor/sprite issue Nash raised:

- Per-draw on-demand grab with dedup logic, sprites and actors are now correctly captured so refraction works through them
- New samplesscene GLDEFS keyword for material opt-in (same explicit declaration pattern as Unity GrabPass and Unreal SceneColor node)
- Re-uploaded demo PK3 with the keyword added to GLDEFS
- Tested on both Vulkan and OpenGL
- PR title and description updated accordingly

Please let me know if you guys have any questions.

RicardoLuis0:
thinking back to this again, exposing the background color to shaders would make it impossible to do order-independent transparency, which is something that **is** planned for the future, so i think this will be unmergeable unfortunately

vortexdesign:
hmm I don't think OIT actually conflicts with this here cause like refraction has to sample the composited scene at other pixels so it doesn't happen inside a resolve, pretty sure engines that expose scene color pin the read to a defined boundary instead and that stays valid no matter how the transparents themselves get composited, for example HDRP does it with the Before Refraction vs Default pass split and unreal has shipped SceneColor alongside its experimental OIT since 5.1.

So basically samplesscene already gives us that partition here and the only timing-aware code is the check in DoDraw, the per-draw grab is just what today's sorted pass needs, under an OIT it'd collapse to one grab at the resolve boundary with nothing mod-facing changing. 

Did you have an OIT approach planned out yet or anything cause I'm happy to move the grab boundary to fit whatever it is or otherwise accommodate it in anyway I can.

PR #1369, opened 2026-05-10
Apply LOOKALLAROUND to enemy+TID search

### LINKED ISSUE #796: A_LookEX Acquires Target Outside of FOV
### UZDoom Version

UZDoom 4.14.3

### Game

All

### Operating System

Windows 11

### OS Version

_No response_

### Hardware

_No response_

### Duplicate Confirmation

### Bug Description

A_LookEX still acquires a target outside its FOV, as if +LookAllAround is used.  This appears to have been broken in GZDoom 4.10 and hasn't been fixed correctly.  Please see the attached demo, where each beam from a BFG projectile all fire at the same target, whether the target is inside the declared FOV or not (60 degrees is used for the demo).

### Replication

[bfg10kTest.zip](https://github.com/user-attachments/files/24274105/bfg10kTest.zip)

Using the BFG, fire into an area with multiple monsters.  Observed behavior shows six beams all targeting the same monster.  Correct behavior is to only fire arcs at targets found within their FOV of 60 degrees, spawning multiple arcs in different directions instead of all six arcs firing at one.

### Configuration

```ini

```

### Log

_No response_

TehRealSalt:
Should resolve #796.

I narrowed this down to ZDoom/GZDoom#2433. The follow-up at ZDoom/GZDoom#2547 suggests that the solution is to apply the `LOOKALLAROUND` flag instead of `true` in these cases.

Neither of these pull requests have any additional details or example mods that needed this behavior changed, so beyond just the example mod in #796, I was not sure what to test. If this change has any adverse side effects, then the pull requests mentioned earlier would need reverted instead.

## Checklist

### DIFF
--- a/src/playsim/p_enemy.cpp
+++ b/src/playsim/p_enemy.cpp
@@ -1401,7 +1401,7 @@ bool ValidEnemyInBlock(AActor* lookee, AActor* other, void* lookparams)
 
 	// [KS] Hey, shouldn't there be a check for MF3_NOSIGHTCHECK here?
 
-	if (!keepChecking || !P_IsVisible(lookee, other, true, params))
+	if (!keepChecking || !P_IsVisible(lookee, other, (lookee->flags4 & MF4_LOOKALLAROUND), params))
 		return false;			// out of sight
 
 	return true;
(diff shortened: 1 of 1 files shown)


Boondorl (approved):

HexaDoken:
> so beyond just the example mod in #796, I was not sure what to test.

The biggest mod I know that was broken by this was Xaser's Argent - the BFG set to "modern" mode in it would behave extremely strangely precisely because of this issue.

I'll be sure to take a look myself at it once autobuilds catch up.

TehRealSalt:
> The biggest mod I know that was broken by this was Xaser's Argent - the BFG set to "modern" mode in it would behave extremely strangely precisely because of this issue.

I was more asking about what mods might have inspired the original change that introduced the regression in the first place, not what else is affected by this issue.

I'm pretty sure this fixes everything, but I still feel like I am missing context :p

PR #1372, opened 2026-05-11
Added changeepisode command

Boondorl:
Allows switching to maps via episode numbers instead of needing to know the actual map name for the start of an episode. As a bonus, this also allows passing -1 as the position in `changemap`, making going to a map with pistol starts possible in multiplayer (`map` is disabled).

## Checklist

### DIFF
--- a/src/console/c_cmds.cpp
+++ b/src/console/c_cmds.cpp
@@ -330,7 +330,7 @@ CCMD (changemap)
 {
 	if (!players[consoleplayer].mo || !usergame)
 	{
-		Printf ("Use the map command when not in a game.\n");
+		Printf ("Cannot use changemap command when not in a game. Use map instead\n");
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


src/console/c_cmds.cpp
```diff
@@ -330,7 +330,7 @@ CCMD (changemap)
 {
 	if (!players[consoleplayer].mo || !usergame)
 	{
-		Printf ("Use the map command when not in a game.\n");
+		Printf ("Cannot use map command when not in a game.\n");
```
TehRealSalt: Nitpick; feel free to ignore

I think the original message was trying to inform the user that they should use `map` in this instance, not just tell them `changemap` doesn't work. It was certainly vague though.

Maybe something like `Cannot use "changemap" outside of a game. Use "map" instead.` would be the best of both?

TehRealSalt (dismissed):
Works great, the one changed message looks weird to me but it's very minor

src/console/c_cmds.cpp
Boondorl: I'm not gonna lie, I totally thought it was just horrible grammar but you're probably correct lmao. I'll fix it soon

the-phinet (approved):

TehRealSalt (approved):

PR #1383, opened 2026-05-17
Update zdoom-info.plist

CJM7:
Just added a line to enable OSX's built-in compatibility mode to accommodate the camera notch

## Checklist

### DIFF
--- a/src/posix/osx/zdoom-info.plist
+++ b/src/posix/osx/zdoom-info.plist
@@ -50,5 +50,7 @@
 	<string>NSApplication</string>
 	<key>NSSupportsAutomaticGraphicsSwitching</key>
 	<true/>
+	<key>NSPrefersDisplaySafeAreaCompatibilityMode</key>
+    <true/>
 </dict>
 </plist>


the-phinet:
I have the opportunity to test this later today :)

the-phinet:
Nice, this works well, thank you. Here are some photos of what the change does:

| Before | After |
|---|---|
| PXL_20260525_150848801 MP-outPXL_20260525_150853355-outPXL_20260525_150926208-out | PXL_20260525_150945256 MP-out PXL_20260525_150957873-out PXL_20260525_151023260-out |

While it does solve the issue where text is cut-off, I don't think this should be the long-term solution. The image quality is noticeable reduced, which is a shame on such a nice screen.

The person I was borrowing the laptop from to test suggested a way I can iterate on a deeper solution without owning the required hardware. I should be able to work on this eventually, if someone else doesn't beat me to it

the-phinet (approved):

RicardoLuis0:
> While it does solve the issue where text is cut-off, I don't think this should be the long-term solution. The image quality is noticeable reduced, which is a shame on such a nice screen.

might be possible to exclude the notch area from any UI drawing, while keeping it as valid for non-ui? (ex. game world, titlepic, etc)

the-phinet:
> might be possible to exclude the notch area from any UI drawing, while keeping it as valid for non-ui? (ex. game world, titlepic, etc)

That's exactly the plan. I'm hoping for 3 levels of notch-y-ness:

1) just make a black bar, pretend the computer has a big bezel
2) only shift uzdoom text out of the way (the console, optionmenu, and notifications). This would prevent huds from breaking, while allowing for most text to always be visable.
3) shift all ui stuffs, like your suggestion. This has the potential of breaking some huds, but will work well in most cases.

There are some cursory talks for notch things in a future sbardef spec. If those get in, I would apply them at level 2

PR #1396, opened 2026-05-23
Fix ThickFog in Skybox when rendered in other portals 

### LINKED ISSUE #1375: New ThickFog renders incorrectly with sector portals + 3D skyboxes
### UZDoom Version

5.0.0-pre240-g7605cd089f

### Game

All

### Operating System

Windows 11

### OS Version

Windows 11 Pro 10.0.26100 Build 26100

### Hardware

13th Gen Intel(R) Core(TM) i9-13900HX, GeForce RTX 4060 Ti, 32GB memory

### Duplicate Confirmation

### Bug Description

The new ThickFog in UZDoom v5.0 (https://github.com/UZDoom/UZDoom/commit/c6a6ae23a6209467c6417c38230aa25673d44ca5) renders incorrectly on maps with:
- interactive sector portals
- where both sectors have skies and Line Horizons
- and the map has a 3D skybox

The map renders correctly when using normal fog. When using ThickFog, the horizon sky in one of the two sectors will be rendered incorrectly as pure black.

This bug has different permutations depending on the map, the number of portals, etc., but they all seem to have the same effect: the sky renders incorrectly when >2 scenes are being rendered (portals, skybox, etc.). My guess would be that the issue is in [`hw_portal.cpp`](https://github.com/UZDoom/UZDoom/blob/trunk/src/rendering/hwrenderer/scene/hw_portal.cpp) or [`hw_portal.h`](https://github.com/UZDoom/UZDoom/blob/trunk/src/rendering/hwrenderer/scene/hw_portal.h), which seem to handle how skies render in the context of portals and occasionally reference ThickFogDistance.

### Replication

See the sample zip. [ThickFogDistance test.zip](https://github.com/user-attachments/files/27735782/ThickFogDistance.test.zip)

These two maps have interactive sector portals set with Set Sector Portal, as well as a 3D skybox set with a Skybox Viewpoint. The 3D skybox is just a box with brick textures.

MAP01 and MAP02 are identical. MAP01 does not have ThickFog enabled in MAPINFO, while MAP02 does.

**MAP01 No ThickFog**
Image

**MAP02 ThickFog**
Image

You can also achieve the same ThickFog effect with the console command `ThickFogDistance [any positive number]` on MAP01.

dileepvr:
Fixing `ThickFog` calculations in `Skybox` portals recursively rendered in other portals, like stacked sector portals and line mirrors.

Moved `ThickFog` distance and density multiplier modifications (by factor of 16) out of the `RenderFirstSkyPortal()` function, and into the specific `RenderPortal()` functions. The `RenderFirstSkyPortal()` optimization is skipped when recursion is active and the depth buffer is needed.

Closes #1375.

## Checklist

### DIFF
--- a/src/rendering/hwrenderer/scene/hw_portal.cpp
+++ b/src/rendering/hwrenderer/scene/hw_portal.cpp
@@ -157,13 +157,13 @@ bool FPortalSceneState::RenderFirstSkyPortal(int recursion, HWDrawInfo *outer_di
 
 void FPortalSceneState::RenderPortal(HWPortal *p, FRenderState &state, bool usestencil, HWDrawInfo *outer_di)
 {
-	if (outer_di->Viewpoint.bDoOrtho && (strcmp(p->GetName(), "Sky") == 0))
+	if (outer_di->Viewpoint.bDoOrtho && p->IsRealSky())
 	{
 		tempmatrix = outer_di->VPUniforms.mProjectionMatrix; // ensure perspective projection matrix for skies
 		outer_di->VPUniforms.mProjectionMatrix = outer_di->ProjectionMatrix2;
 	}
 	if (gl_portals) outer_di->RenderPortal(p, state, usestencil);
-	if (outer_di->Viewpoint.bDoOrtho && (strcmp(p->GetName(), "Sky") == 0)) outer_di->VPUniforms.mProjectionMatrix = tempmatrix;
+	if (outer_di->Viewpoint.bDoOrtho && p->IsRealSky()) outer_di->VPUniforms.mProjectionMatrix = tempmatrix;
 }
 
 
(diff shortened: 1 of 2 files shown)


src/rendering/hwrenderer/scene/hw_portal.cpp
```diff
@@ -168,7 +157,13 @@ bool FPortalSceneState::RenderFirstSkyPortal(int recursion, HWDrawInfo *outer_di
 
 void FPortalSceneState::RenderPortal(HWPortal *p, FRenderState &state, bool usestencil, HWDrawInfo *outer_di)
 {
+	if (outer_di->Viewpoint.bDoOrtho && (strcmp(p->GetName(), "Sky") == 0))
```
RicardoLuis0: wait, i missed it in the original review of the thickfog PR, but why is the render code using strcmp/GetName in the first place? if it needs to differentiate between sky/skybox/regular portals should just add a new virtual for that to HWPortal

dileepvr: It's hard because I need to update `VPUniforms`, which only happens via a call to `StartDrawInfo()`  from inside `HWDrawInfo::RenderPortal()`.

I wish I could've relied on `bool HWPortal::IsSky()` but stacked sector portals also return true for that. Should I create a `bool HWPortal::IsSkyDome()` method?

RicardoLuis0: > Should I create a `bool HWPortal::IsSkyDome()` method?

no need to make it that specific, better to have a virtual that returns an enum with the same "values" as GetName, that way if any other type of portal needs differentiating it'll be possible as well

dileepvr: Gah. Just as I pushed.

I'll create an enum.

dileepvr: Done.

theshadsy:
I can confirm that this fixes the issue! Thank you!

RicardoLuis0 (approved):

PR #1397, opened 2026-05-24
Let all canvas textures be accessed via ZScript

AFADoomer:
Currently any canvas texture that shares a name with a non-wall image (e.g., a sprite) cannot be accessed via ZScript because the canvas gets assigned to that image's namespace.  

This change allows canvas textures to be looked up in any namespace.

## Checklist

### DIFF
--- a/src/common/2d/v_2ddrawer.cpp
+++ b/src/common/2d/v_2ddrawer.cpp
@@ -1233,9 +1233,9 @@ public:
 	}
 };
 
-FCanvas* GetTextureCanvas(const FString& texturename)
+FCanvas *GetTextureCanvas(const FString &texturename, ETextureType usetype = ETextureType::Wall, BITFIELD flags = 0)
 {
-	FTextureID textureid = TexMan.CheckForTexture(texturename.GetChars(), ETextureType::Wall, FTextureManager::TEXMAN_Overridable | FTextureManager::TEXMAN_TryAny);
+	FTextureID textureid = TexMan.CheckForTexture(texturename.GetChars(), usetype, FTextureManager::TEXMAN_Overridable | flags);
 	if (textureid.isValid())
 	{
 		// Only proceed if the texture is a canvas texture.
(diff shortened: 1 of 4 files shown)


src/common/2d/v_2ddrawer.cpp
```diff
@@ -1235,7 +1235,7 @@ class InitTextureCanvasGC
 
 FCanvas* GetTextureCanvas(const FString& texturename)
 {
-	FTextureID textureid = TexMan.CheckForTexture(texturename.GetChars(), ETextureType::Wall, FTextureManager::TEXMAN_Overridable);
+	FTextureID textureid = TexMan.CheckForTexture(texturename.GetChars(), ETextureType::Wall, FTextureManager::TEXMAN_Overridable | FTextureManager::TEXMAN_TryAny);
```
RicardoLuis0: this will hide sprites that have the same names as walls, a better fix would be to add the texture types/flags parameters to GetTextureCanvas itself

RicardoLuis0 (changes requested):

src/common/2d/v_2ddrawer.cpp
AFADoomer: Just updated with those changes made.

RicardoLuis0 (dismissed):

src/gamedata/textures/animations.cpp
```diff
@@ -731,6 +731,7 @@ void FTextureAnimator::ParseCameraTexture(FScanner &sc)
 		fitwidth = oldtex->GetDisplayWidth ();
 		fitheight = oldtex->GetDisplayHeight ();
 		viewer->SetUseType(oldtex->GetUseType());
+		viewer->SetOffsets(oldtex->GetDisplayLeftOffset(), oldtex->GetDisplayTopOffset());
```
RicardoLuis0: changing existing behavior should be avoided - this needs to be gated behind a new flag, ex.:
```cpp
if (sc.GetString())
{
	if (sc.Compare ("useoffsets"))
	{
		viewer->SetOffsets(oldtex->GetDisplayLeftOffset(), oldtex->GetDisplayTopOffset());
	}
	else
	{
		sc.UnGet();
	}
}
```

RicardoLuis0 (changes requested):

RicardoLuis0 (approved):

PR #1400, opened 2026-05-24
Allow fog compute from points other than Viewpoint position

dileepvr:
Added ability to compute fog from a point different from Viewpoint position. Useful for third-person cameras, cinematic cameras with far-away subject emphasis, etc.

Orthographic example with thickfog enabled: 
[viewpos_thickfog.zip](https://github.com/user-attachments/files/28199260/viewpos_thickfog.zip)

https://github.com/user-attachments/assets/f8988324-2c2c-47db-bb46-b0a2adb6310e

We need to decide on whether or not to add a flag to the camera actor to activate this.

In the first commit, camera actor's `ViewPos` field has to be non-`null` to activate. Sets `Viewpoint.FogCenterPos` to `camera->tracer` if not `null`, else tries to undo `ViewPos` offset and sets `Viewpoint.FogCenterPos` that way (unless `VPSF_ABSOLUTEPOS` flag is set).

Works with all manner of portals and even in camera textures.

## Checklist

### DIFF
--- a/src/common/rendering/gl/gl_shader.cpp
+++ b/src/common/rendering/gl/gl_shader.cpp
@@ -309,6 +309,7 @@ bool FShader::Load(const char * name, const char * vert_prog_lump, const char * 
 			mat4 NormalViewMatrix;
 
 			vec4 uCameraPos;
+			vec4 uFogCenterPos;
 			vec4 uClipLine;
 
 			float uGlobVis;			// uGlobVis = R_GetGlobVis(r_visibility) / 32.0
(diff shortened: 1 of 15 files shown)


dileepvr:
Thinking about the viewpoint Uniforms some more (in light of the renderer refactor), I think that we should pass on `Viewpoint.bDoOrtho` somehow to the GPU. Could be useful for people that compute normalmaps and stuff in shaders by using `PixelPos.xyz - uCameraPos.xyz`, prompting them to switch to a global direction vector instead. Maybe we should pass `Viewpoint.ViewVector3D` too, instead of making users compute that from matrices.
I don't know if I should try to add that in a PR, since that part of the code is in churn.

PR #1401, opened 2026-05-25
Debugger improvements

nikitalita:
## Checklist

This is a set of improvements to the DAP debugger:
* Fixes a rare segfault that I was getting due to a stack trace being requested in the middle of clearing the state
* Fixes to the pathing to be more consistent and fix errors with wads loaded from folders
* automatic conversion from paths-to-uris and uris-to-paths, and column and line numbers from clients that have differing capabilities

### DIFF
--- a/src/common/scripting/dap/ZScriptDebugger.h
+++ b/src/common/scripting/dap/ZScriptDebugger.h
@@ -29,10 +29,9 @@
 #include "PexCache.h"
 #include "BreakpointManager.h"
 #include "DebugExecutionManager.h"
-#include "IdMap.h"
 #include "Protocol/struct_extensions.h"
+#include "Protocol/converters.h"
 
-#include <thread>
 
 namespace DebugServer
 {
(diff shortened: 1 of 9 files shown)


src/common/scripting/dap/ZScriptDebugger.h
```diff
@@ -94,10 +101,11 @@ class ZScriptDebugger
 	std::shared_ptr<BreakpointManager> m_breakpointManager;
 	std::shared_ptr<RuntimeState> m_runtimeState;
 	std::shared_ptr<DebugExecutionManager> m_executionManager;
-	std::map<int, dap::Source> m_projectSources;
-	std::string m_projectPath;
-	std::string m_projectArchive;
+	caseless_path_map<ProjectItem> m_projectPaths;
```
RicardoLuis0, line 104: i may be misunderstanding what this variable is used for, but if i'm not, why are _project_ paths being treated as caseless? that'd break on any OS with case-sensitive filesystems

nikitalita: the caseless key is just for the lookup; the `archive` member is used for the actual path returned

RicardoLuis0 (approved):

PR #1420, opened 2026-05-31
cmake: make IWYU opt-in to prevent crash on ObjC++ files

tezatay:
- [x] I have reviewed [CONTRIBUTING.md] and agree to its terms.
### Problem

On macos, if "include-what-you-use" is installed (for example via "brew install include-what-you-use"),
CMake's "find_program" silently picks it up and enables it for the entire zdoom target.
This causes the build to crash when compiling .mm files (Objective-C++). "fatal error: 'cstdio' file not found
Assertion failed: !isa<ObjCObjectType>(type) && "IWYU doesn't support Objective-C"
Subprocess aborted"
### Root cause

`find_program(IWYU ...)` finds any system-installed IWYU binary and unconditionally sets
`CXX_INCLUDE_WHAT_YOU_USE` on the target, with no way to opt out without uninstalling the tool.

### Fix

Wrap IWYU in an `option(USE_IWYU OFF)` so it is disabled by default and only enabled explicitly:

```bash
cmake .. -DUSE_IWYU=ON
```

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1294,18 +1294,17 @@ set_source_files_properties( common/engine/sc_man.cpp PROPERTIES OBJECT_DEPENDS 
 set_source_files_properties( ${NOT_COMPILED_SOURCE_FILES} PROPERTIES HEADER_FILE_ONLY TRUE )
 set_source_files_properties( ${GAME_NONPCH_SOURCES} common/textures/hires/hqresize.cpp PROPERTIES SKIP_PRECOMPILE_HEADERS TRUE )
 
-option(ENABLE_IWYU "Enable Include-What-You-Use analysis" ON)
-if (ENABLE_IWYU)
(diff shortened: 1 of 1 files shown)


the-phinet:
You can already disable iwyu by turning off `ENABLE_IWYU`. It defaults to on, since the user needs to choose to install iwyu, and it follows what the rest of the optional config stuff does.

It seems there's a deeper issue that the iwyu config file is looking at c# files. That would be the correct fix here

tezatay:
I tried to build the project using the compilation instructions for macOS. Because the documentation doesn’t mention this by default, that i should disable IWYU, it didn't build. I thought that this is incorrect behavior, because it should build with instructions listed in documentation, so maybe change documentation instead. But still i don't undersand why IWYU is enabled by default when building the project

the-phinet:
The incorrect behaviour is the crashing :sweat_smile: I'll look into it when I have a chance. Thanks for bringing it up!

PR #1422, opened 2026-06-01
Auto Updater

### LINKED ISSUE #1309: Implement Auto-Updater
just adding this as a task so that it shows in the 5.0 milestone, it is currently mid-implementation

RicardoLuis0:
closes #1309

(for windows only, linux appimage should already have it with #1415 and MacOS i have no clue how to)

## Checklist


CMakeLists.txt
```diff
@@ -35,6 +35,19 @@ if(NOT FORCE_NO_LTO)
...
 endif()
 
+if(WIN32)
+	option( USE_UPDATER "enable auto-updater" OFF )
+endif()
+if(USE_UPDATER)
+	set( CURL_REQUIRED TRUE )
+endif()
+
+if(WIN32 AND USE_UPDATER AND MSVC)
```
the-phinet: Instead of wrapping the affirmative case, I would throw an error if updater is requested on an unsupported system.

CMakeLists.txt
```diff
@@ -243,6 +255,22 @@ option( NO_OPENAL "Disable OpenAL sound support" OFF )
 
 find_package( BZip2 )
 find_package( VPX )
+
+if (CURL_REQUIRED)
```
the-phinet, line 255: Using this comment to hold a couple curl issues

the-phinet: Instead of keeping the libcurl dll in repo, please include it as a packaging step like we do for oal and sndfile. Also currently the build is broken, since the dll is missing. This is easily fixed by amending the package step.

the-phinet: The program should function without curl being present. Currently uzdoom fails to even launch when the dll is missing. The dll can go missing for a couple reasons on a user's system, so keeping the program resilient to this in important.

the-phinet: I'd like to avoid vendoring stuff with such a large attack surface. Since windows 10+ includes curl by default, I think it might be better to use curl.exe instead of libcurl. That allows us to move the burden of keeping up with cve's to microsoft. If a user wants to try running uzdoom on an unsupported version of windows, they can easily add curl.exe to their path.

src/common/engine/sc_man.cpp
```diff
@@ -57,29 +57,51 @@
...
+{
+	(*this) = VersionInfo(string);
+}
+
+VersionInfo::operator FString()
+{
+	FString tmp;
+	if(distance != 0 && distance != RC_REVISION_NOTRC)
+	{
+		tmp.Format("%u.%u.%u.%u", major, minor, revision, distance);
```
the-phinet: Best to use actual semver here. `5.0.0-322`

src/common/utility/basics.h
```diff
@@ -141,3 +143,101 @@ static inline void PrefetchL3(const void* Address)
 	_mm_prefetch(static_cast<const char*>(Address), _MM_HINT_T1);
 #endif
 }
+
+class FString;
+
+struct VersionInfo
```
the-phinet: This would make way more sense in version.h

src/common/utility/zstring.h
```diff
@@ -377,6 +377,10 @@ class FString
 	void Split(TArray<FString>& tokens, const FString &delimiter, EmptyTokenType keepEmpty = TOK_KEEPEMPTY) const;
 	void Split(TArray<FString>& tokens, const char *delimiter, EmptyTokenType keepEmpty = TOK_KEEPEMPTY) const;
 
+
+	TArray<FString> SplitNewLines(int minWrapLen = -1, int maxLineLen = -1, EmptyTokenType keepEmpty = TOK_KEEPEMPTY) const;
+	void SplitNewLines(TArray<FString>& tokens, int minWrapLen = -1, int maxLineLen = -1, EmptyTokenType keepEmpty = TOK_KEEPEMPTY) const;
```
the-phinet, line 383: Can you document what this does? Based on the method signature, there are a number of very different things it could be doing

src/version.h
```diff
@@ -95,6 +95,19 @@
...
 #define QUERYIWADDEFAULT true
 #define BUGS_URL "https://github.com/UZDoom/UZDoom/issues"
+
+#define UPDATER_URL_STABLE "https://zdoom.org/uzdoom-latest.php"
+#define UPDATER_URL_STABLE_BACKUP "https://api.github.com/repos/UZDoom/UZDoom/releases/latest"
+
+#define UPDATER_URL_PREVIEW "https://zdoom.org/uzdoom-preview.php"
+#define UPDATER_URL_PREVIEW_BACKUP "https://api.github.com/repos/UZDoom/UZDoom/releases/tags/x-preview"
+
+#define UPDATER_URL_TESTING "https://zdoom.org/uzdoom-testing.php"
```
the-phinet, line 105: I don't see a reason to have this as an option. Looking at the download metrics, basically nobody uses this stream, and If you are, you know how to update/look for updates. The auto-updater is a set-and-forget thing. The testing branch is very much not that.

wadsrc/static/ui/arrow.png
the-phinet, line 1: What is the license on these images?

src/widgets/updatebuttonbar.h
```diff
@@ -0,0 +1,246 @@
...
+#pragma once
+
+#include <zwidget/core/widget.h>
+#include <optional>
+#include "basics.h"
+
+class LauncherWindow;
+class PushButton;
+
+struct date_t
```
the-phinet: While this is cool, dates are different in different locales. It would probably be better to just use days since the epoch for a timestamp. If we were to keep using the date struct, defining exactly what format it is (looks like 8601) and to break it out into its own file

src/widgets/updatebuttonbar.cpp
```diff
@@ -0,0 +1,1639 @@
...
+
+
+//ugh something is including windows.h, probably curl? disable parts of it to minimize conflicts
+#ifndef NOMINMAX
+	#define NOMINMAX // mingw already defines NOMINMAX??????
+#endif
+
+#define WIN32_LEAN_AND_MEAN
+
+// The #defines here *MUST* match serializer.cpp, or we will get countless strange errors.
```
the-phinet: Instead of a comment like this, these should be moved over to cmake

src/widgets/updatebuttonbar.cpp
```diff
@@ -0,0 +1,1639 @@
...
+**
+** SPDX-License-Identifier: GPL-3.0-or-later
+**
+**---------------------------------------------------------------------------
+**
+*/
+
+//TESTING TESTING TESTING
+
+//#define DEBUG_FORCE_UPDATE
```
the-phinet: Having these as cvars would be helpful for future debugging. (and me who would prefer downloading ci binaries over wresting with msvc)

src/widgets/updatebuttonbar.cpp
```diff
@@ -0,0 +1,1639 @@
...
+#include <thread>
+#include <mutex>
+#include <filesystem>
+#include "filesystem.h"
+#include "cmdlib.h"
+#ifdef _WIN32
+	#include <shellapi.h>
+#endif
+
+CVAR(String, cached_update, "", CVAR_ARCHIVE | CVAR_CONFIG_ONLY | CVAR_GLOBALCONFIG | CVAR_NOSET | CVAR_HIDDEN);
```
the-phinet: Can these all get a prefix to keep them in one place in the cfg file? `updater_*`

src/widgets/updatebuttonbar.cpp
```diff
@@ -0,0 +1,1639 @@
...
+constexpr int arrow_margin = 0;
+
+void LauncherWindow::OnWindowClose()
+{ // don't close launcher window if popup is being shown
+	if(!currentPopup) Close();
+}
+
+UpdateButtonBar::UpdateButtonBar(LauncherWindow *parent) : Widget(parent)
+{
+	SetStyleColor("bg-default-color", Colorf(0.0f, 0.3f, 0.5f, 1.0f));
```
the-phinet, line 447: Why can't these use the theme system?

src/widgets/updatebuttonbar.cpp
```diff
@@ -0,0 +1,1639 @@
...
+}
+
+LauncherWindow* UpdateButtonBar::GetLauncher() const
+{
+	return static_cast<LauncherWindow*>(Parent());
+}
+
+void UpdateButtonBar::OpenUpdateInitChoice()
+{
+	OpenPopup(this, "Update Checker", {"Would you like to automatically check for updates?", "(this can be changed later in the options tab)"}, // TODO: localize
```
the-phinet, line 725: Please add a "check for updates now" button in the about page, too.

src/widgets/updatebuttonbar.cpp
```diff
@@ -0,0 +1,1639 @@
...
+	{
+		{
+			"Yes, and auto-install them", 4, [this](auto &self) // TODO: localize
```
the-phinet: I don't see a reason to have a autoinstall option. I don't know how common it is to have individual games expose an option like this on windows, but it feels like added complexity for little reason. A popup saying an update is available is enough imo

src/widgets/updatebuttonbar.cpp
```diff
@@ -0,0 +1,1639 @@
...
+				OpenPopup(buttonBar, "Failed to open zip file", {"Update was cancelled"}, // TODO: localize
+				{
+					{"Back", 0, [](auto &self){
+						self.Close();
+					}}
+				});
+			}
+			else
+			{
+				#ifdef _WIN32	// this is technically 'portable' code since it only uses the stdlib, but what it does only makes sense on windows,
```
the-phinet, line 1204: Remove this. Since this entire file is wrapped in a cmake option, I think this ifdef is asking for bugs

src/widgets/updatebuttonbar.cpp
the-phinet, line 1: Can this file be broken up into the view and controller? Decoupling the zwidget stuff from the update logic will make future maintenance way easier

src/win32/updater.c
the-phinet, line 1: This needs a copyright header

the-phinet (commented):
Looks good. A couple things I think need addressing.

One thing of note when testing is that the updater would sometimes take the startup time from a fraction of a second to 10+ seconds before the launcher window popped up.

src/widgets/updatebuttonbar.cpp
RicardoLuis0: > Remove this. Since this entire file is wrapped in a cmake option, I think this ifdef is asking for bugs

it's wrapped in cmake currently, but it is written to be expansible to further platforms in the future, and most of the code is platform-agnostic

src/widgets/updatebuttonbar.cpp
RicardoLuis0: it's not a fully automatic install, it's the difference between an instant pop-up for new versions (that you can still skip/refuse) vs the banner saying there's an update

src/widgets/updatebuttonbar.cpp
RicardoLuis0: i'll try to do it for 5.x/etc, still need to refactor the popups as well, but this isn't a concern for 5.0

src/widgets/updatebuttonbar.cpp
RicardoLuis0: ah, i knew i was forgetting about something 😅

src/widgets/updatebuttonbar.cpp
RicardoLuis0: didn't think it'd be useful, since any custom builds (ex. for standalone games, etc) would simply want to keep the updater disabled

RicardoLuis0: but i can add it later for 5.x

src/widgets/updatebuttonbar.cpp
RicardoLuis0: good catch, will do

src/widgets/updatebuttonbar.cpp
RicardoLuis0: fair, i just didn't want anyone enabling them by accident lol

RicardoLuis0: would debug-build-only cvars work?

src/widgets/updatebuttonbar.cpp
RicardoLuis0: yeah, but this is what was already done for serializer_doom.cpp - the includes were simply copied over

wadsrc/static/ui/arrow.png
RicardoLuis0: CC0 i guess, made it myself with paint.net

src/version.h
RicardoLuis0: yeah, but it's trivial to support, so why not?

src/common/utility/zstring.h
RicardoLuis0: basically, they split a string into different lines of a specified maximum length (and when it needs to wrap, the maximum length can be reduced down to minWrapLen to avoid splitting words)

RicardoLuis0: i'll add that info to a comment

src/widgets/updatebuttonbar.h
RicardoLuis0: > While this is cool, dates are different in different locales. It would probably be better to just use days since the epoch for a timestamp. If we were to keep using the date struct, defining exactly what format it is (looks like 8601) and to break it out into its own file

format is not quite 8601, it's just `<year>-<month>-<day>`, no support for timezones since the updater works on a minimum resolution of 2 days, so it doesn't need that much precision

src/common/utility/basics.h
RicardoLuis0: IIRC there was a header loop when i tried to have it on version.h, so i had to move it

src/common/engine/sc_man.cpp
RicardoLuis0: > Best to use actual semver here. `5.0.0-322`

this is just for the ini serialization pretty much, keeping it as all-dots makes it simpler

CMakeLists.txt
RicardoLuis0: > I'd like to avoid vendoring stuff with such a large attack surface. Since windows 10+ includes curl by default, I think it might be better to use curl.exe instead of libcurl. That allows us to move the burden of keeping up with cve's to microsoft. If a user wants to try running uzdoom on an unsupported version of windows, they can easily add curl.exe to their path.

curl.exe is not a replacement in the slightest for the library, it cannot do things like progress bars, etc, and the version of curl that windows ships is horrendously out of date

RicardoLuis0: > Instead of keeping the libcurl dll in repo, please include it as a packaging step like we do for oal and sndfile. Also currently the build is broken, since the dll is missing. This is easily fixed by amending the package step.

the vendored version is just for compilation, the dll will/should be the newest one for every release, just don't know how to do that in CI

RicardoLuis0: > The program should function without curl being present. Currently uzdoom fails to even launch when the dll is missing. The dll can go missing for a couple reasons on a user's system, so keeping the program resilient to this in important.

?????? that's for every dll, can't really do anything else on windows unless every single function is loaded manually with LoadLibrary + GetProcAddress, which would lead to a horrible to work with and a bunch of win32 code in every call to a curl function

CMakeLists.txt
RicardoLuis0: would you mind changing that? i'm really not great with CMake 😅

CMakeLists.txt
RicardoLuis0: > > I'd like to avoid vendoring stuff with such a large attack surface. Since windows 10+ includes curl by default, I think it might be better to use curl.exe instead of libcurl. That allows us to move the burden of keeping up with cve's to microsoft. If a user wants to try running uzdoom on an unsupported version of windows, they can easily add curl.exe to their path.
> 
> curl.exe is not a replacement in the slightest for the library, it cannot do things like progress bars, etc, and the version of curl that windows ships is horrendously out of date

image

curently, on a windows 10 machine **with extended support enabled**, curl was last updated _over a year ago_, and while no high severity CVEs have happened since 2023, multiple medium ones have since may 2025, the very month after the last update to windows curl.exe

src/common/utility/basics.h
the-phinet: Weird. I'll take a peak

src/version.h
the-phinet: Code surface and overwhelming people with options. I'd rather this be removed, but I won't block over it either

src/widgets/updatebuttonbar.cpp
the-phinet: I understand that. Having the option set in one place (serializer) is fine. The moment it needs to be synced between multiple places, copypasta is not the right tool

src/version.h
RicardoLuis0: > Code surface and overwhelming people with options. I'd rather this be removed, but I won't block over it either

it's not an option, each build of the engine only updates to/from its own branch (with the exception of RC builds, which also update to stable releases), and it uses the exact same code as the preview updates - the only change is the URL

src/widgets/updatebuttonbar.cpp
the-phinet: Maybe? I don't think worrying about cvars being set by mistake is much of an issue. In general, I prefer keeping options consistent between builds.

src/widgets/updatebuttonbar.cpp
RicardoLuis0: > I understand that. Having the option set in one place (serializer) is fine. The moment it needs to be synced between multiple places, copypasta is not the right tool

it was two places - serializer.cpp and serializer_doom.cpp - i'll move it to a separate header though, easier than changing it on CMake

src/widgets/updatebuttonbar.cpp
the-phinet: In that case, I think that's even more of a better reason to not have two options

src/widgets/updatebuttonbar.cpp
the-phinet: That's what I mean. Since the code is platform-agnostic, having a platform check is a problem

src/widgets/updatebuttonbar.cpp
RicardoLuis0: > In that case, I think that's even more of a better reason to not have two options

i mean, i added specifically because it's something i myself would want instead of the update banner, but i also added the banner because i know many people can and will get annoyed at an automatic pop-up

src/common/engine/sc_man.cpp
the-phinet: It's a difference of one char, I don't think using all dots for a custom format is that much simpler. Besides, using distance is an unreliable way of detecting changes. Especially in the testing branch, where the number of commits can decrease. For the testing branch in particular, using the a semver with a sha as the build number would be much more reliable.

CMakeLists.txt
the-phinet: > curl.exe is not a replacement in the slightest for the library, it cannot do things like progress bars, etc

That's fine. The updater really doesn't need a progress bar, just a throbber. The updates are small enough that a progressbar is not particularly useful.

> the version of curl that windows ships is horrendously out of date
> curl was last updated over a year ago

That's also fine. While I don't trust microsoft, I trust us less to keep up with things. Being updated recently is not a metric of how secure something is. I trust that the CVEs are low priority enough (or in parts of the curl exe that have been stripped out of the windows build), that an update is not required. Plus, once an old version of ours is out there, it will forever be vulnerable.

> the vendored version is just for compilation, the dll will/should be the newest one for every release, just don't know how to do that in CI

Oh, okay. Is there a reason to have it (and the .a files) in the repo at all? We just need the headers, correct?

> unless every single function is loaded manually with LoadLibrary + GetProcAddress

That's fine and normal. UZDoom does not refuse to start when oal is missing, it just has no sound. If we include libcurl (which I don't think we should), it is the most likely part of the engine to be removed by antivirus or bad group policies.

src/widgets/updatebuttonbar.h
the-phinet: My point still stands. Caring about leap years (which are also inconsistent in different locales) for something like this is over-engineering. Days since the epoch or an ordinal date would be simpler and better.

CMakeLists.txt
RicardoLuis0: > > curl.exe is not a replacement in the slightest for the library, it cannot do things like progress bars, etc
> 
> That's fine. The updater really doesn't need a progress bar, just a throbber. The updates are small enough that a progressbar is not particularly useful.

it's not useful for people with fast internet, but it very much is needed for someone with a slower one, because 33 MB isn't large but it also isn't small enough to be insignificant, plus calling out to an external exe is just no, no

> > the vendored version is just for compilation, the dll will/should be the newest one for every release, just don't know how to do that in CI
> 
> Oh, okay. Is there a reason to have it (and the .a files) in the repo at all? We just need the headers, correct?

yeah, i just forgot to remove it

> > unless every single function is loaded manually with LoadLibrary + GetProcAddress
> 
> That's fine and normal. UZDoom does not refuse to start when oal is missing, it just has no sound. If we include libcurl (which I don't think we should), it is the most likely part of the engine to be removed by antivirus or bad group policies.

oal has all the shitty manual loading done for it, yes, and it's horrible, though seems the platform-specific stuff is separated into it's own file with FModule, but it's still horrible to work with

src/common/engine/sc_man.cpp
RicardoLuis0: > It's a difference of one char, I don't think using all dots for a custom format is that much simpler.

simpler because the parsing does the same thing for all numbers, instead of doing one thing for the first 3 and a different one for the last

> Besides, using distance is an unreliable way of detecting changes. Especially in the testing branch, where the number of commits can decrease. For the testing branch in particular, using the a semver with a sha as the build number would be much more reliable.

ah, i didn't think of that, good catch

src/widgets/updatebuttonbar.h
RicardoLuis0: > which are also inconsistent in different locales

oh fuck, didn't know that 😱

src/widgets/updatebuttonbar.cpp
RicardoLuis0: > That's what I mean. Since the code is platform-agnostic, having a platform check is a problem

the code inside the check isn't platform-agnostic (it only uses the stdlib, but what it _does_ is specific to windows), the code in the file itself as a whole is

src/common/engine/sc_man.cpp
RicardoLuis0: > > Besides, using distance is an unreliable way of detecting changes. Especially in the testing branch, where the number of commits can decrease. For the testing branch in particular, using the a semver with a sha as the build number would be much more reliable.
> 
> ah, i didn't think of that, good catch

i'll change it to do a `!=` comparison for now, should work for force pushes/resets/etc, if it ends up not being sufficient, i'll change it to the git hash in a subsequent PR, since that needs major changes to the code

CMakeLists.txt
the-phinet: Done

src/common/utility/basics.h
the-phinet: Fixed. The gitinfo.h being included in basics.h was also causing a bunch of compilation churn (and cache thrashing), so this is kinda a required fix. The only downside is the one if statement is no longer compiletime

src/widgets/updatebuttonbar.cpp
the-phinet: That makes sense. I misunderstood what the comment was referring to

src/version.h
the-phinet: That's fair. I guess there's no way to choose a stream from the engine itself (but that might be a good future thing)

CMakeLists.txt
the-phinet: libcurl gets packaged through ci now.

If we're set on using libcurl.dll, I think the FModule route is the best plan of action. I'm not sure how to go about implementing that

src/widgets/updatebuttonbar.cpp
the-phinet: Only the popup is probably the best choice -- especially since we have the check for updates now button. If we were to keep both around, the buttons asking what to do need to be clarified.

RicardoLuis0: > Only the popup is probably the best choice -- especially since we have the check for updates now button. If we were to keep both around, the buttons asking what to do need to be clarified.

the popup is an interruption though, and when people get annoyed at interruptions they'll just disable update-checking period

CMakeLists.txt
RicardoLuis0: > libcurl gets packaged through ci now.
> 
> If we're set on using libcurl.dll, I think the FModule route is the best plan of action. I'm not sure how to go about implementing that

replacing all calls to curl functions to calls to function pointers, and loading curl (with `FModule::Open`)/fetching those pointers (with `FModule::GetSym`) in `UpdateButtonBar::InitCurl()`

RicardoLuis0: seems `FModule::RegisterStatic` and `FModule::Load` can be used too? bit hard to parse how to at a glance though

RicardoLuis0: something to do with TReqProc and TOptProc

RicardoLuis0: seems the openal code has a helper function that could be copied
https://github.com/UZDoom/UZDoom/blob/749859d20c6f678cfb0025803115414993b2d8a8/src/common/audio/sound/oalload.h#L29-L122

the-phinet (approved):

PR #1450, opened 2026-06-08
Logging improvements

the-phinet:
Game would swallow any prints that happen early in the engine's boot sequence, making debugging annoying

## Checklist

### DIFF
--- a/src/common/console/c_console.cpp
+++ b/src/common/console/c_console.cpp
@@ -409,7 +409,9 @@ void WriteLineToLog(FILE *LogFile, const char *outline)
 	fflush(LogFile);
 }
 
-inline int _PrintString(PrintFlag iprintlevel, const char *outline)
+namespace detail
+{
+inline int PrintString(PrintFlag iprintlevel, const char *outline)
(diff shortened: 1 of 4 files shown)


src/common/console/c_console.cpp
```diff
@@ -408,49 +409,74 @@ void WriteLineToLog(FILE *LogFile, const char *outline)
 	fflush(LogFile);
 }
 
+inline int _PrintString(PrintFlag iprintlevel, const char *outline)
```
kcat: Names that start with an underscore followed by an uppercase letter are reserved (as well as names that start with a double underscore, or names that start with a single underscore in the global namespace).

the-phinet: Oh, I wasn't aware of that. I'll fix that up

the-phinet: Oof. Enabled the compiler flag to warn about issues like this. 2503 instances, with only 239 of them being in thirdparty code

Boondorl (approved):

PR #1455, opened 2026-06-11
Better backtrace in debug builds on windows

RicardoLuis0:
to make debugging the inersection between zscript and the engine easier
image
(does not at all affect release builds)

## Checklist


RicardoLuis0:
the code for grabbing backtrace info on linux seems to be entirely non-functional though, so i couldn't do much there
image

PR #1459, opened 2026-06-12
Change minimum sound channels from 64 back to 8.

g0mb4:
It is handy to lower the number of sound channels in chaotic battles.

## Checklist

### DIFF
--- a/src/common/audio/sound/oalsound.cpp
+++ b/src/common/audio/sound/oalsound.cpp
@@ -51,7 +51,7 @@ FModule OpenALModule{"OpenAL"};
 
 CUSTOM_CVAR(Int, snd_channels, 128, CVAR_ARCHIVE | CVAR_GLOBALCONFIG)	// number of channels available
 {
-	if (self < 64) self = 64;
+	if (self < 8) self = 8;
 }
 CVARD(String, snd_aldriver, DEFAULT_DRIVER, CVAR_ARCHIVE|CVAR_GLOBALCONFIG, "See alsoftrc.sample for details")
 CVAR(Bool, snd_waterreverb, true, CVAR_ARCHIVE | CVAR_GLOBALCONFIG)
--- a/src/gameconfigfile.cpp
+++ b/src/gameconfigfile.cpp
@@ -430,9 +430,9 @@ void FGameConfigFile::DoGlobalSetup ()
 			var = FindCVar("snd_channels", NULL);
 			if (var != NULL)
 			{
-				// old settings were default 32, minimum 8, new settings are default 128, minimum 64.
+				// old settings were default 32, minimum 8, new settings are default 128, minimum 8.
 				UCVarValue v = var->GetGenericRep(CVAR_Int);
-				if (v.Int < 64) var->ResetToDefault();
+				if (v.Int < 8) var->ResetToDefault();
 			}
 		}
 		if (EngineLastRunVer < 214)
--- a/wadsrc/static/menudef.txt
+++ b/wadsrc/static/menudef.txt
@@ -2533,7 +2533,7 @@ OptionMenu SoundOptions protected
 	Submenu "$SNDMNU_MIDIDEVICE",       "MidiDevicesMenu"
 	Submenu "$SNDMNU_MIDIPLAYER",       "MidiPlayerOptions"
 	StaticText " "
-	Slider "$SNDMNU_CHANNELS",          "snd_channels",        8, 256, 8, 0
+	Slider "$SNDMNU_CHANNELS",          "snd_channels",        64, 256, 8, 0
 	Option "$SNDMNU_BACKGROUND",        "i_soundinbackground", "OnOff"
 	Option "$SNDMNU_UNDERWATERREVERB",  "snd_waterreverb",     "OnOff"
 	Option "$SNDMNU_RANDOMIZEPITCHES",  "snd_pitched",         "OnOff"
--- a/wadsrc/static/menudef.txt
+++ b/wadsrc/static/menudef.txt
@@ -2533,7 +2533,7 @@ OptionMenu SoundOptions protected
 	Submenu "$SNDMNU_MIDIDEVICE",       "MidiDevicesMenu"
 	Submenu "$SNDMNU_MIDIPLAYER",       "MidiPlayerOptions"
 	StaticText " "
-	Slider "$SNDMNU_CHANNELS",          "snd_channels",        64, 256, 8, 0
+	Slider "$SNDMNU_CHANNELS",          "snd_channels",        8, 256, 8, 0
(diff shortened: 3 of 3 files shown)


RicardoLuis0:
this will break mods that need to play more than 8 sounds at once though, since 64 has been the guaranteed minimum supported channels for almost a decade now https://github.com/UZDoom/UZDoom/commit/005e6871f9324d408eb719c496b6ad4360e06003

the-phinet:
> this will break mods that need to play more than 8 sounds at once

Ehh, I think that's fine. It's not changing the default. Mods that require pp shaders 'break' when using gles, too

RicardoLuis0:
(keep in mind, this is the minimum channels **for the sound engine** not for A_StartSound/etc, those always have infinite channels, and the currently playing sounds are slotted into physical channels based on distance/etc)

g0mb4:
> this will break mods that need to play more than 8 sounds at once though

I see, but I think the user should have the power to change this.

RicardoLuis0:
> > this will break mods that need to play more than 8 sounds at once
> 
> Ehh, I think that's fine. It's not changing the default. Mods that require pp shaders 'break' when using gles, too

having guaranteed minimums for mods is a good thing, the GLES breakage is explicitly because GPUs that need to run with GLES just don't have the power to run arbitrary shaders, it'd just turn into a slideshow

g0mb4:
> keep in mind, this is the minimum channels for the sound engine not for A_StartSound/etc, those always have infinite channels, and the currently playing sounds are slotted into physical channels based on distance/etc

Hmm, I want to avoid the situation when there are too many things happening at once and the sounds superposed resulting a value clamping (very unpleasing sounds).

RicardoLuis0:
> > keep in mind, this is the minimum channels for the sound engine not for A_StartSound/etc, those always have infinite channels, and the currently playing sounds are slotted into physical channels based on distance/etc
> 
> Hmm, I want to avoid the situation when there are too many things happening at once and the sounds superposed resulting a value clamping (very unpleasing sounds).

unpleasant sounds should be fixed by just making overlapping stuff not sound bad, regardless of channel count

g0mb4:
> > > keep in mind, this is the minimum channels for the sound engine not for A_StartSound/etc, those always have infinite channels, and the currently playing sounds are slotted into physical channels based on distance/etc
> > 
> > 
> > Hmm, I want to avoid the situation when there are too many things happening at once and the sounds superposed resulting a value clamping (very unpleasing sounds).
> 
> unpleasant sounds should be fixed by just making overlapping stuff not sound bad, regardless of channel count

Sure, but it is tricky on slaughter maps.

RicardoLuis0:
@kcat do you know if unpleasant overlapping sounds are at all fixable without limiting channels?

g0mb4:
Maybe a boolean option like "hard-limit sound channels" would help, that would enforce the `$sound_channels` on A_StartSound/etc.

HexaDoken:
> unpleasant sounds should be fixed by just making overlapping stuff not sound bad, regardless of channel count

In principle I agree, but a very big offender on this issue are stock doom sounds, which I imagine aren't going to get fixed any time soon. They are very clearly designed with max 8 slots in mind and on certain maps it really shows. The default `$limit 2` helps somewhat but it has never been a perfect solution.

RicardoLuis0:
> > unpleasant sounds should be fixed by just making overlapping stuff not sound bad, regardless of channel count
> 
> In principle I agree, but a very big offender on this issue are stock doom sounds, which I imagine aren't going to get fixed any time soon. They are very clearly designed with max 8 slots in mind and on certain maps it really shows. The default `$limit 2` helps somewhat but it has never been a perfect solution.

by "making overlapping stuff not sound bad" i mean a fix in the sound backend, not in the sound effects themselves

kcat:
> @kcat do you know if unpleasant overlapping sounds are at all fixable without limiting channels?

IIRC, the engine already checks if too many instances of the same sound try to play too close together:

```c++
        // If this is a singular sound, don't play it if it's already playing.
        if ((sfx->bSingular || (flags & CHANF_SINGULAR)) && CheckSingular(sound_id))
        {
                chanflags |= CHANF_EVICTED;
        }

        // If the sound is unpositioned or comes from the listener, it is
        // never limited.
        if (type == SOURCE_None || source == listener.ListenerObject)
        {
                near_limit = 0;
        }

        // If this sound doesn't like playing near itself, don't play it if that's what would happen.
        if (near_limit > 0 && CheckSoundLimit(sfx, pos, near_limit, limit_range, type, source, channel, attenuation))
        {
                chanflags |= CHANF_EVICTED;
        }

        // If the sound is blocked and not looped, return now. If the sound
        // is blocked and looped, pretend to play it so that it can
        // eventually play for real.
        if ((chanflags & (CHANF_EVICTED | CHANF_LOOP)) == CHANF_EVICTED)
        {
                return NULL;
        }
```

The `bSinglular` and `NearLimit` properties need to be set in the sound definition.

SanyaWaffles:
I'd rather not see the engine revert this because some mods don't know how to do sound design.

the-phinet:
What happens when the engine runs out of sound channels? The audio messes up a bit. Who are we to deny the user from lowering the number to something we consider unusable. I see no problem at all with allowing this. 

If it made people feel better, how about only soft-limiting it? The menu's minimum stays at 64, but if you set the cvar manually, it can be whatever

mc776:
Could it be one of those options that isn't available on the menu but can be done through the console or editing the ini? That way people who need it will have it but no one's going to break things without realizing it.

g0mb4:
> If it made people feel better, how about only soft-limiting it? The menu's minimum stays at 64, but if you set the cvar manually, it can be whatever

My main concern with that method is that the regular user won't know that this option is tied to a 'cvar', nor will they know the name of the cvar.

But sure, if the community votes for this approach, I'll change the code.

RicardoLuis0:
> What happens when the engine runs out of sound channels? The audio messes up a bit. Who are we to deny the user from lowering the number to something we consider unusable. I see no problem at all with allowing this.
> 
> If it made people feel better, how about only soft-limiting it? The menu's minimum stays at 64, but if you set the cvar manually, it can be whatever

taking what kcat said, seems same-sounds are already limited to 4 by default

https://github.com/UZDoom/UZDoom/blob/38032fff26b0e03419718acc4b1df5f41b6c7d89/src/common/audio/sound/s_soundinternal.h#L116

the-phinet (approved):

the-phinet:
Head's up, unless there's a merge conflict (or the potential for one), there's no need to resync your pr branches. And if you do, rebase instead of making a merge commit :)

g0mb4:
> Head's up, unless there's a merge conflict (or the potential for one), there's no need to resync your pr branches. And if you do, rebase instead of making a merge commit :)

Noted, sorry.

Issue #1387, opened 2026-05-20
Move "Earthquake shake intensity" to accessibility options


HexaDoken:
Currently in Display Options, which feels a bit weird, though it does make _some_ sense. Still, with A_Quake absolutely being a photosensitivity hazard, accessibiltiy options is probably a more logical home for it.

the-phinet:
Agreed. There is a menu re-work coming up eventually, where we plan to re-organize the menus a lot. If someone wants to pr this before that, I'll happily merge it.

nashmuhandes:
Done with https://github.com/UZDoom/UZDoom/commit/5bf01ebb6bca2fe5d767f275af9b99dd2ee401fb

PR #1460, opened 2026-06-12
Move Earthquake Intensity slider to Accessibility settings

mattsilverwood:
Moves the Earthquake intensity slider definition to the accessibility menu. Addresses #1387.

Not very sure about the separation, but I felt it necessary to separate it visually as no other setting is "related" to it, in a manner of speaking. Not sure if another header should be added, nor what it should be called. Capitalisation may also need adjusting? Seems inconsistent when compared to the rest of the options.

First contribution, keeping this one minor as I get more comfortable with the codebase. I know it's mentioned in the contributing file and as silly it is to say for such a small change, no AI is involved in this.

## Checklist

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -274,3 +274,4 @@ Sam "LordEntr0py" Widdowson
 Sierra0451
 CM "Vault."
 Jay Smithen
+Matt Silverwood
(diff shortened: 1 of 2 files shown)


RicardoLuis0 (dismissed):

RicardoLuis0:
btw, don't forget to add yourself to https://github.com/UZDoom/UZDoom/blob/trunk/CONTRIBUTORS

mattsilverwood:
Thanks for the reminder.

RicardoLuis0 (approved):

PR #1500, opened 2026-06-20
Add the Suffix to the message itsself to prevent breakage

heiko123abc:
## Checklist

While testing the notification system in combination with the recently added Fadeout option, I realized that there is a bug in the message stacking. When the message is pushed up in the queue, it loses its suffix. This PR fixes that by tying the suffix to the message itsself rather than appending it dynamically, thus staying together when pushed up.

Before:
grafik

After:
grafik

### DIFF
--- a/src/console/c_notifybuffer.cpp
+++ b/src/console/c_notifybuffer.cpp
@@ -115,6 +115,12 @@ void FNotifyBuffer::AddString(int printlevel, FString source)
 		countedIdentical = 1;
 	}
 
+    // Make the suffix directly a part ofthe message string
+    if (countedIdentical > 1)
+    {
+        source.AppendFormat(" (x%d)", countedIdentical);
+    }
+
 
 	// [MK] allow the status bar to take over notify printing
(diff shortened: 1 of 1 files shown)


src/console/c_notifybuffer.cpp
```diff
@@ -115,6 +115,13 @@ void FNotifyBuffer::AddString(int printlevel, FString source)
 		countedIdentical = 1;
 	}
 
+	source.StripRight();
+    // insert the suffix directly as part ofthe message string 
+    if (countedIdentical > 1)
+    {
+        source.AppendFormat(" (x%d)", countedIdentical);
```
the-phinet: Can this use the new [`VALFORMAT_MULTIPLIER`](https://hosted.weblate.org/translate/uzdoom/zdoom-engine-strings/en/?checksum=48f9138fbdc7e12b) localization key?

heiko123abc: I have added it in and it works.
grafik

the-phinet (approved):

PR #1521, opened 2026-06-24
Ability to Disable Config Saving

### LINKED ISSUE #1380: quit w/o save
### Game

All

### Problem Description

It would be nice if you implemented a 'quit w/o save' [to .ini] feature. This is handy as you might have experimented with some settings, don't like them, but don't know anymore what the orginal settings were. This setting would only be shown when something got changed.

### Feature Description

1) change something in settings.
2) Try to quit
3) Get the extra option 'quit w/o save' 

### Additional Media

_No response_

heiko123abc:
## Checklist

Fixes #1380 by providing two ways to prevent config saving:

- a new parameter `-nosaveconfig` which when launched with disables the games abiltiy to write to any config file with and associated new misc menu option.


RicardoLuis0:
why have this as a cvar? it'll never be saved to the ini since it disables saving xD

RicardoLuis0:
> why have this as a cvar? it'll never be saved to the ini since it disables saving xD

at the very least it'd need to be a custom cvar that _forces_ saving when it's changed, that way it can be enabled and _persist_

heiko123abc:
> why have this as a cvar? it'll never be saved to the ini since it disables saving xD

> A commandline switch (and cvar mirroring it) to disable config saving altogether could be useful, and would solve this use-case 

 _Originally posted by @the-phinet in [#1380](https://github.com/UZDoom/UZDoom/issues/1380#issuecomment-4467300549)_

I was wondering about that too, but i saw this in the original issue. 

> at the very least it'd need to be a custom cvar that forces saving when it's changed, that way it can be enabled and persist

Is there a simple way to do it? I havent seen it from glancing at the code

the-phinet:
> why have this as a cvar? it'll never be saved to the ini since it disables saving xD

I haven't looked at the implementation here, but hitting the save config button should still work when having this enabled. Also the cvar can be used to enable (or disable) the option after initial boot.

> at the very least it'd need to be a custom cvar that forces saving when it's changed, that way it can be enabled and persist

Please don't do this. The option is designed to allow the user to disable any implicit saving of things

RicardoLuis0:
> > why have this as a cvar? it'll never be saved to the ini since it disables saving xD
> 
> I haven't looked at the implementation here, but hitting the save config button should still work when having this enabled. Also the cvar can be used to enable (or disable) the option after initial boot.

this disables the very function that saves though (M_SaveDefaults), so literally no saving will go through, hence the CVar would never be persisted once changed (it doesn't disable automatic saving, as it is right now it disables **all** config saving in any form)

PR #1525, opened 2026-06-25
Properly orient sprite canvas textures when drawing them

AFADoomer:
- Flip sprites and weapon sprites vertically if they are canvas textures.

This matches the handling of textures and flats, where the engine vertically flips the contents of any hardware canvases before rendering them.

The lack of handling for this case led to automap sprites being flipped vertically from their in-game counterparts.

## Checklist

### DIFF
--- a/src/rendering/hwrenderer/scene/hw_sprites.cpp
+++ b/src/rendering/hwrenderer/scene/hw_sprites.cpp
@@ -1072,7 +1072,7 @@ void HWSprite::Process(HWDrawInfo *di, AActor* thing, sector_t * sector, area_t 
 			return;
 
 		// Canvas textures are stored upside down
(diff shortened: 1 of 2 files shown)


RicardoLuis0 (changes requested):
this needs to be gated by a flag, otherwise it'll break existing mods

AFADoomer:
> this needs to be gated by a flag, otherwise it'll break existing mods

~~How?  Sprites couldn't be canvas textures before my last PR (and that is still only in preview builds, I think).~~

Nevermind...  Added check to not automatically flip sprites if the texture was set via picnum, which is the only way that I know of for a game sprite to have been set to a canvas texture prior to my previous PR.

the-phinet (dismissed):

RicardoLuis0 (approved):

PR #1533, opened 2026-06-28
Updated A_FreezeDeath() and A_GenericFreezeDeath()

KozmunkasKalman:
Added inputs to the A_FreezeDeath() and the A_GenericFreezeDeath() action functions to allow changing the palette translation and the minimum freeze tics.

## Checklist

### DIFF
--- a/wadsrc/static/zscript/actors/shared/ice.zs
+++ b/wadsrc/static/zscript/actors/shared/ice.zs
@@ -98,11 +98,11 @@ extend class Actor
 	//
 	//============================================================================
 
-	void A_FreezeDeath()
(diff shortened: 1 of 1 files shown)


wadsrc/static/zscript/actors/shared/ice.zs
```diff
@@ -203,11 +203,10 @@ extend class Actor
 		SetStateLabel('null');
 	}
 
-	void A_GenericFreezeDeath()
+	void A_GenericFreezeDeath(string translation = "Ice", int mintics = 75)
```
Boondorl, line 206: `translation` should be a Name rather than a string

PR #1535, opened 2026-06-29
Add actor render flag to set an actor's sprite's normal

AFADoomer:
Added the +SETSPRITENORMAL flag for actors.

- This flag triggers the engine to set the normal of each rendered sprite for that actor to the vector facing toward the current viewpoint.

- With this flag set, dynamic lights can individually affect the actor's sprite (with no shader-side changes necessary), and you can have 'back-lit' sprite actors who are silhouetted against a dynamic light.

## Checklist

### DIFF
--- a/src/playsim/actor.h
+++ b/src/playsim/actor.h
@@ -505,6 +505,7 @@ enum ActorRenderFlag2
 	RF2_INTERPOLATESCALE		= 0x1000,
 	RF2_INTERPOLATEALPHA		= 0x2000,
 	RF2_NODYNAMICLIGHTING		= 0x4000,	// [MC] Disable dynamic lighting effects on sprites/models
+	RF2_SETSPRITENORMAL			= 0x8000,	// [AFA] Set the sprite's normal and allow individual dynamic lights to affect sprites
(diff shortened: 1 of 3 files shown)


RicardoLuis0:
> Added the +SETSPRITENORMAL flag for actors.
> 
> * With this flag set, dynamic lights can individually affect the actor's sprite (with no shader-side changes necessary), and you can have 'back-lit' sprite actors who are silhouetted against a dynamic light.

this isn't enough for that, for it to work it needs to pass in the full light list like it does for models instead of calculating it on the CPU

RicardoLuis0:
> > Added the +SETSPRITENORMAL flag for actors.
> > 
> > * With this flag set, dynamic lights can individually affect the actor's sprite (with no shader-side changes necessary), and you can have 'back-lit' sprite actors who are silhouetted against a dynamic light.
> 
> this isn't enough for that, for it to work it needs to pass in the full light list like it does for models instead of calculating it on the CPU

(i'll probably merge this PR manually into the 5.x renderer rework and do that myself if you can't get to it by then)

Issue #15, opened 2025-10-14
[Feature] Controlling Autosaves


MajorCooke:
Quoting @XaserAcheron :

> someone up above mentioned 'controlling savings', presumably talking about allowing maps to prevent mid-game saves, which has historically always been shot down by graf
> 
> going forward, there's a pattern here that works very well IMO: do it the same way jumping/crouching is handled. have it so mid-game saves can be disabled in zmapinfo, but there's an opt-in user setting that lets folks force it on anyway, so everyone's happy

Battandi:
Yes please, it's probably one of the bigger limitations as of now regarding GzDoom, especially if one wants to design a game with a checkpoint-save point system (like a metroidvania type game for example) or just an arcadey type game.

If I had the chance to disable quicksaves, Simon's Destiny would've been a pretty different game entirely.

PR #1538, opened 2026-06-30
Add `nosave` and `noautoautosave` to MAPINFO.

g0mb4:
A new mechanism was added to control the user's option to manually save the game.
`nosave` setting can be overwritten from the Options.

The change is based on the already existing `nojump`, `nocrouch`, etc.

New MAPINFO flags:
+ `allowsave`  - Allows the user to save the game. (default)
+ `nosave`  - Forbids the user to save the game.
+ `allowautoautosave` - Enables the automatic autosave on loading a map. (default)
+ `noautoautosave` - Disables the automatic autosave on loading a map.

New CVARs:
+ `sv_save 0` - Use MAPINFO setting to determine saving availability. (default)
+ `sv_save 1` - Forbids the user to save the game.
+ `sv_save 2` - Allows the user to save the game.
+ `sv_nosave 1` - Forbids the user to save the game.
+ `sv_allowsave 1` - Allows the user to save the game.

Note: If `sv_nosave` and `sv_allowsave` are both set (invalid state) `sv_allowsave`  will be active.

New `LevelLocals` ZScipt functions:
+ `IsSavingAllowed()` - Determines if the user can save.

New language keys:
+ `GMPLYMNU_ALLOWSAVE` - "Allow saving"
+ `TXT_CANNOTSAVE` - "You cannot save the game."

It is connected to #15.

## Checklist

### DIFF
--- a/src/doomdef.h
+++ b/src/doomdef.h
@@ -176,6 +176,8 @@ enum : unsigned
 	DF3_NO_COOP_ONLY_THINGS	= 1 << 5,	// Any Actor that only appears in co-op is disabled
 	DF3_REMEMBER_LAST_WEAP	= 1 << 6,	// When respawning in co-op, keep the last used weapon out instead of switching to the best new one.
 	DF3_PISTOL_START		= 1 << 7,	// Take player inventory when exiting to the next level.
+	DF3_NO_SAVE				= 1 << 8,	// Don't allow manual saving
+	DF3_YES_SAVE			= 1 << 9,	// Allow manual saving
 };
 
 // [RH] Compatibility flags.
--- a/src/g_game.cpp
+++ b/src/g_game.cpp
@@ -2416,7 +2416,7 @@ void G_DoSaveGame (bool okForQuicksave, bool forceQuicksave, FString filename, c
 
 	if (!(primaryLevel->IsSavingAllowed () || saveFromScript))
 	{
-		Printf (PRINT_HIGH, "Saving is not allowed.\n");
+		Printf (PRINT_HIGH, "%s\n", GStrings.GetString("TXT_CANNOTSAVE"));
 		return;
 	}
 	
--- a/src/g_game.cpp
+++ b/src/g_game.cpp
@@ -119,7 +119,7 @@ void G_PlayerReborn (int player);
 
 void G_DoAutoSave ();
 void G_DoPlayDemo (void);
-void G_DoSaveGame (bool okForQuicksave, bool forceQuicksave, FString filename, const char *description);
+void G_DoSaveGame (bool okForQuicksave, bool forceQuicksave, FString filename, const char *description, bool saveFromScript);
 void G_DoVictory (void);
 
 void SetupLoadingCVars();
@@ -1263,13 +1263,13 @@ void G_Ticker ()
 			G_DoLoadGame ();
 			break;
 		case ga_savegame:
-			G_DoSaveGame (true, false, savegamefile, savedescription.GetChars());
+			G_DoSaveGame (true, false, savegamefile, savedescription.GetChars(), false);
 			gameaction = ga_nothing;
 			savegamefile = "";
 			savedescription = "";
 			break;
 		case ga_quicksave:
-			G_DoSaveGame(true, true, savegamefile, savedescription.GetChars());
+			G_DoSaveGame(true, true, savegamefile, savedescription.GetChars(), false);
 			gameaction = ga_nothing;
 			savegamefile = "";
 			savedescription = "";
@@ -2316,7 +2316,7 @@ void G_DoAutoSave ()
 
 	readableTime = myasctime ();
 	description.Format("Autosave %s", readableTime);
-	G_DoSaveGame (false, false, file, description.GetChars());
+	G_DoSaveGame (false, false, file, description.GetChars(), true);
 }
 
 void G_DoQuickSave ()
@@ -2400,7 +2400,16 @@ static void PutSavePic (FileWriter *file, int width, int height)
 	}
 }
 
-void G_DoSaveGame (bool okForQuicksave, bool forceQuicksave, FString filename, const char *description)
+static bool IsSavingAllowed ()
+{
+	if (dmflags3 & DF3_NO_SAVE)
+		return false;
+	if (dmflags3 & DF3_YES_SAVE)
+		return true;
+	return !(level.flags3 & LEVEL3_SAVE_NO);
+}
+
+void G_DoSaveGame (bool okForQuicksave, bool forceQuicksave, FString filename, const char *description, bool saveFromScript)
 {
 	TArray<FCompressedBuffer> savegame_content;
 	TArray<FString> savegame_filenames;
@@ -2414,6 +2423,12 @@ void G_DoSaveGame (bool okForQuicksave, bool forceQuicksave, FString filename, c
 		return;
 	}
 
+	if (!(IsSavingAllowed () || saveFromScript))
+	{
+		Printf (PRINT_HIGH, "Saving is not allowed.\n");
+		return;
+	}
+	
 	if (demoplayback)
 	{
 		filename = G_BuildSaveName ("demosave");
--- a/src/g_game.cpp
+++ b/src/g_game.cpp
@@ -2400,15 +2400,6 @@ static void PutSavePic (FileWriter *file, int width, int height)
 	}
 }
 
-static bool IsSavingAllowed ()
-{
-	if (dmflags3 & DF3_NO_SAVE)
-		return false;
-	if (dmflags3 & DF3_YES_SAVE)
-		return true;
-	return !(level.flags3 & LEVEL3_SAVE_NO);
-}
-
 void G_DoSaveGame (bool okForQuicksave, bool forceQuicksave, FString filename, const char *description, bool saveFromScript)
 {
 	TArray<FCompressedBuffer> savegame_content;
@@ -2423,7 +2414,7 @@ void G_DoSaveGame (bool okForQuicksave, bool forceQuicksave, FString filename, c
 		return;
 	}
 
-	if (!(IsSavingAllowed () || saveFromScript))
+	if (!(primaryLevel->IsSavingAllowed () || saveFromScript))
 	{
 		Printf (PRINT_HIGH, "Saving is not allowed.\n");
 		return;
--- a/src/g_levellocals.h
+++ b/src/g_levellocals.h
@@ -861,6 +861,20 @@ public:
 		return !(flags & LEVEL_FREELOOK_NO);
 	}
 
+	//==========================================================================
+	//
+	//
+	//==========================================================================
+
+	bool IsSavingAllowed() const
+	{
+		if (dmflags3 & DF3_NO_SAVE)
+			return false;
+		if (dmflags3 & DF3_YES_SAVE)
+			return true;
+		return !(flags3 & LEVEL3_SAVE_NO);
+	}
+
 	bool MissileShouldClip() const
 	{
 		return (i_compatflags & COMPATF_MISSILECLIP) ||
--- a/src/g_levellocals.h
+++ b/src/g_levellocals.h
@@ -868,10 +868,10 @@ public:
 
 	bool IsSavingAllowed() const
 	{
-		if (dmflags3 & DF3_NO_SAVE)
-			return false;
 		if (dmflags3 & DF3_YES_SAVE)
 			return true;
+		if (dmflags3 & DF3_NO_SAVE)
+			return false;
 		return !(flags3 & LEVEL3_SAVE_NO);
 	}
 
--- a/wadsrc/static/menudef.txt
+++ b/wadsrc/static/menudef.txt
@@ -2219,6 +2219,7 @@ OptionMenu GameplayOptions protected
 	Option "$GMPLYMNU_ALLOWJUMP",				"sv_jump", "JumpCrouchFreeLook"
 	Option "$GMPLYMNU_ALLOWCROUCH",				"sv_crouch", "JumpCrouchFreeLook"
 	Option "$GMPLYMNU_ALLOWFREELOOK",			"sv_freelook", "JumpCrouchFreeLook"
+	Option "$GMPLYMNU_ALLOWSAVE",			    "sv_save", "JumpCrouchFreeLook"
 	Option "$GMPLYMNU_ALLOWFOV",				"sv_nofov", "NoYes"
 	Option "$GMPLYMNU_BFGFREEAIM",				"sv_nobfgaim", "NoYes"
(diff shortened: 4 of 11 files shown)


wadsrc/static/menudef.txt
```diff
@@ -2178,7 +2178,7 @@ OptionValue  FallingDM
 	3, "$OPTVAL_STRIFE"
 }
 
-OptionValue  JumpCrouchFreeLook
+OptionValue  JumpCrouchFreeLookSave
```
Boondorl: The name here can't be changed since existing menudefs will likely be relying on it

libraries/Translation/engine/zdoom/en_GB.po
```diff
@@ -3110,6 +3110,9 @@ msgstr "Allow crouch"
 msgid "GMPLYMNU_ALLOWFREELOOK"
 msgstr "Allow freelook"
 
+msgid "GMPLYMNU_ALLOWSAVE"
```
Boondorl: These should be sent upstream to the engine Weblate. That can be done after it's merged, feel free to just use the language keys for now

g0mb4: Sure, thanks.

Boondorl:
Looks good so far. I couldn't tell from the quick review I did, but does this have a way to re-enable saves if a map has it disabled? I think, for accessibility reasons, there should always be a way to override this option, and preferably it would be located in the new Accessibility menu (even if it's a duplicate of an existing option elsewhere).

g0mb4:
> Looks good so far. I couldn't tell from the quick review I did, but does this have a way to re-enable saves if a map has it disabled? I think, for accessibility reasons, there should always be a way to override this option, and preferably it would be located in the new Accessibility menu (even if it's a duplicate of an existing option elsewhere).

Sure, it is under "Gameplay Options".

src/g_game.cpp
```diff
@@ -2414,6 +2423,12 @@ void G_DoSaveGame (bool okForQuicksave, bool forceQuicksave, FString filename, c
 		return;
 	}
 
+	if (!IsSavingAllowed ())
+	{
+		Printf (PRINT_HIGH, "Saving is not allowed.\n");
```
nashmuhandes: - Export this string to a LANGUAGE entry so that creators can customize it (as well as allowing it to be localized)
- I'd suggest a message tone that's less aggressive. Instead of "not allowed", go for "You cannot save the game right now.".

RicardoLuis0:
this is really nice, but it wouldn't close #15, since it doesn't allow zscript control of saving

RicardoLuis0:
(also you should remove the "closes #15" text from the main description or github itself will auto-close the issue on merge 😅)

src/g_game.cpp
```diff
@@ -2414,6 +2423,12 @@ void G_DoSaveGame (bool okForQuicksave, bool forceQuicksave, FString filename, c
 		return;
 	}
 
+	if (!IsSavingAllowed ())
```
RicardoLuis0: G_DoSaveGame should have an extra flag so that script-triggered saves (from ACS/ZScript) are still allowed through

RicardoLuis0 (changes requested):

RicardoLuis0:
and it should also ideally have a separate flag to control autosaves on map load, but that's just a minor thing

src/g_game.cpp
RicardoLuis0: (a flag which would be true for calls from G_DoAutoSave and false for any other calls, and if true allow saving even if IsSavingAllowed returns false)

g0mb4:
> this is really nice, but it wouldn't close #15, since it doesn't allow zscript control of saving

Ohh, I thought ZScript can set CVARs, my bad.

src/g_game.cpp
RicardoLuis0: (G_DoAutoSave is called from both the autosave line special, so ACS/map saves, and from zscript via Level.MakeAutoSave, though there's a bit of indirection through gameaction for both)

g0mb4: Okay, I'll try to add that.

RicardoLuis0:
> > this is really nice, but it wouldn't close #15, since it doesn't allow zscript control of saving
> 
> Ohh, I thought ZScript can set CVARs, my bad.

it can but having that be the only way to control it is not ideal

g0mb4:
> and it should also ideally have a separate flag to control autosaves on map load, but that's just a minor thing

Can you explain it?

src/doomdef.h
```diff
@@ -176,6 +176,8 @@ enum : unsigned
 	DF3_NO_COOP_ONLY_THINGS	= 1 << 5,	// Any Actor that only appears in co-op is disabled
 	DF3_REMEMBER_LAST_WEAP	= 1 << 6,	// When respawning in co-op, keep the last used weapon out instead of switching to the best new one.
 	DF3_PISTOL_START		= 1 << 7,	// Take player inventory when exiting to the next level.
+	DF3_NO_SAVE				= 1 << 8,	// Don't allow manual saving
+	DF3_YES_SAVE			= 2 << 8,
```
RicardoLuis0: this should be 1 << 9, not 2 << 8

g0mb4:
> > > this is really nice, but it wouldn't close #15, since it doesn't allow zscript control of saving
> > 
> > 
> > Ohh, I thought ZScript can set CVARs, my bad.
> 
> it can but having that be the only way to control it is not ideal

Sure, I can add custom ZScript functions if needed, `ForceQuickSave()` ?

src/doomdef.h
g0mb4: Yup, it is strange, just followed the pattern :).

RicardoLuis0:
> > > > this is really nice, but it wouldn't close #15, since it doesn't allow zscript control of saving
> > > 
> > > 
> > > Ohh, I thought ZScript can set CVARs, my bad.
> > 
> > 
> > it can but having that be the only way to control it is not ideal
> 
> Sure, I can add custom ZScript functions if needed, `ForceQuickSave()` ?

not a function to save - a function to enable and disable saving without needing to use the cvars (since dmflags don't persist through saves)

RicardoLuis0:
> > and it should also ideally have a separate flag to control autosaves on map load, but that's just a minor thing
> 
> Can you explain it?

just a mapinfo flag to disable the automatic autosave

it happens here:
https://github.com/UZDoom/UZDoom/blob/ceec0c52dc5f6c80a26374e5a9410abb092cc1fe/src/g_level.cpp#L1542-L1545

g0mb4:
> > Sure, I can add custom ZScript functions if needed, `ForceQuickSave()` ?
> 
> not a function to save - a function to enable and disable saving without needing to use the cvars (since dmflags don't persist through saves)

Okay, I think I understand. I'll see what can I do about it.

g0mb4:
> just a mapinfo flag to disable the automatic autosave
> 
> it happens here:
> 
> https://github.com/UZDoom/UZDoom/blob/ceec0c52dc5f6c80a26374e5a9410abb092cc1fe/src/g_level.cpp#L1542-L1545

So like `noautosave` that disables only the autosave, and `nosave`  that disables every save?

RicardoLuis0:
> > just a mapinfo flag to disable the automatic autosave
> > it happens here:
> > https://github.com/UZDoom/UZDoom/blob/ceec0c52dc5f6c80a26374e5a9410abb092cc1fe/src/g_level.cpp#L1542-L1545
> 
> So like `noautosave` that disables only the autosave, and `nosave` that disables every save?

might be better as `noautoautosave`, since it'd only disable the level-load autosave, not autosaves in general

g0mb4:
Okay, I summarize the scripting changes needed, just to see if I understood correctly.

Function for ZScript: `EnableSaving()`, `DisableSaving()`.
These functions will allow the script to enable/disable the manual saving.

Function for ZScript: `EnableScriptSaving()`, `DisableScriptSaving()`.
These functions will allow the script to implement saving without allowing the user to save.

RicardoLuis0:
> Okay, I summarize the scripting changes needed, just to see if I understood correctly.
> 
> Function for ZScript: `EnableSaving()`, `DisableSaving()`. These functions will allow the script to enable/disable the manual saving.
> 
> Function for ZScript: `EnableScriptSaving()`, `DisableScriptSaving()`. These functions will allow the script to implement saving without allowing the user to save.

script saving should always be enabled

src/g_game.cpp
```diff
@@ -197,6 +197,8 @@ FString         savegamefile;
 FString         savedescription;
 FString         BackupSaveName;
 
+bool            savefromscript = false;	// if true, the savegame was created from a script
```
RicardoLuis0: this shouldn't be done as a global, just as a parameter to G_DoSaveGame

RicardoLuis0 (changes requested):

src/g_game.cpp
g0mb4: I see, okay.

src/g_levellocals.h
```diff
@@ -861,6 +861,53 @@ struct FLevelLocals
...
+	//==========================================================================
+	//
+	//
+	//==========================================================================
+
+	void DisableSaving()
+	{
+		dmflags3 = dmflags3 | DF3_NO_SAVE;
+		dmflags3 = dmflags3 & ~DF3_YES_SAVE;
+	}
```
RicardoLuis0: again, the script save enable/disable should not use dmflags (because those are user options)

(i'd rather it be absent from this PR at all than it using dmflags)

RicardoLuis0 (changes requested):

src/g_levellocals.h
RicardoLuis0: (not asking for it to be removed specifically, but for it to be either implemented without dmflags, _or_ removed,  it being absent would not be a blocker for this PR being accepted)

g0mb4: Yep, I though of using a simple CVAR+levelflags, I just wanted to follow the pattern of nojump etc.

g0mb4: What would you suggest?

RicardoLuis0: > What would you suggest?

just as a field in FLevelLocals

and make sure that it's saved by adding it in `FLevelLocals::Serialize` (https://github.com/UZDoom/UZDoom/blob/ceec0c52dc5f6c80a26374e5a9410abb092cc1fe/src/p_saveg.cpp#L988) and that it's initialized by adding it in `FLevelLocals::DoLoadLevel` (https://github.com/UZDoom/UZDoom/blob/ceec0c52dc5f6c80a26374e5a9410abb092cc1fe/src/g_level.cpp#L1389) (but only if the `newGame` flag is set, so that it persists between level transitions)

RicardoLuis0: but also feel free to alternatively just remove the functions entirely, it can always be implemented later

g0mb4: Thanks! I'm already invested in it, so I'll rather do it right.

g0mb4: Well, I gave it some thought and I think you are right. I'll remove the script functions, it needs a better architecture. On the one hand it is a level parameter, since it is in the MAPINFO but on the other hand it is a user choice since the user can override it. I do not really understand why it is different from `nojump`, which is handled through `dmflags`.

RicardoLuis0: > Well, I gave it some thought and I think you are right. I'll remove the script functions, it needs a better architecture. On the one hand it is a level parameter, since it is in the MAPINFO but on the other hand it is a user choice since the user can override it. I do not really understand why it is different from `nojump`, which is handled through `dmflags`.

nojump doesn't need swapping mid-level, levels are either made with jump in mind, or without it, but levels may want to switch saves on/off based on safe rooms/areas/etc, which, if it needs to be done via dmflags, would override user preferences

g0mb4: Thank you for the explanation! I won't add any new feature to this PR. I'll create a new for the scripting functions if this is merged.

RicardoLuis0 (dismissed):
looks good, won't be merged now since 5.0 is in feature freeze though (also, if you could clean up the commit history that'd be nice, but it won't be a blocker for merging)

g0mb4:
> if you could clean up the commit history that'd be nice

Sure. I'll squash them into 1 if this is ok.

g0mb4:
Where can I find the official documentation? I'd like to update it when this PR is merged.

RicardoLuis0:
> Where can I find the official documentation? I'd like to update it when this PR is merged.

https://zdoom.org/w/index.php?title=Main_Page

RicardoLuis0:
might need to request a wiki an account on the forums, open registration is disabled due to bot abuse

https://forum.zdoom.org/

g0mb4:
> > Where can I find the official documentation? I'd like to update it when this PR is merged.
> 
> https://zdoom.org/w/index.php?title=Main_Page

Thank you!

nashmuhandes (dismissed):

src/g_game.cpp
RicardoLuis0: there might already be a string for that as well, i think there are other instances where the game can refuse to save (ex. if a save is already pending for the current tick)

g0mb4: thanks! I'll change the text

nashmuhandes:
Just to re-iterate; looks good, but as has been said before, currently in feature-freeze so this'll only get merged after 5.0

I do have one question for @g0mb4 though - any chance you could add similar features but for the automap instead? (as a separate PR, most likely, heh)

EDIT: in addition to paralleling the mechanisms of this PR; there should additionally be some scripting functions to toggle the automap on or off

Boondorl:
> I do have one question for @g0mb4 though - any chance you could add similar features but for the automap instead? (as a separate PR, most likely, heh)

I'd prefer not to sidetrack this PR with discussion about a different feature

mc776:
> and preferably it would be located in the new Accessibility menu (even if it's a duplicate of an existing option elsewhere).

Seconding/bumping this, this seems the most logical place for this kind of override

the-phinet:
> Just to re-iterate; looks good

Head's up, your change request is blocking the merge of this. You'll need to mark it as resolved. @nashmuhandes

RicardoLuis0:
> > Just to re-iterate; looks good
> 
> Head's up, your change request is blocking the merge of this. You'll need to mark it as resolved. @nashmuhandes

IIRC that wouldn't be enough, needs to be either dismissed or approved

g0mb4:
> Just to re-iterate; looks good, but as has been said before, currently in feature-freeze so this'll only get merged after 5.0

Sure, I'm aware.

> any chance you could add similar features but for the automap instead? (as a separate PR, most likely, heh)

I'd like to finish the ZScript related features (turn it on/off/default) of this first (next PR), but yeah, I could do that as well.

> EDIT: in addition to paralleling the mechanisms of this PR; there should additionally be some scripting functions to toggle the automap on or off

Sure.

PR #1539, opened 2026-06-30
Tweaked double click values

Boondorl:
Currently the default time window for double tap binds is 250ms. While this is fine for the double tap execution, it means the single tap command on it must wait at least this long before it can also execute. To try and help alleviate this gap, I did some testing to find a better balance and landed on around 225ms. While this can go lower, it depends on the level of deliberateness we want to enforce as I found values lower than 200ms to be noticeably intense. As a reminder, an accessibility option is included to fine tune this manually, this is just to get the most "useful" default out of the box.

Also included are some tweaks to the menu ranges with the max being reduced reduced to a more reasonable 600ms (the original value before the rework was a bit over 500ms), though this can easily be changed back if desired. I found increments of 25ms instead of 50ms to give better control when adjusting via keyboard as 50ms can be too large at times.

There's likely ways this can be improved to more dynamically tell what a player wants to do, but seeing as 5.0 is the first time this feature is being exposed proper, that can wait until later. In general I'd like to expand input functionality to also allow things like hold binds and possibly bringing back the old hidden double tap system.

## Checklist

### DIFF
--- a/src/common/console/c_bind.cpp
+++ b/src/common/console/c_bind.cpp
@@ -136,7 +136,7 @@ const char *KeyNames[NUM_KEYS] =
 	"Guide",       "Pad_Misc",   "Pad_Touchpad",               //
 };
 
-CVAR(Int, cl_doubleclickthreshold, 250, CVAR_ARCHIVE | CVAR_GLOBALCONFIG)
+CVAR(Int, cl_doubleclickthreshold, 225, CVAR_ARCHIVE | CVAR_GLOBALCONFIG)
 
 FKeyBindings Bindings;
 FKeyBindings DoubleBindings;
--- a/wadsrc/static/menudef.txt
+++ b/wadsrc/static/menudef.txt
@@ -2934,10 +2934,10 @@ OptionMenu AccessibilityOptions protected
 	Tooltip "$ACCMNU_TOOLTIP_LIGHT_EXTRA"
 	Slider "$ACCMNU_LIGHT_RADIUS", "r_visibility", 0.0, 16.0, 1.0, 1
 	Tooltip "$ACCMNU_TOOLTIP_LIGHT_RADIUS"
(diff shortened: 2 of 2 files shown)


MajorCooke:
> In general I'd like to expand input functionality to also allow things like hold binds and possibly bringing back the old hidden double tap system.

Would that even include things like control, shift and alt? 

Also, thank you for making this.

Boondorl:
> Would that even include things like control, shift and alt?

Yes, things like that could be on the docket as well since I know there's a lot of use for key modifiers, though that'll likely require some special handling to differentiate a key modifier from someone just pressing shift to run lol. We have to be careful about what we allow since something like that could impact general play unlike this which only affects things that specifically have double tap binds.

This feature has actually existed for a really long time, it was just never properly documented or exposed in menudef. Dynam1k recently reworked it because we found the previous implementation to be pretty lackluster and not really useful for general modding appeal. It'll need further clean up later since this new one still isn't fully there imo, but is a good launching point.

RicardoLuis0:
> I found anything below 150ms to be extremely uncomfortable so have set that to the minimum

reaction times depend per person though, i'd leave the minimum as 100ms still

Boondorl:
While testing at 100ms I had to swap to using my right hand to mash fast enough and even then it was still inconsistent. Like, it might be possible, but I genuinely think it wouldn't be very reasonable to do it casually lol. 150ms is where I could still somewhat reasonably do it with my left (non-dominant) hand

RicardoLuis0:
> While testing at 100ms I had to swap to using my right hand to mash fast enough and even then it was still inconsistent. Like, it might be possible, but I genuinely think it wouldn't be very reasonable to do it casually lol. 150ms is where I could still somewhat reasonably do it with my left (non-dominant) hand

again, still depends on the person, and even on the keyboard (some gaming keyboards have mechanical switches optimized for minimal travel/etc), i see no reason to raise the minimum when it's something that the user themselves is meant to tweak to their liking

RicardoLuis0 (approved):

PR #1550, opened 2026-07-04
Fix mouse4 focus bug

### LINKED ISSUE #694: Hitting 'mouse 4' to close any in-game menu corrupts Windows focus
### UZDoom Version

UZDoom 4.14.3

### Game

All

### Operating System

Windows 11

### OS Version

Windows 11 10.0.22631 Build 22631

### Hardware

NVIDIA RTX 5070-Ti (driver version: 581.15 and 591.44)

### Duplicate Confirmation

### Bug Description

After replicating this bug, it's impossible to click anything on your desktop, but keyboard works.

When using a powershell script to figure out who/what holds the focus when clicking, it shows PID 0.

This bug persists even if you close the process and I can reproduce it with a 99% consistency.

Closing the game allows me to click on one window, but does not allow me to click on any other windows, nor does it let me click taskbar.

### Replication

- Open menu **while in a level**.
- Press **mouse 4** to close it.
- Option A: ALT+TAB: nothing on the desktop can be clicked on (including the taskbar)
- Option B: close the game: nothing can be clicked on besides one lucky window.

This problem persists until you use CTRL+ALT+DEL to open the "security options" and click with your mouse (you don't need to open task manager or anything, just clicking is fine).

This was reproduced on two different Windows 11 computers with different configurations and even two different source ports (fork of VkDoom).

**Using 'escape key' or the top left corner button, will not trigger this bug. It must be 'mouse 4'.**

### Configuration

```ini
defaults (but it seems to make no difference)
```

### Log

_No response_

heiko123abc:
## Checklist

Fixes #694.

The issues seems to reside with `RIDEV_CAPTUREMOUSE` in the grab function. According to https://stackoverflow.com/a/25398731 simply removing it and manually creating a rect to let the mouse focus to solves the issue. I have tested this with mouse 4 and the issue vanished.

### DIFF
--- a/src/common/platform/win32/i_mouse.cpp
+++ b/src/common/platform/win32/i_mouse.cpp
@@ -524,17 +524,25 @@ void FRawMouse::Grab()
 
 		rid.usUsagePage = HID_GENERIC_DESKTOP_PAGE;
 		rid.usUsage = HID_GDP_MOUSE;
-		rid.dwFlags = RIDEV_CAPTUREMOUSE | RIDEV_NOLEGACY;
+		rid.dwFlags = RIDEV_NOLEGACY;
 		rid.hwndTarget = mainwindow.GetHandle();
 		if (RegisterRawInputDevices(&rid, 1, sizeof(rid)))
 		{
 			GetCursorPos(&UngrabbedPointerPos);
-			Grabbed = true;
-			SetCursorState(false);
+			ClipCursor(NULL);
+
+			// Manually create the window rect for the cursor to grab
+			RECT rect;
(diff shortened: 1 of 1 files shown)


Boondorl:
Ooooo, this is a spicy fix. We were looking for this and assumed it was something buried deep in the Win32 code, so this is fantastic. This issue goes even further and closing a menu with any mouse button caused focus to break on Windows specifically, so hopefully this will fix all of that.

Boondorl (approved):

Boondorl:
Can confirm this fixes all instances of it bugging out. Didn't seem to have any side effects I could see from a quick test.

PR #1551, opened 2026-07-04
Ignore missing SNDINFO defs from the internal pk3

Boondorl:
These are often defined in case an outside resource needs it, so don't report these. Also quiets warnings from a few other possible locations that exist outside of SNDINFO.

## Checklist

### DIFF
--- a/src/sound/s_advsound.cpp
+++ b/src/sound/s_advsound.cpp
@@ -407,11 +407,11 @@ DEFINE_ACTION_FUNCTION(DObject,S_GetLength)
 // lump. Otherwise, adds the new mapping by using S_AddSoundLump().
 //==========================================================================
 
-FSoundID S_AddSound (const char *logicalname, const char *lumpname, FScanner *sc)
+FSoundID S_AddSound (const char *logicalname, const char *lumpname, FScanner *sc, bool warnMissing)
 {
 	int lump = fileSystem.CheckNumForFullName (lumpname, true, ns_sounds);
 
-	if (developer >= DMSG_WARNING && lump <= -1)
+	if (warnMissing && developer >= DMSG_WARNING && lump <= -1)
 	{
 		if (sc)
 		{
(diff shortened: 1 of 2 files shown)


the-phinet:
Even though the warnings are expected here, I don't think we should just be hiding them. I would change internal missing (expected) definitions to print to `DMSG_SPAMMY`.

Looking at the cause of the re-print, it seems the [entire sound engine is flushed](https://github.com/UZDoom/UZDoom/blob/e3aca285b3c23831b8f592cb589c34b866499efc/src/g_game.cpp#L1179) when the console is opened from the main menu. This seems like an issue (or a hack at best) to me.

Boondorl:
To clarify, these are intentionally missing and I don't think there's a way around that; many of them are for dehacked resources. Essentially you're going to be guaranteed flooded with these messages in almost every instance and I think that arguably makes them less than useful and will simply make people turn it off entirely. The engine pk3 of a standalone game should realistically not be including these at all (see: Disdain putting all of it in its own separate ipk3) so this is really more of a UZDoom-ism than anything because it has to support these esoteric things that standalones won't.

the-phinet:
Understood. I still believe it should be logged in the extremely verbose mode.

```cpp
if (lump <= -1)
{
    DPrintf(warnMissing? DMSG_WARNING: DMSG_SPAMMY, PRINT_NONOTIFY,
        TEXTCOLOR_ORANGE "Player sound " TEXTCOLOR_WHITE "%s, %s" TEXTCOLOR_ORANGE " - Lump doesn't exist: " TEXTCOLOR_WHITE "%s\n",
        pclass, soundEngine->GetSoundName(refid), lumpname);
}
```

Boondorl:
I think the sound flushing every time the full console is opened might be a bit of a problem for this, mainly because it will make SPAMMY useless as it gets flooded with hundreds of messages every time it goes to it. Considering someone might be using this to track down some esoteric behaviors, I think we'll need to make a task to fix this first, then we can trivially switch it over.

the-phinet (approved):

PR #1566, opened 2026-07-07
update-subtrees.sh: provide simple linear commit history instead of merge commits

charlesthobe:
~This PR is meant to be merged via "Rebase and merge" option because it also fixes ZMusic subtree tracking~
Edit: Squash and merge for this PR is fine as long as the commit message, specially the bottom of it isn't touched.

With this change, now "Sync subtrees" PRs can be safely merged via "Rebase and merge" option, which preserves git subtree tracking, so the maintainer/s doesn't have to reset subtrees each time they want to open a PR to update them.

## Checklist

### DIFF
--- a/tools/update-subtrees.sh
+++ b/tools/update-subtrees.sh
@@ -7,8 +7,21 @@
 
 export GITROOT="$(git rev-parse --show-toplevel)"
 
-pull() {
-	name=${1}; shift; dest=${1}; shift; repo=${1}; shift; ref=${1}; shift
+if [[ "${1}" == "--allow-empty" ]]
+then
+	EMPTY_COMMIT_FLAG="--allow-empty"
+	shift
+fi
+if [[ "${1}" == "add" || "${1}" == "pull" ]]
+then
+	operation="${1}"
+	shift
+else
+	operation="pull"
+fi
+
+subtree_action() {
+	name="${1}"; shift; dest="${1}"; shift; repo="${1}"; shift; ref="${1}"; shift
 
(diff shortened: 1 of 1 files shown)


the-phinet (approved):
Looking at this, I'm a little weary about a couple things that I'll note below. But I'm still going to merge this for 2 reasons: The only people running this script are maintainers, and I hope they know the risks and are able to troubleshoot; and that there is a plan to eventually replace the subtrees with something less awful.

Things Marcus is worried about:
- This requires bash 5.1, which is really new. I know there are still folks who contribute running ancient Debian boxes.
- I don't know how this will fair run from a mac or a windows box.
- The HEAD^2 assumption feels like it might be naive, mixed with the history re-writing, I'm not confident that this will work in all configurations

charlesthobe:
1- I didn't realize it needed a newer bash version, I'm gonna make an ubuntu 22.04 chroot via debootstrap and test in it, and change any new syntax to work on older bash.

2- I think as long as it uses bash it will run, I might change the shebang to #!/bin/env so if there was a newer bash version installed somewhere else on the system it will be used instead.

3- HEAD^2 points to the second parent of the commit, which is always the commit with the message that contains all the changes when you use `git subtree` with `--squish` flag

charlesthobe:
or it could be merged right away lol

the-phinet:
Ehh, I'm not that worried about it (and I need to push some changes to the subtrees)

Annoyingly, github decided to drop the empty commit when merging this. I had no control to prevent that. The history should get corrected anyways the next time the script is used

PR #1574, opened 2026-07-08
Fix lock on Intermission screens when Chat is open

RageSpark:
This bug can happen in two occasions afaik:

1. Open the Console and type 'messagemode' - Done, you won't be able to do anything.
2. In Multiplayer, another player changes the level while you have the chat open - You won't able to interact, have to wait for the time to count down or the host to advance.

What this PR does is
- Give priority to the chat input if its open, letting you finish typing or close. (tho the message will only show up in the console)
- Blocks opening the chat if not on a level or in a demo.

## Checklist

### DIFF
--- a/src/ct_chat.cpp
+++ b/src/ct_chat.cpp
@@ -536,7 +536,7 @@ static bool DoSubstitution (FString &out, const char *in)
 
 CCMD (messagemode)
 {
-	if (menuactive != MENU_Off)
+	if (gamestate != GS_LEVEL || demoplayback || menuactive != MENU_Off)
 		return;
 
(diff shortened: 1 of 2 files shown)


Boondorl (dismissed):

Boondorl:
You might need to make sure this properly disables chat on the title screen/demos as well. It seems like the main oversight here was intermissions being moved to cutscenes and the screenjob eating the input. Moving this check so early could break this instance where it absolutely cannot run due to the fact that demo commands aren't supported in this state of the game

the-phinet (approved):

PR #1581, opened 2026-07-09
Add ``-complevel`` CLI parameter

### LINKED ISSUE #324: Add `-complevel` CLI param, applying `COMPLVL` lump compatibility flags
### Game

All

### Problem Description

Most users are already familiar with the `-complevel` param -- and most mappers still severely under-utilize the existing `COMPLVL` lump, even as way of improving compatibility between other ports and UZDoom.

### Feature Description

The `-complevel` should be able to read the major known CL values:
 - 2 for vanilla Doom 1.9
 - 3 for vanilla Ultimate Doom
 - 4 for vanilla Final DOom
 - 9 for Boom
 - 11 for MBF
 - 21 for MBF21

And hopefully the string values as aliases, matching the strings from `COMPLVL`, as well as "1.9" , "ultimate" and "final".

### Additional Media

_No response_

RageSpark:
Closes #324 
Also accepts COMPLVL strings

## Checklist


RicardoLuis0 (dismissed):

src/d_main.cpp
```diff
@@ -323,6 +323,8 @@ FARG(bots, "", "", "",
 FARG(debug, "", "", "",
 	"");
 
+FARG(complevel, "", "", "", "");
```
the-phinet: Can you fill in the fields here? It's `FARG(name, section, short_description, usage, detailed_description)`

Maybe something like (I don't know all the valid levels): 

```cpp
FARG(complevel, "Configuration", "Sets engine compatibility level.", "boom",
	"Sets engine compatibility level (COMPLVL) to more closely follow quirks of other engines."
	"Valid options: 1,2,3,4,vanilla,mbf21,etc...");
```

the-phinet (changes requested):

src/d_main.cpp
RageSpark: sure thing

the-phinet (dismissed):

elf-alchemist:
Couple of things should be noted here:

The `COMPLVL` lump is only designed to expect the strings `vanilla`, `boom`, `mbf`, `mbf21`, in all the others ports that already implement it. Extending this set of strings further could certainly eventually lead to someone using the non-standard strings at some point and lead some folk to wonder why it doesn't work on whichever ports ran the wad that accidentally used it that way. Perhaps my wording on the linked issue was not ideal, I apologize.

The CLI option itself should take the actual numbers used in the demo/purist ports (2, 3, 4, 9, 11, 21), or optionally named aliases based on the (strict set of) `COMPLVL` strings, this'll cause a mismatch in programs which might expect the CLI params to otherwise behave the same.

P.S. complevel 1 refers to Doom 1.666 but there's hardly any actual difference that matters (compared to cl2, its just 2sided midtexetures not being animated), complevel 0 is a bit more interesting for demo ports, since 1.2 is a very early look into some of the subtle early development differences of the vanilla engine, but are otherwise not worth bothering about for most engines. None of the other complevel numbers matter outside of DSDA's intended goal of running as many demos as possible, there are (for example) many minute little PrBoom+ compat options for very specific demo cases that not even Woof has, so its really only really 2, 3, 4, 9, 11 and 21 that matter for now.

See the wiki page for additional references: https://doomwiki.org/wiki/Compatibility#Compatibility_levels

EDIT: Here's a pro tip, noone should be reading PRs while sleep-deprived. I misread the finer implementation loop as taking the internal enum value as input for the CLI, apologies, the param is fine. Point still stands about COMPLVL itself, though, it should only take the stricter subset of strings.

Boondorl (dismissed):

the-phinet (approved):

PR #1586, opened 2026-07-11
Added "keyboard" param to CheatGive

sierra0451:
"keyboard" is set to true if the cheat was put in with a cheat code (things like "idkfa" or "beholdX").

## Checklist

### DIFF
--- a/src/m_cheat.cpp
+++ b/src/m_cheat.cpp
@@ -248,19 +248,19 @@ void cht_DoCheat (player_t *player, int cheat)
 		break;
 
 	case CHT_IDKFA:
-		cht_Give (player, "backpack");
-		cht_Give (player, "weapons");
-		cht_Give (player, "ammo");
-		cht_Give (player, "keys");
-		cht_Give (player, "armor");
+		cht_Give (player, "backpack", 1, true);
+		cht_Give (player, "weapons", 1, true);
+		cht_Give (player, "ammo", 1, true);
+		cht_Give (player, "keys", 1, true);
+		cht_Give (player, "armor", 1, true);
 		msg = GStrings.GetString("STSTR_KFAADDED");
 		break;
 
 	case CHT_IDFA:
-		cht_Give (player, "backpack");
-		cht_Give (player, "weapons");
-		cht_Give (player, "ammo");
-		cht_Give (player, "armor");
+		cht_Give (player, "backpack", 1, true);
+		cht_Give (player, "weapons", 1, true);
+		cht_Give (player, "ammo", 1, true);
+		cht_Give (player, "armor", 1, true);
 		msg = GStrings.GetString("STSTR_FAADDED");
 		break;
 
@@ -289,7 +289,7 @@ void cht_DoCheat (player_t *player, int cheat)
 			{
 				if (i != 0)
 				{
-					cht_Give(player, BeholdPowers[i]);
+					cht_Give(player, BeholdPowers[i], 1, true);
 					if (cheat == CHT_BEHOLDS)
 					{
 						P_GiveBody (player->mo, -100);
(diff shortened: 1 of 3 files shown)


RicardoLuis0 (dismissed):
this won't work without changing the zscript function itself

sierra0451:
> this won't work without changing the zscript function itself

so i completely forgot to push that part of the changes i made. oops. they're working their way through CI now!

RicardoLuis0:
(will leave the actual review/approval to someone else as i have no idea whether this is useful or wanted as a feature at all, was just checking in on the technical side of things as i saw an issue)

the-phinet:
What is the motivation behind adding this?

sierra0451:
Honestly, the main reason I'm adding it is because I want to add something to a mod I'm working on where, because it removes the armor mechanic entirely at the moment, if the player tries to cheat armor in using `give armor`, it'll respond to that (I'm planning a simple console message, nothing crazy).

The basic functionality of it, I can get working easily, but the issue is that at the moment, it's impossible to separate that scenario from the same `give armor` call used internally by IDFA and IDKFA, so those cheat codes would also trigger it. This would allow me, and anyone else who'd like to do something similar, to fix that issue trivially.

RicardoLuis0:
> Honestly, the main reason I'm adding it is because I want to add something to a mod I'm working on where, because it removes the armor mechanic entirely at the moment, if the player tries to cheat armor in using `give armor`, it'll respond to that (I'm planning a simple console message, nothing crazy).
> 
> The basic functionality of it, I can get working easily, but the issue is that at the moment, it's impossible to separate that scenario from the same `give armor` call used internally by IDFA and IDKFA, so those cheat codes would also trigger it. This would allow me, and anyone else who'd like to do something similar, to fix that issue trivially.

wouldn't that be better solved by making those cheats go through zscript, instead of just adding a parameter to give?

sierra0451:
i'm not sure what exactly you mean by that if this isn't going through zscript- apologies, i'm just a little confused ^^;

sierra0451:
> i'm not sure what exactly you mean by that if this isn't going through zscript- apologies, i'm just a little confused ^^;

actually, thinking more about it, i think i understand- basically instead of adding a parameter, having the cheat code itself go through a zscript virtual that then calls CheatGive by default?

if i'm understanding correctly, that would make sense to me. might need a little time to figure out an _exact_ implementation for it if i were to do it myself, but i'd be down to take a crack at it at some point

RicardoLuis0:
> basically instead of adding a parameter, having the cheat code itself go through a zscript virtual that then calls CheatGive by default?

yeah, it'd give way more flexibility for mods as well

sierra0451:
> yeah, it'd give way more flexibility for mods as well

yee, makes sense to me!

Boondorl:
Revisiting this, the requested changes make no sense. This functionality already goes through ZScript, the issue is it's impossible to differentiate the keyboard cheats (idkfa, etc.) from console commands _because_ they go through the existing ZScript method. Catching cases of cheats going through the keyboard would be much simpler with the proposed solution than exporting the entire cheat logic structure to ZScript and then having to override every single individual case across every game to account for it.

RicardoLuis0:
> Revisiting this, the requested changes make no sense. This functionality already goes through ZScript, the issue is it's impossible to differentiate the keyboard cheats (idkfa, etc.) from console commands _because_ they go through the existing ZScript method. Catching cases of cheats going through the keyboard would be much simpler with the proposed solution than exporting the entire cheat logic structure to ZScript and then having to override every single individual case across every game to account for it.

yeah, but allowing custom keyboard cheats to be created is something that's good in general for mods, and it wouldn't be one virtual per cheat, it'd just be one virtual for all of them

RicardoLuis0:
ex.:
```
virtual void OnKeyboardCheat(Name cheatKeys)
{
    switch(cheatKeys)
    {
    case 'IDKFA':
        // ...
    }
}
```

Boondorl (approved):

PR #1589, opened 2026-07-12
update-subtrees.sh: macos fix and "usage" message

charlesthobe:
Now the script should work on macos bash 3.2 by removing "$RANDOM".
also the script now prints a help/usage message when invoked with no arguments or when invoked with `--help` or `-h`
Also the script should work on systems where bash isn't in `/bin` such as BSDs where it gets installed as `/usr/local/bin/bash` although this bit doesn't matter so much at least currently.

## Checklist

### DIFF
--- a/tools/update-subtrees.sh
+++ b/tools/update-subtrees.sh
@@ -1,9 +1,16 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
-# usage:
-# ./update-subtrees all
-# ./update-subtrees component
-# ./update-subtrees component commit
+print_usage() {
+	printf "Usage: "$0" [add | pull] (all | <component>) [<commit>]\n\n"
+	printf "\"pull\" is the default operation.\n\n"
+	printf "[<commit>] may not be set if \"all\" is specified\n"
+	exit
+}
+
+if [[ $# -eq 0 || "${1}" == "--help" || "${1}" == "-h" ]]
+then
+	print_usage
+fi
 
 export GITROOT="$(git rev-parse --show-toplevel)"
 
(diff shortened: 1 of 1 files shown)


charlesthobe:
I have a better idea, checking if the last commit hash has changed.
Man I wish git subtree pull would have a different exit status if the subtree was up-to-date, but it does make sense that it doesn't error out unless it fails (read/write permissions, bad internet,... etc)

charlesthobe:
~~Somehow it still doesn't work on macos bash 3.2
I'm gonna try ubuntu 22.04 and if the bash on it works I'm gonna make it a minimum requirement.~~
Edit: I was fetching the wrong branch in my mac VM, it does work fine on bash 3.2

charlesthobe:
Interesting coincidence the bash version on ubuntu 22.04 main repo with no upates is 5.1 :laughing:

charlesthobe:
~~While checking for a change in the commit hash is neater, checking for the exact merge commit message is safer, IMO
Realistically it shouldn't matter unless multiple git commands are issued simultaneously which  doesn't happen in any sane setup or even less than sane.~~
Edit: I changed my mind, neater is better.

tools/update-subtrees.sh
```diff
@@ -1,10 +1,17 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # usage:
 # ./update-subtrees all
 # ./update-subtrees component
 # ./update-subtrees component commit
 
+if (( ${BASH_VERSINFO[0]} < 5 || ${BASH_VERSINFO[0]} == 5 && ${BASH_VERSINFO[1]} < 1 ))
```
charlesthobe: && take precedence over ||

the-phinet (approved):

PR #1591, opened 2026-07-12
Allow manual highcontrast setting for launcher window

dileepvr:
Expanded `Int` `CVar` `ui_theme` to have 'high-contrast dark' and 'high-contrast light' values. Only make system `gdbus` calls if `ui_theme == 0` ('auto' setting).

I don't know how much of a general problem this is, but on my franken-linux machine, gdbus timesout. So the two `gdbus` calls put together for theme and contrast mode queries results in a very slow startup.

## Checklist


the-phinet (approved):
Nice :)

PR #1592, opened 2026-07-12
Remove Vertical Spread player option

### LINKED ISSUE #1531: Remove vertical spread player option
This has no real place in the engine as it can trivially be done as a mod and already only affects vanilla weapons. Only features that are impossible to properly modify from ZScript at this current time should be considered, especially if it universally impacts behaviors regardless of game or mod loaded.

RageSpark:
Closes #1531 

## Checklist


the-phinet (changes requested):
Instead of removing the CVARs, they need to be deprecated. You can use the `DEPR_CVAR` macro for that

something like:
```cpp
DEPR_CVAR(Flag, sv_novertspread, dmflags2, "Engine feature removed in favour of modding");
```

```cpp
DEPR_CVAR(Bool, vertspread, false, "Engine feature removed in favour of modding");
```

Boondorl (approved):

the-phinet (approved):

PR #1601, opened 2026-07-13
Use additive blending for the Disc of Repulsion effect

nashmuhandes:
(for consistency with all other effects in the rest of the game which DO use additive blending)... and allow it to be affected by UZDoom's configurable vanilla transparency option

## Checklist

## Remarks
While this may seem slightly contentious at first, I'm making the argument that every other effects-related Actors defined in Hexen's ZScript all use the `Add` RenderStyle, therefore the change in this PR is for visual consistency. Additionally, the +ZDOOMTRANS flag ensures the vanilla transparency option is respected - again, a flag that's on most (if not all) of the Hexen effects Actors.


Boondorl:
Will have to disagree with this one. Heretic and Hexen had transparency and imo that needs to be respected. This option should only be for things that ZDoom itself specifically changed the render style of (meaning pretty much exclusively for Doom actors). If other things are using Add, we should be scrutinizing it and fixing that if it's a change from vanilla Hexen or if it actually needs the +ZDOOMTRANS flag proper

the-phinet:
Seconding Boon here. Could you add screenshots to the OP?

PR #1609, opened 2026-07-15
Added -noverfication switch

Boondorl:
This is a temporary switch to disable file verification when playing online (engine verification is always on). This will later be reverted when `-optfile` can work as intended (currently optional files are only reserved for internal engine files). The main problem being that cvars and ACS modules are not handled properly, so that will need more time to fix.

Also adds in the missing description for `-password`.

## Checklist

### DIFF
--- a/src/common/engine/i_net.cpp
+++ b/src/common/engine/i_net.cpp
@@ -83,6 +83,8 @@ const char* neterror(void);
 #define neterror() strerror(errno)
 #endif
 
+// TODO: Remove noverification after optfile allows properly handling unverified files.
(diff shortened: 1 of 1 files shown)


the-phinet (dismissed):

the-phinet:
- having `hunter2` as the password example [would be so funny](https://web.archive.org/web/20101129230846/http://www.bash.org/?244321)
- The password example usage should probably convey that it's a password. `passw0rd!` or something would be better than `x`

mc776:
If there's a vote on this mine's for hunter2

the-phinet (approved):

TehRealSalt (approved):

PR #1612, opened 2026-07-17
Use semvers in engine, not just release notes

the-phinet:
Swaps engine over from using a free-form version string based in the git description into an [semver](https://semver.org/). It should look pretty similar to how things were before (`5.0.0-pre.477+39813d4` vs `5.0.0-pre-477-g39813d42cb`), but it comes with the added benefit of encoding more data. Removes some assumptions from the updater, hopefully keeping things future-proof.

- There's some mixing of old cstring code and newer stl code, but it's not a big deal.
- ~~The PR was tested with dummy data, but I'm not sure how to test with actual data. Some added scrutiny would be good (and any pointers on how to test)~~ Tested on my personal fork a bunch

---

Specific changes:
- VersionInfo now holds more data and is sorted by the rules of a semver
- VersionInfo holds the distance to the last tag. Distance is negative if the code has been modified. Distance holds the lowest precedence in the sort order.
- All instances of GIT_DESCRIPTION have been replaced with a semver
- Updater no longer uses the distance key to determine if it is an rc
- Updater no longer relies on github api

---

~~Question for consideration:~~
> ~~Currently the updater relies on github's api (which it really shouldn't) and makes twive the number of network requests as needed. The reason for this is that RCs are treated differently than release and preview builds. This would easily be resolved by not marking RCs as prerelease, and using the latest tag directly. I think this is the correct plan of action, but it would result in RCs being advertised on the front page of the repo (I don't think this is a big problem)~~

Implemented a system to deduce the next version strings

## Checklist


theshadsy:
> Currently the updater relies on github's api (which it really shouldn't) and makes twive the number of network requests as needed. The reason for this is that RCs are treated differently than release and preview builds. This would easily be resolved by not marking RCs as prerelease, and using the latest tag directly. I think this is the correct plan of action, but it would result in RCs being advertised on the front page of the repo (I don't think this is a big problem)

How often are RCs expected in the future? I get why they're rolling out for 5.0 because there's major feature changes/rewrites. But if the goal for UZDoom is to release more frequent, smaller updates rather than yearly major releases, that might make this a non-issue beyond the next month or so.

Boondorl:
I'm hoping that most releases will get them tbh. Ideally `trunk` should be the experimental branch that only people who want to test bleeding edge tech use and RCs will be the general preview for people who just want stuff early. I think the current environment hasn't been good for actually developing things and it's led to us making silly decisions because too many people are using `trunk` solely for the QoL features, not because they're interested in reporting issues.

TehRealSalt:
Unsure the best way to test this, so I just gave the CI's build to the Windows user in my house, enabled spammy messages, and clicked "Check for Updates". My prediction was that this would not do any actual updating, but I wanted to check that it's not receiving any unexpected values.

Clicking "Check for Updates" just results in the "Up to date!" message, as I was expecting. However, looks like `downloadName` and `downloadUrl` are printing weird garbage. I'd think it would be caught by the HAS_MEMBER checks, or print something normal? I haven't really been paying attention to how the updater works until now, so I could be missing something

image

---

> Currently the updater relies on github's api (which it really shouldn't) and makes twive the number of network requests as needed. The reason for this is that RCs are treated differently than release and preview builds. This would easily be resolved by not marking RCs as prerelease, and using the latest tag directly. I think this is the correct plan of action, but it would result in RCs being advertised on the front page of the repo (I don't think this is a big problem)

I think advertising RCs as latest is fine, personally

the-phinet:
> However, looks like downloadName and downloadUrl are printing weird garbage

Weird, you're right. I will investigate Sunday afternoon

---

It was just an issue with the log function itself. `DEBUG_LOG("%s", downloadName);` vs `DEBUG_LOG("%s", downloadName.c_str());`. I'm surprised it was even compiling.

TehRealSalt (dismissed):

RicardoLuis0 (dismissed):
fixed any issues i ran into, should be fine

the-phinet:
What is the purpose of this commit? [do NOT require semver tags when running without the updater](https://github.com/UZDoom/UZDoom/pull/1612/commits/dc84f55643e7ae945c6dae05d179a8f986953fc6)

If there is an issue in some build configurations, the solution is not to make a new code path, instead it is to fix the underlying issue.

It's harmless to add now, but it does need to be addressed in the future

RicardoLuis0:
> What is the purpose of this commit? [do NOT require semver tags when running without the updater](https://github.com/UZDoom/UZDoom/pull/1612/commits/dc84f55643e7ae945c6dae05d179a8f986953fc6)
> 
> If there is an issue in some build configurations, the solution is not to make a new code path, instead it is to fix the underlying issue.
> 
> It's harmless to add now, but it does need to be addressed in the future

github doesn't sync tags when you fork, anything relying on tags should be solely for the updater

Boondorl:
In this case it was throwing an assert on the version validation when the updater was turned off since the tag it was grabbing from created garbage. Since the only time you'd want this off is while devving, its relevancy to forks is very important

RicardoLuis0:
> In this case it was throwing an assert on the version validation when the updater was turned off since the tag it was grabbing from created garbage. Since the only time you'd want this off is while devving, its relevancy to forks is very important

not just the assert, it was returning `0.0.0` as the engine version, which would break a bunch of stuff

the-phinet:
> In this case it was throwing an assert on the version validation when the updater was turned off since the tag it was grabbing from created garbage. Since the only time you'd want this off is while devving, its relevancy to forks is very important

I see, thanks! I probably didn't notice since I'm pretty anal about keeping stale tags out of local repos. Should be an easy fix. I'll look into it after release, since Jay's workaround works.

> not just the assert, it was returning 0.0.0 as the engine version, which would break a bunch of stuff

lol that's what asserts are for: finding bugs. If the assert is failing in debug builds, the release builds have an issue there

Boondorl (approved):

PR #1625, opened 2026-07-25
Add TexMan.ResetAnimationTimer ZScript Function

XaserAcheron:
## Checklist

## What the is this?
I've been tinkering with using ANIMDEFS for ZScript UI stuff (Supplice terminals, to be specific), and it works pretty damn great, except for one missing piece: the ability to restart an animation from the beginning (e.g. for triggering an animation when the player opens a menu page, or somesuch). This PR adds a quick function to TexMan in ZScript to do just that, which smooths it out rather nicely without having to write an entire second animation system in ZScript to redo what ANIMDEFS already does.

[Side note: I could see this being extended to a general "SetAnimationTime" function, but that requires a bit more housekeeping, i.e. jumping to the middle of the animation sequence based on the specified time; I'm just keeping it simple for now.]

Feedback welcome, since I haven't tinkered with this area of code before and FTextureID is a bit fiddly to work with.

### DIFF
--- a/src/common/scripting/interface/vmnatives.cpp
+++ b/src/common/scripting/interface/vmnatives.cpp
@@ -39,6 +39,7 @@
 #include "image.h"
 #include "s_soundinternal.h"
 #include "i_time.h"
+#include "animations.h"
 
 #include "maps.h"
 #include "types.h"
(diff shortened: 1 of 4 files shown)


MajorCooke:
Sweet! I can definitely see this being very useful. Thank you!

PR #1630, opened 2026-07-27
OG Blood_Flade_Scalar levels implemented

### LINKED ISSUE #1576: Add an option to enabled the old-school RED RED pain flash for Doom games
### Game

Doom

### Problem Description

Bring back the blinding redness! please :)

### Feature Description

Y'know when you eat a rocket to the face you get a beautiful shade of painful red for like 5ish solid seconds? I miss that blinding red screen when taking mega damage, it really sold the pain. But in more modern ports like UzDoom here the redness is capped, so you can actually SEE. That's sacrilegious if you ask me, bring back the legacy red screen effect! I'm sure I won't regret asking for this later.. The blinding red screen effect is too funny and panic-educing to be forgotten to time! The default UzDoom effect should probably be kept as the default, blood flashbangs are probably an acquired taste

TL:DR, Add an option to make the red pain effect more true to Classic Doom, even when at the cost of visibility. See Chocolate Doom for examples. Oh and let it work with hardware rendering pls. Thanks! Oh and I already tried editing the .ini file to crank up the existing "Blood_Flade_Scalar" does not produce the desired effect, rip

### Additional Media

UzDoom Sacralige, odd amount of sight after eating an explosion
Image

Classic holyness, can't see shit
Image

Sempiternal-Futility:
## Checklist

Closes #1576

This one change here I believe we need to rethink: `V_AddPlayerBlend(player, blend, 0.5, 175);` -> `V_AddPlayerBlend(player, blend, 0.8, 225);`

The fourth parameter (175 -> 225) is "maxpainblend", which is fine to change as it only affects the red pain effect. However the third parameter (0.5 -> 0.8) is "maxinvalpha", which affects other effects, by being an opacity cap:

`// cap opacity if desired`
`if (blend[3] > maxinvalpha) blend[3] = maxinvalpha;`

I have tested other effects and everything appears to be unaffected, but I believe it would be a good idea to handle this a different way just to be sure, I just don't quite know how...

If anyone has any feedback/idea I'm very much open :)

EDIT: I preserved the opacity cap as 0.5 by using a new variable "bloodcnt", which creates an exception to the opacity cap only when the pain effect is at its most intense levels (aka when it's blinding the player)

### DIFF
--- a/src/rendering/2d/v_blend.cpp
+++ b/src/rendering/2d/v_blend.cpp
@@ -40,7 +40,7 @@
 #include "v_draw.h"
 
 CVAR(Float, underwater_fade_scalar, 1.0f, CVAR_ARCHIVE) // [Nash] user-settable underwater blend intensity
-CVAR( Float, blood_fade_scalar, 1.0f, CVAR_ARCHIVE )	// [SP] Pulled from Skulltag - changed default from 0.5 to 1.0
+CVAR( Float, blood_fade_scalar, 0.8f, CVAR_ARCHIVE )	// [SP] Pulled from Skulltag - changed default from 0.5 to 1.0 (changed again from 1.0 to 0.8)
 CVAR( Float, pickup_fade_scalar, 1.0f, CVAR_ARCHIVE )	// [SP] Uses same logic as blood_fade_scalar except for pickups
 CVAR(Float, powerup_fade_scalar, 1.0f, CVAR_ARCHIVE) // [Sal] Adjust screen fades for all inventory items
 
@@ -133,7 +133,7 @@ void V_AddPlayerBlend (player_t *CPlayer, float blend[4], float maxinvalpha, int
 
 	if (painFlash.a != 0)
 	{
-		cnt = DamageToAlpha[min (CPlayer->damagecount * painFlash.a / 255, (uint32_t)113)];
+		cnt = (DamageToAlpha[min (CPlayer->damagecount * painFlash.a / 255, (uint32_t)113)]) * 1.20; // "1.20" is here as a "20%" increase (default "blood_fade_scalar" 100% -> 80%, so 100% is now like the old-school red blend)
 
 		// [BC] Allow users to tone down the intensity of the blood on the screen.
 		cnt = (int)( cnt * blood_fade_scalar );
@@ -333,7 +333,7 @@ FVector4 V_CalcBlend(sector_t* viewsector, PalEntry* modulateColor)
 
 	if (player)
 	{
-		V_AddPlayerBlend(player, blend, 0.5, 175);
+		V_AddPlayerBlend(player, blend, 0.8, 225);
 	}
 
 	if (players[consoleplayer].camera != NULL)
--- a/src/rendering/2d/v_blend.cpp
+++ b/src/rendering/2d/v_blend.cpp
@@ -143,6 +143,9 @@ void V_AddPlayerBlend (player_t *CPlayer, float blend[4], float maxinvalpha, int
 			if (bloodcnt > maxpainblend)
 				bloodcnt = maxpainblend;
 
+			if (bloodcnt > maxinvalpha)
+				maxinvalpha = bloodcnt;
+
 			V_AddBlend (painFlash.r / 255.f, painFlash.g / 255.f, painFlash.b / 255.f, bloodcnt / 255.f, blend);
 		}
 	}
@@ -203,7 +206,7 @@ void V_AddPlayerBlend (player_t *CPlayer, float blend[4], float maxinvalpha, int
 	}
 
 	// cap opacity if desired
-	if (blend[3] > maxinvalpha && bloodcnt < 160) blend[3] = maxinvalpha; // "bloodcnt < 160" allows an exception to the red pain flash effect
+	if (blend[3] > maxinvalpha) blend[3] = maxinvalpha;
 }
 
 //==========================================================================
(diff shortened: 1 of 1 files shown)


src/rendering/2d/v_blend.cpp
```diff
@@ -333,7 +333,7 @@ FVector4 V_CalcBlend(sector_t* viewsector, PalEntry* modulateColor)
 
 	if (player)
 	{
-		V_AddPlayerBlend(player, blend, 0.8, 225);
+		V_AddPlayerBlend(player, blend, 0.5, 225);
```
RicardoLuis0, line 339: why are you lowering maxinvalpha to 0.5 and then ignoring it afterwards?

RicardoLuis0: lowering maxinvalpha while ignoring it for blood _will_ change how other screen effects render for the worse

RicardoLuis0: wait nvm i misread 🤦 (was looking at a single commit while i thought i was looking at the whole)

src/rendering/2d/v_blend.cpp
```diff
@@ -133,17 +133,17 @@ void V_AddPlayerBlend (player_t *CPlayer, float blend[4], float maxinvalpha, int
 
 	if (painFlash.a != 0)
 	{
-		cnt = DamageToAlpha[min (CPlayer->damagecount * painFlash.a / 255, (uint32_t)113)];
+		bloodcnt = (DamageToAlpha[min (CPlayer->damagecount * painFlash.a / 255, (uint32_t)113)]) * 1.20; // "1.20" is here as a "20%" increase (default "blood_fade_scalar" 100% -> 80%, so 100% is now like the old-school red blend)
```
RicardoLuis0: 80% of 120% is not 100%, it's 96%, this should just be done as `bloodcnt = (int)( bloodcnt * (blood_fade_scalar  + 0.2))` on the next line

src/rendering/2d/v_blend.cpp
```diff
@@ -203,7 +203,7 @@ void V_AddPlayerBlend (player_t *CPlayer, float blend[4], float maxinvalpha, int
 	}
 
 	// cap opacity if desired
-	if (blend[3] > maxinvalpha) blend[3] = maxinvalpha;
+	if (blend[3] > maxinvalpha && bloodcnt < 160) blend[3] = maxinvalpha; // "bloodcnt < 160" allows an exception to the red pain flash effect
```
RicardoLuis0: why 160? if the effect looks bad for certain damage levels the scaling should just be adjusted instead of hardcoding this

Sempiternal-Futility: hi ricardo, thanks for the review

&nbsp;

> why 160?

the bloodcnt works by decrementing bit by bit over the span of a few seconds, starting from ~225 (when the player has just taken damage) to 0. As the value lowers, the intensity of the effect does as well.

values between 225-160 are above the opacity cap of 0.5f. Once we reach values below 160 is when we can  turn the opacity cap on without blocking the effect.

&nbsp;

excuse me, but I don't quite understand what you meant by this:
>  if the effect looks bad for certain damage levels the scaling should just be adjusted instead of hardcoding this

What should we adjust the scale of?

adjusting the scale of bloodcnt would be futile, as the screen needs to have an opacity of 1.0f to "blind" the player, however the opacity cap prevents that, regardless of the scale of bloodcnt.

You could perhaps mean adjusting the scale of blend[3] to make it "more sensitive" (so that it produces more opacity with lower values), but that couldn't be it either because raising the cap from 0.5 -> 0.8 as I did earlier would achieve a similar result and be less messy, right?

Could you expand a little bit on what you meant?

&nbsp;

Again, pardon me I'm a bit slow

RicardoLuis0: > values between 225-160 are above the opacity cap of 0.5f. Once we reach values below 160 is when we can turn the opacity cap on without blocking the effect.

yeah, and when it goes past 160 it'll just jump from 0.5f to whatever random value 161 corresponds to, if 20% extra looks bad for values below 160 it should be adjusted from 20% to something else, not arbitrarily cut-off

Sempiternal-Futility: hello again

&nbsp;

there seems to be a little misunderstanding in our part. It will not go from 160 -> 161, instead it will go 160 -> 159, since it always goes from a greater value to 0.

&nbsp;

>if 20% extra looks bad for values below 160 it should be adjusted from 20% to something else

the problem with this is that adjusting 20% to a greater value won't work as the opacity cap will just limit the result, and adjusting 20% to something smaller will decrease the effect's intensity, which won't be enough to blind the player, therefore will not close #1576 .

&nbsp;

The best I managed to do was `if (blend[3] > maxinvalpha && bloodcnt < maxpainblend / 2) blend[3] = maxinvalpha;` which removes the magic number problem of 160, but this still is not quite enough to solve this problem, I know. I will need some more time, and will come with a solution, but I will need more time like I said.

Sempiternal-Futility: What do you think of this solution? [abb16b8](https://github.com/UZDoom/UZDoom/pull/1630/commits/abb16b862693a80877f3f24b35fefcac81d14415)

It is similar to ln. 112: `if (invalpha > maxinvalpha) maxinvalpha = invalpha;`

RicardoLuis0: that's better yeah

RicardoLuis0 (approved):

Lemono-o:
Is this in the new UZDoom v 5.0.0? If so that would be awwwweeesoommmee!! (Also, if so, where is the option lol)

RicardoLuis0:
> Is this in the new UZDoom v 5.0.0? If so that would be awwwweeesoommmee!! (Also, if so, where is the option lol)

no, it'll only be in future versions as it is still unmerged

TehRealSalt (approved):

the-phinet:
Am I good to clean up the commit history and to add you to the contributor list?

PR #1634, opened 2026-07-28
Skip remarks column when parsing csv

the-phinet:
Having remarks filled in was causing a crash in some instances

## Checklist


Boondorl (approved):

Boondorl:
Tested it on a couple mods I had, one with a regular LANGUAGE lump and one with the CSV format. Both seemed to boot up fine

PR #1646, opened 2026-07-30
Explicitly set X11 WM_CLASS/Wayland app_id

electricbrass:
## Checklist

This PR sets the Wayland app_id and X11 WM_CLASS to org.zdoom.UZDoom, and adds the StartupWMClass to the desktop file. The latter fixes issues on GNOME where, under certain circumstances, the desktop file will be unable to be matched to the window. Just setting the StartupWMClass isn't a stable solution, because SDL by default makes the app_id/WM_CLASS be the name of the executable (for example, the executable name changed from uzdoom.bin to uzdoom between 4.14.3 and 5.0-rc1), so it is best to explicitly set it instead. Reverse DNS names, just the same as what is recommended or required by many other Freedesktop and adjacent standards, is the recommended form for this to take.

BTW the CONTRIBUTING.md link doesn't seem quite right, it took me here while editing: https://github.com/UZDoom/UZDoom/compare/blob/trunk/CONTRIBUTING.md

### DIFF
--- a/src/common/platform/posix/sdl/i_main.cpp
+++ b/src/common/platform/posix/sdl/i_main.cpp
@@ -206,6 +206,8 @@ int main (int argc, char **argv)
 #endif
 */
 
+	// despite the name, this also sets the wayland app_id
+	SDL_SETENV("SDL_VIDEO_X11_WMCLASS", APPID);
 	if (SDL_Init (0) < 0)
 	{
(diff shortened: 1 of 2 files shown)


electricbrass:
I don't use KDE, so I don't know if there were issues with this, but if there are any issues currently where the top left corner of the window displays a generic Wayland/X11 icon on KDE, this should fix that as well.

the-phinet (approved):

PR #1647, opened 2026-07-30
Fixes for drag and drop file behavior in the launcher

### LINKED ISSUE #1641: [rc1] Drag-drop file on launcher, when it's in the "Release Notes" state, closes the app. (Windows 10)
Steps to repro:

- Get the launcher running in the "Release Notes" state (`i_display_new_release 2`)
- Click the "Game" tab
- Try to drag and drop a WAD on to the launcher

The app will just close.

(Windows 10 here)

Boondorl:
This fixes #1641 which was a stack overflow being caused by a recursive call. Also switches to `"` instead of `'` when appending file paths since I'm not entirely sure that was supported? I've only ever seen `"` used for encapsulating paths.

## Checklist

### DIFF
--- a/libraries/ZWidget/include/zwidget/core/widget.h
+++ b/libraries/ZWidget/include/zwidget/core/widget.h
@@ -248,6 +248,7 @@ private:
 	Widget* CaptureWidget = nullptr;
 	Widget* HoverWidget = nullptr;
 	bool HiddenFlag = false;
+	bool Dropping = false;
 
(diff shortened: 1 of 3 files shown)


the-phinet (approved):
Looks good to me. Single quotes work just fine (and are technically more correct, since the paths should not have variables expanded), but since the launcher doesn't understand variables (yet), I'm not worried about the change.

PR #1652, opened 2026-07-30
Debugger fixes

nikitalita:
* We shouldn't allow using the non-vendored cppdap, we rely on patches in our version that are still on hanging PRs
* Fixes normalizing windows paths that contain a drive name and forward slashes

## Checklist


nikitalita:
superseded by #1473

PR #1667, opened 2026-08-01
Fix RespawnDice logic + Add Dehacked entries

RageSpark:
## Checklist

### DIFF
--- a/src/playsim/p_mobj.cpp
+++ b/src/playsim/p_mobj.cpp
@@ -5149,7 +5149,9 @@ void AActor::Tick ()
 		if (Level->time & 31)
 			return;
 
-		if (pr_nightmarerespawn() > RespawnDice)
+		RespawnDice = clamp(RespawnDice, 0, 255);
+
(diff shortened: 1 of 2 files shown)


src/playsim/p_mobj.cpp
```diff
@@ -5149,7 +5149,9 @@ void AActor::Tick ()
 		if (Level->time & 31)
 			return;
 
-		if (pr_nightmarerespawn() > RespawnDice)
+		RespawnDice = clamp(RespawnDice, 0, 255);
```
Boondorl, line 5152: The type for this should be changed to `uint8_t` (`uint8` in ZScript) if it's only meant to be a [0, 255] range

src/playsim/p_mobj.cpp
```diff
@@ -5149,7 +5149,9 @@ void AActor::Tick ()
...
+
+		if (pr_nightmarerespawn() < 255 - RespawnDice)
```
Boondorl, line 5154: This can be `pr_nightmarerespawn() <= RespawnDice`

PR #1670, opened 2026-08-02
Make line-to-line portals facing a sky display the sky after portal recursion end

dileepvr:
Make line-to-line portals that face a sky floor or ceiling display the sky instead of being cleared to black when over the portal recursion limit.

Tested with 
[serial_portals.wad.zip](https://github.com/user-attachments/files/30626964/serial_portals.wad.zip)

Change value of `r_portal_recursions` higher or lower in the console.

## Checklist

### DIFF
--- a/src/common/rendering/gl/gl_renderstate.cpp
+++ b/src/common/rendering/gl/gl_renderstate.cpp
@@ -564,12 +564,12 @@ void FGLRenderState::EnableLineSmooth(bool on)
 //
 //
 //==========================================================================
-void FGLRenderState::ClearScreen()
+void FGLRenderState::ClearScreen(PalEntry color)
 {
 	bool multi = !!glIsEnabled(GL_MULTISAMPLE);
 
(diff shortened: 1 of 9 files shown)


RicardoLuis0 (dismissed):

dileepvr:
Did you still want this modified J? 
So far this only kicks in if the line portal line's frontsector has a sky floor or sky ceiling.
A mapinfo option that can clear to some other color is possible. It can be another PR though (distance culling is almost doing this. I don't think it covered the clearing after portal recursions though)

dileepvr:
Imma make the approval stale by pushing another commit that sets the fade color while accounting for distance. Basically repeat the fog density calculation for the clear color (if not sky that is).

RicardoLuis0 (dismissed):

MajorCooke:
Just a note, this isn't done yet. When testing it, all it did was draw the sky itself without any skybox or skymist. That part still needs fixing.

image

RicardoLuis0 (approved):

MajorCooke:
Tested and working, thank you!

PR #1674, opened 2026-08-02
Enable GLDEF materials for HUD/weapon PSprites

dileepvr:
Allow weapon PSprites to display GLDEFs material shader effects (normal map, specularity, and PBR) just like textures. Every layer can have an independent neutral normal direction. They respond to dynamic lights per pixel. The 2D image is being rendered in the world location in 3D with dept-filtering disabled.

Just define material in the `GLDEFS` lump for PSprites. Normal maps or specular maps. Default neutral normal direction is out of screen and tilted upward by 45 degrees to reflect visible sources of light into the camera.

Every layer has a `materialnormal` field which can be used (independently for each layer) as the neutral normal (about which normal map applies) as long as the new overlay flag `PSPF_USEMATNORMAL` is set. A new `A_OverlayMatNormal()` function has been defined to supply the normal to each layer.

You can basically mimic a 3d model as long as every face projected into the 2D plane is in its own layer.

Tested with windowed mode and arbitrary resolutions for checking onscreen PSprite location. Tested relative positioning of multiple layers. Tested pivoting and rotating.

Test map (better with lights.pk3 loaded): 
[pbr_PSprites.zip](https://github.com/user-attachments/files/30635230/pbr_PSprites.zip)

Demo video (I didn't have normalmap textures for any PSprites to use, so I modified a door and split it into three layers facing three different normal directions): https://youtu.be/1c22JP1eIYo

## Checklist


MajorCooke:
**YES!!!** Holy crap, I've been waiting for this for a long time!

src/playsim/p_pspr.cpp
```diff
@@ -895,6 +897,41 @@ DEFINE_ACTION_FUNCTION(AActor, A_OverlayPivot)
 	return 0;
 }
 
+//---------------------------------------------------------------------------
+//
+// PROC A_OverlayMatNormal
+//
+//---------------------------------------------------------------------------
+
+DEFINE_ACTION_FUNCTION(AActor, A_OverlayMatNormal)
```
Boondorl: This should be defined solely in ZScript if it's not accessing anything exclusive to the engine/used within it. I know the other functions are native, but this is probably just a relic from their development.

wadsrc/static/zscript/actors/actor.zs
```diff
@@ -1435,6 +1435,7 @@ class Actor : Thinker native
 	action native void A_OverlayPivot(int layer, double wx = 0.5, double wy = 0.5, int flags = 0);
 	action native void A_OverlayPivotAlign(int layer, int halign, int valign);
 	action native void A_OverlayVertexOffset(int layer, int index, double x, double y, int flags = 0);
+	action native void A_OverlayMatNormal(int layer = PSP_WEAPON, double nx = 0, double ny = 0.707, double nz = -0.707, bool setusenorm = true);
```
Boondorl: What do these defaults represent? It might also be better to use 0 here as the layer since I see that grabs the layer the function is called from which feels like it would be closer to the intended usage of it rather than purely for weapons.

dileepvr: Pointing out of the screen and tiled up by 45 degrees. In case someone sets the `PSPF_USEMATNORMAL` flag without choosing a normal and sees nothing happening.

src/playsim/p_pspr.cpp
dileepvr: Should I remove the `A_` prefix as well?

MajorCooke: Nah that can stay. Might as well keep it so it matches with the other overlay functions. They're all action functions after all which is what the `A_` is for.

PR #1677, opened 2026-08-03
Pbr for all sprites

dileepvr:
Presence of a normal map texture triggers the mode. No new flags. Every sprite (frame and rotation) is independent. Neutral normal always points to camera viewpoint in first commit.
This was asked for for testing purposes.

Tiny test map: 
[PBR_all_sprites.zip](https://github.com/user-attachments/files/30642289/PBR_all_sprites.zip)

## Checklist

### DIFF
--- a/src/rendering/hwrenderer/scene/hw_drawstructs.h
+++ b/src/rendering/hwrenderer/scene/hw_drawstructs.h
@@ -380,6 +380,7 @@ public:
 	int modelframeflags;
 	FRenderStyle RenderStyle;
 	int OverrideShader;
+	bool normspec;			// Whether or not the texture is PBR/normal+specular
 
 	FTranslationID translation;
 	int index;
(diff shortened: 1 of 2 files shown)


RicardoLuis0 (changes requested):
checking for just normal texture existence isn't really valid, because the render modes need more than one texture to be present:
normspec needs normal + specular
pbr needs normal + metallic + roughness + ao
otherwise it just renders without effects

https://github.com/UZDoom/UZDoom/blob/fbad53bff568997bc77be83320619df8fa47f6a9/src/r_data/gldefs.cpp#L1522-L1536

https://github.com/UZDoom/UZDoom/blob/fbad53bff568997bc77be83320619df8fa47f6a9/src/common/textures/hw_material.cpp#L76-L91

RicardoLuis0 (approved):

dileepvr:
Looks like #1535 is also setting the normal. We can probably combine both approaches, and choose one method over the other depending on `viewpoint.bDoOrtho`. Should I move this out of draft status?

dileepvr:
And an additional check for `FLATSPRITE` and `WALLSPRITE` flags.

MajorCooke:
> normspec needs normal + specular

That reminds me. In GLDefs, if a normal is present but specular isn't, it doesn't assume specular to be just black meaning it requires both defined. Can this be changed? Or does it have to be this way?

PR #1681, opened 2026-08-03
Switched to name-appending system for log files

Boondorl:
Ensure these are always created where the game is being executed from. Instead, use the name as a quick way to differentiate log files e.g. `+logfile abcd` will be created as `log-abcd.txt`.

## Checklist

### DIFF
--- a/src/common/console/c_enginecmds.cpp
+++ b/src/common/console/c_enginecmds.cpp
@@ -101,7 +101,15 @@ UNSAFE_CCMD (exec)
 
 void execLogfile(const char *fn, bool append)
 {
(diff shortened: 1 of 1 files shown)


the-phinet (commented):
Instead of `log-foo.txt`, can we do `foo.log`? `.log` is in general the preferred extension for plaintext logs, which these are

Boondorl:
My opinion is that if this PR fixes the issue it's purported to then it should be merged first and a task can be made to clean up these aspects later

nashmuhandes (approved):

PR #1682, opened 2026-08-03
Add missing support for NoPerPixelLighting to weapon models

nashmuhandes:
## Checklist


nashmuhandes:
This PR is technically ready for review, but as is customary with stuff like this, a test mod file is usually required, but it's 4 in the morning currently and I don't have time to put one together, therefore I'm marking this as a draft. Will undraft when the example file is ready

nashmuhandes:
Example file added. The first weapon slot (1) can be used to toggle between a PPL- versus a non-PPL weapon model.

[WeaponNoPerPixelLighting.zip](https://github.com/user-attachments/files/30686901/WeaponNoPerPixelLighting.zip)

Boondorl (approved):

PR #1698, opened 2026-08-07
Added file name sanitizing

Boondorl:
## Checklist

### DIFF
--- a/src/common/console/c_dispatch.cpp
+++ b/src/common/console/c_dispatch.cpp
@@ -181,6 +181,12 @@ static const char *KeyConfCommands[] =
 
 // CODE --------------------------------------------------------------------
 
+void C_SanitizeFileName(FString& file)
+{
+	file.ReplaceChars('.', '-');
+	file.ReplaceChars(':', '-');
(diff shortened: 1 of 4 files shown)


the-phinet (approved):

src/common/console/c_dispatch.cpp
```diff
@@ -181,6 +181,12 @@ static const char *KeyConfCommands[] =
 
 // CODE --------------------------------------------------------------------
 
+void C_SanitizeFileName(FString& file)
+{
+	file.ReplaceChars('.', '-');
+	file.ReplaceChars(':', '-');
+}
+
```
RicardoLuis0, line 189: shouldn't it also sanitize slashes? that way ex. symlinks can't be traversed

RicardoLuis0 (dismissed):

src/common/console/c_dispatch.cpp
Boondorl: Don't know, but I want local directories to still be possible for organization reasons

PR #1713, opened 2026-08-10
Fix linebreak issue in `echo`

### LINKED ISSUE #1708: cvar echo command has odd formatting
don't know if this is something worth bringing up, but i made a custom alias bind that's supposed to show an echo message when used, but the text is formatted in a weird way:

`alias mode_vanilla 		"echo Vanilla DOOM Gameplay;..."`

the result being this at the top corner:

> VANILLA
> DOOM
> GAMEPLAY

again, i'm not sure if it's just me or if it's a legitimate error in the new release candidate.

the-phinet:
Fixes #1708, introduced in 4cf08b9

There are some other functions that I suspect will have similar issues, too. I've been meaning to fix the linebreak inconsistencies anyways, so I'll tackle those when I have some time. Just so I don't lose it, this should find most if not all of the candidate issues: `grep -noP 'Printf\s*\(\s*"(?:[^"\x5c]|\x5c.)*"' src -r | grep -vF '\n' | cut -d : -f 1,2`

There is a deeper fix that will need doing, where string repeat buffer needs to only apply once a newline is encountered, but that can come in a future patch.

## Checklist

### DIFF
--- a/src/common/console/c_console.cpp
+++ b/src/common/console/c_console.cpp
@@ -1234,12 +1234,14 @@ CCMD (clear)
 
 CCMD (echo)
 {
-	int last = argv.argc()-1;
-	for (int i = 1; i <= last; ++i)
+	int count = argv.argc();
+	FString str = (count <= 1)? "": argv[1];
+	for (auto i = 2; i < count; i++)
 	{
-		FString formatted = strbin1 (argv[i]);
-		Printf ("%s%s", formatted.GetChars(), i!=last ? " " : "\n");
+		str.AppendCharacter(' ');
+		str += argv[i];
 	}
+	Printf("%s\n", str.GetChars());
 }
 
 CCMD(toggleconsole)


Boondorl (approved):

RicardoLuis0:
C++-side Printf (as opposed to zscript-side console.printf) as a whole is supposed to be safe for composing a line in multiple calls though, so this isn't addressing the root of the issue

RicardoLuis0:
what it needs to do is delay calling FNotifyBuffer::AddString until the accumulated buffer ends in `\n`

RicardoLuis0:
ex. add a new `AddStringPart` function that does said buffering and call that from `PrintString` instead of `AddString`

https://github.com/UZDoom/UZDoom/blob/08b19217a9950e7c3bcf106f32b89d4c48603d2f/src/common/console/c_console.cpp#L438

the-phinet:
Yes I know. I quite literally said that in the PR's description.

Also, please try to edit your comments instead of adding new ones. I really don't need my email spammed by a stream-of-consciousness string of comments.

RicardoLuis0:
> Yes I know. I quite literally said that in the PR's description.

yes, but i'm saying how to fix it so that it can be fixed properly from the start instead of adding workarounds to every function that calls printf before actually fixing the issue in some unspecified future time

Boondorl:
Yes, please don't comment multiple times in a row. This isn't a messenger app and it makes the email messages much messier. Also, there's no point commenting on what the PR needs to do to fix the root issue if the body of it already explicitly lays this out. Asking for a task to be created would be a more reasonable request to ensure it doesn't get lost.

Issue #1527, opened 2026-06-27
Assorted `wadsrc/` issues omnibus


elf-alchemist:
Various assets in the `wadsrc*/` folders have a variety of small related problems.

Many use mixed casing, either full upper case or full lowercase lump naming conventions.

While some are stored in the original doom patch format, others are stored as PNGs.

Sound effects have mixed usage in both FLAC and OGG extensions.

Various graphical assets are of dubious, if not outright copyrighted assets. `wadsrc/static/fonts/indexfont/` is an obvious recolor of the Doom yellow number font from the STBAR, while `wadsrc/static/graphics/AMMNUM*.png` looks to be straight up the gray number font used in the STBAR. `wadsrc/static/graphics/stkeys*.png` should be credited to Boom/TeamTNT, but should otherwise just be GPLv2. What is `wadsrc/static/sprites/SGRN*.png` even used for anyway? Is `wadsrc/static/sprites/pista0.png` from Romero's development asset dump or was it original? The ice shards in `wadsrc/static/sprites/icec*.png` look similar to Hexen's IWAD own set, but appear to be original, Zauberer/Lastermaul maybe? `wadsrc_extra/static/sprites/PLS*.png` is pulled from the Doom beta, and ostensibly used for the MBF-originated 'classic BFG' weapon function. `wadsrc_extra/static/sprites/minza0.png` is from Chex. `wadsrc_extra/static/sprites/GWANA0.png` is from Heretic. `wadsrc_extra/static/graphics/stkeys*.lmp` seems to be a duplicate of the previous stkeys set. All the unicode fonts in `wadsrc_extra/nonfree/filter` are in Doom patch format, with player crouch sprites also in patch format. Almost all graphics in `wadsrc_widepix` are in patch format.

PR #1722, opened 2026-08-12
Replace Doom beta sprites with freely licensed variants

elf-alchemist:
Addresses one point from #1527

Beta plasma balls 1 & 2 are based on Freedoom's normal plasma balls, as is used in Woof and DSDA.

EvilScepter and UnholyBible sprites created from scratch by Amaruq Wulfe, at my request, for the explicit purpose of inclusion in source ports -- as documented in https://github.com/fabiangreffrath/woof/pull/2768.

Screenshot_Doom_20260812_125155
Screenshot_Doom_20260812_125200

## Checklist


the-phinet (dismissed):

RicardoLuis0:
shouldn't this only replace it for non-BUILD_NONFREE builds? just outright replacing them with freedoom graphics will make mods that use those sprites clash visually otherwise (well, for the plasma sprites that is)

the-phinet:
> shouldn't this only replace it for non-BUILD_NONFREE builds?

I don't think so for 2 reason:
- non-BUILD_NONFREE shouldn't be a second-class citizen 
- I'd prefer to minimize any non-free stuff, in general

RicardoLuis0:
> * non-BUILD_NONFREE shouldn't be a second-class citizen

it can only be, if we're replacing the beta bfg sprites with freedoom sprites then any mod that relies on them to make the beta bfg _look like the damn beta bfg_ will just look wrong

mc776:
My only issue would just be that the main primary colour should be the same as the nonfree, just to keep any dynamic lights consistent. (Messing around with filters for this for Freedoom and forks has never not been a pain in the ass)

EDIT: real quick palette swap for the PLS2s
pls2e0
pls2d0
pls2c0
pls2b0
pls2a0

mc776:
What are the scepter and bible used for? I don't see what they replace.

RicardoLuis0:
> What are the scepter and bible used for? I don't see what they replace.

they don't replace anything IIRC

elf-alchemist:
> My only issue would just be that the main primary colour should be the same as the nonfree, just to keep any dynamic lights consistent. (Messing around with filters for this for Freedoom and forks has never not been a pain in the ass)
> 
> EDIT: real quick palette swap for the PLS2s

Updated this PR to use the red sprites above, out of an interest to not actively delay this PR until the next breaking changes dev cycle.

Personally, I think the green and blue sprites look much better, at any rate:
image
image

the-phinet (approved):

PR #1740, opened 2026-08-16
Fix MBF21 linedef flags on old maps when using `COMPLVL` patches

elf-alchemist:
Third charm's the time, right? #1516

Devalaous' QoL patches for old WADs uses `COMPLVL` to set vanilla mapping compatibility, my previous changes regarding the removal of `NOMBF21` have thus seemingly created a regression with the `COMPLVL` (though I have not been able to locally reproduce said problem at this moment), this should fix those.

## Checklist


RicardoLuis0 (approved):

the-phinet:
Would you be able to re-write the commit so when looking at a blame it's possible to see your reasoning? The body of the PR description would be fine I think

elf-alchemist:
> Would you be able to re-write the commit so when looking at a blame it's possible to see your reasoning? The body of the PR description would be fine I think

Pushed the commit again with a better message. Ideally the compat menu flags and COMPLVL flags would be merged to only have a single set, but I digress -- that'd be for 5.1 or what have you.

PR #1776, opened 2026-08-24
UDMF key clean-up/update. Remove unsupported VKDoom stuff, while updating/adding actually-supported ZDRay stuff

nashmuhandes:
… also adding lm_bounces and lm_gridsize that WILL be supported by the impending v1 of ZDRay for UZDoom

## Checklist


RicardoLuis0:
light_noshadowmap/light_dontlightactors/light_dontlightmap are still useful for uzdoom, just need to be ported back from vkdoom's udmf reader, and i do plan to add back inverse-square lights together with light_strength for the render rework

nashmuhandes:
Okay, I've reverted the edits to the stuff that you intend to work on in future.

RicardoLuis0:
oh right, with 3d shadowmaps we'll probably need light_shadowminquality as well (to allow turning off shadows for only some lights)

the-phinet:
I believe we should remove everything currently unused, and only add it back when it's actually implemented. Even if we're confident that it will get used in the future, it shouldn't be in the codebase right now.

nashmuhandes:
Finalized this.

nashmuhandes:
Closing this, it'll be folded into the lightmap feature PR eventually

PR #1792, opened 2026-08-29
Reuse bone cache across portals

MajorCooke:
I have a map with *many* tiles in an effort to create manipulatable terrain, but there was a lot of lag involving portals because every model would be recalculated per *pass*. 

But with this change, bones are no longer recalculated on each render pass - they're simply copied over, which will boost performance when portals are involved.

If proof is needed, [here's a demo mod.](https://drive.google.com/file/d/1UG7Qh5vHaFfvXkT6AEoAiS8HJHrXYWsS/view?usp=sharing)
It features one massive room with four line portals.
* Load MAP01
* Type the following in console:
  * `r_portal_recursions 2`
  * `tile`

(Recursions set to 2 is recommended because I don't want to accidentally fry anyone's hardware. It's very performance intensive. `tile` will spawn all the tiles.)

The models in question have `DistanceCheck` with a cvar set to 1024, changed their stats to a non-thinker, and have 255 bones in the tiles, but performance is significantly hampered due to portals. With the change to reusing the bone cache instead of recalculating every pass, performance is greatly increased.

I also tried this with smaller tiles (8 x 8 instead of 15 x 15 and it quadrupled the actor count), but that only made a miniscule improvement to performance.

## Checklist

### DIFF
--- a/src/common/rendering/v_video.h
+++ b/src/common/rendering/v_video.h
@@ -147,6 +147,7 @@ public:
 	HWViewpointBuffer *mViewpoints = nullptr;	// Viewpoint render data.
 	FLightBuffer *mLights = nullptr;			// Dynamic lights
 	BoneBuffer* mBones = nullptr;				// Model bones
+	unsigned int mBoneCacheGeneration = 0;		
 	IShadowMap mShadowMap;
 
 	int mGameScreenWidth = 0;
--- a/src/playsim/actor.h
+++ b/src/playsim/actor.h
@@ -750,6 +750,8 @@ public:
 	int							 flags;
 	int							 overrideFlagsSet;
 	int							 overrideFlagsClear;
+	unsigned int				 cachedBoneFrame;
+	int							 cachedBoneStartPos = -1;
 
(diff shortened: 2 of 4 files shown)


RicardoLuis0:
this will break animations across portals for every model that isn't using MDL_MODELSAREATTACHMENTS

RicardoLuis0:
it should be a bone index per model index, not global per actor (just a TArray<int> instead of an int) (and when model 0 is rendered for MDL_MODELSAREATTACHMENTS, it'll hit the cache and make evaluatedSingle true)

MajorCooke:
Ah, didn't think of that. I'll work on it.

RicardoLuis0:
(other than that it looks fine though)

MajorCooke:
Closing this down and starting a fresh pull since there's a whole bunch of PRs that were merged, and things are likely to get messy. *IF* I can figure this out. 

EDIT: I'm going to open an issue now actually, so awareness is made. I'll try to fix it but this took several weeks to figure out.

PR #1811, opened 2026-08-30
Handle long lines getting culled if vertices are far but line is near

### LINKED ISSUE #1809: Line-Culled Fog Problem with Long Lines
### UZDoom Version

5.0.1 dev

### Duplicate Confirmation

### Bug Description

#404 's line culling is undercut by the fact that it only checks a [line's distance to its vertexes](https://github.com/UZDoom/UZDoom/pull/404/changes#diff-e87c129a3a0ecbc0419c8a9d67a56f03d26434c5f336b556e1ead6acbd5b4219R262) - what it doesn't have is a special case where if the line is longer than either of the checks, it will undoubtedly cull the line if nearest the center.

### Replication

 [Here's a demo mod.](https://drive.google.com/file/d/1UG7Qh5vHaFfvXkT6AEoAiS8HJHrXYWsS/view?usp=sharing)
* Load MAP01
* Set fog mode to `Radial`
* Enable fog culling under `Options -> Display Options -> DSPLYMNU_CULLOPT, set the first option to CULLVAL_ON_WITH_FOG`
* Change `DSPLYMENU_LINECULLDIST` to 2000
* Type `tile` in console so you can see the ground
* Type `am_cheat 1`
* Walk along the portal lines. 

The line will be culled as soon as you get a certain distance away from both vertexes.

dileepvr:
Distance to segment vertices were being used for culling. This presented a problem when:
- incredibly long linedefs were present close to the viewpoint, but the vertices were far away
- linedefs were split into segments due to bsp, and a segment's vertices were far away resulting in the whole linedef getting rendered as skyflat 

Fixed by using the whole linedefs vertices for culling distance calculations, and computing shortest distance to the entire linedef if angle subtended by linedef at viewpoint is greater than 90-degrees. I could compute that every time but wanted to avoid excessive computation for every segment.
Closes #1809 

Fix cullcolor change applying to current level. 
~~Re-read skymist lump if skymisttoggle console command is run. (This wasn't a problem if skymist lump is specified in `MAPINFO`, but if not the internal one from uzdoom.pk3 was not being read on first `InitSkyMap()` for some reason. Maybe it is being called too early?)~~
Enforce skymist `xscale` to 1.0.

## Checklist

### DIFF
--- a/src/console/c_cmds.cpp
+++ b/src/console/c_cmds.cpp
@@ -1011,6 +1011,11 @@ CCMD(changesky)
 CCMD(skymisttoggle)
 {
 	primaryLevel->flags3 ^= LEVEL3_SKYMIST;
+	if (primaryLevel->flags3 | LEVEL3_SKYMIST)
+	{
+		primaryLevel->skymisttexture.SetNull();
+		InitSkyMap (primaryLevel);
+	}
 }
 
 //-----------------------------------------------------------------------------
--- a/src/console/c_cmds.cpp
+++ b/src/console/c_cmds.cpp
@@ -1011,11 +1011,6 @@ CCMD(changesky)
 CCMD(skymisttoggle)
 {
 	primaryLevel->flags3 ^= LEVEL3_SKYMIST;
-	if (primaryLevel->flags3 | LEVEL3_SKYMIST)
-	{
-		primaryLevel->skymisttexture.SetNull();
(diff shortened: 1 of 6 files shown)


MajorCooke:
Awesome! Thank you!

RicardoLuis0 (approved):

PR #1813, opened 2026-08-31
Reuse bone caches across portals

### LINKED ISSUE #1804: Bone caches are remade with every render pass, including portals
### UZDoom Version

5.0.0

### Game

All

### Operating System

Windows 10

### Duplicate Confirmation

### Bug Description

Last month I discovered an issue while working on a mod in an effort to replicate dynamic terrain. Bones used in models are not being copied and are instead recalculated for every single render pass, which means current models with bone support (in this case, IQM) can cause performance drops if portals are involved.

### Replication

 [Here's a demo mod.](https://drive.google.com/file/d/1UG7Qh5vHaFfvXkT6AEoAiS8HJHrXYWsS/view?usp=sharing)
* Load MAP01
* Type the following in console:
  * `r_portal_recursions 2`
  * `tile`

Changing `r_portal_recursions` will dramatically decrease performance the higher it goes. My prior attempt to fix it in #1792 demonstrates this.

(Recursions set to 2 is recommended because I don't want to accidentally fry anyone's hardware. It's very performance intensive. `tile` will spawn all the tiles.)

MajorCooke:
Reuse bone caches across portals, increasing performance whenever portals and models with bones are combined.

Fixes #1804 

## Checklist


src/r_data/models.cpp
```diff
@@ -637,20 +637,46 @@ static inline void RenderModelFrame(FModelRenderer *renderer, int i, const FSpri
...
-		if(frameinfo.smf_flags & MDL_MODELSAREATTACHMENTS || is_decoupled)
+		// [Jay] while per-model animations aren't done, DECOUPLEDANIMATIONS does the same as MODELSAREATTACHMENTS
+	if (!evaluatedSingle)
+	{ // [Jay] TODO per-model decoupled animations
+		//[MC] Grab bone caches and reuse them across portals if present
+		if (modelData && modelData->cachedBoneFrame == screen->mBoneCacheGeneration &&
+		    modelData->cachedBoneStartPos.SSize() > i && modelData->cachedBoneStartPos[i] >= 0)
+		{
+			boneStartingPosition = modelData->cachedBoneStartPos[i];
+			evaluatedSingle = true;
```
RicardoLuis0: again, it should only set evaluatedSingle if MDL_MODELSAREATTACHMENTS

src/r_data/models.cpp
```diff
@@ -637,20 +637,46 @@ static inline void RenderModelFrame(FModelRenderer *renderer, int i, const FSpri
...
+						modelData->cachedBoneStartPos.Resize(i + 1);
+						for (unsigned j = old; j < ui; j++)
+						{
+							modelData->cachedBoneStartPos[j] = -1;
+						}
+					}
+					modelData->cachedBoneStartPos[i] = boneStartingPosition;
+					modelData->cachedBoneFrame = screen->mBoneCacheGeneration;
+				}
+			}
```
RicardoLuis0: this needs to be moved outside of the MDL_MODELSAREATTACHMENTS block

src/r_data/models.cpp
RicardoLuis0: (also this should be covered by the block below, no need to special case it)

RicardoLuis0 (approved):

Boondorl:
Can you squash these two commits down into one? The second one seems to be fixing what the first one implemented

MajorCooke:
Done.

Boondorl (approved):

PR #1820, opened 2026-09-01
Push release tag automatically when version is updated

the-phinet:
Instead of having to manually push a tag after merging the version change, CI will be able to do this automatically now. The tag comes from the contents of VERSIONSTR

The commit in [the PR](https://github.com/the-phinet/UZDoom/pull/38):
image

From the [merged PR run](https://github.com/the-phinet/UZDoom/actions/runs/33460104383):
image

Tagged run properly triggers:
image

From the [tagged run](https://github.com/the-phinet/UZDoom/actions/runs/33460136528):
image

Properly built:
image

If it fails, it shouldn't hold anything up. We just need to manually push the tag

## Checklist

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -16,6 +16,7 @@ jobs:
       tag: ${{ steps.tag.outputs.tag }}
       mode: ${{ steps.emit.outputs.mode }}
       branch: ${{ steps.branch.outputs.branch || steps.tag.outputs.branch }}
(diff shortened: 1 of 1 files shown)


Boondorl:
Doesn't that make commits like f2b94c2f05535cb97bb1ec7c52c4f5d0941e309f impossible without creating a release? Or does it have a mechanism specifically to ignore trunk?

the-phinet:
Trunk is ignored when it comes to making releases. The tag will be pushed, but no release will be triggered

Boondorl (approved):

PR #1839, opened 2026-09-03
add warning to README

RicardoLuis0:
## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,9 @@ A continuation of [ZDoom][zdoom] and [GZDoom][gzdoom], UZDoom enhances the origi
 
 UZDoom is **free and open-source software**, built and maintained by a dedicated community of developers and enthusiasts.
 
+> [!CAUTION]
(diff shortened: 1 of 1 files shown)


River-Salmon (approved):

the-phinet:
I like the idea of this, but I don't think it should be cluttering the readme. I'd rather just have a pinned issue/discussion

RicardoLuis0:
> I like the idea of this, but I don't think it should be cluttering the readme. I'd rather just have a pinned issue/discussion

many people wouldn't look there though, the point of it being in the README is to be in somewhere visible to inform the most amount of people possible, since these are real issues we are currently dealing with (impersonator websites ran by third parties with possibly malicious intentions)

the-phinet:
The readme isn't going to keep people from accessing malicious sites. It's not like you need to open the repo before you navigate to a link on a search engine. Because of that, having an official notice in a non-intrusive place is just as good.

Boondorl:
I feel like this ends up being redundant because if someone is already here, then they likely know where to get it. This feels like something more sensible to put on the main site

the-phinet:
On the main website makes sense. Would another announcement on the forum (which will get mirrored on the main website) be fine? Or should a semi-permanent notice be added to the main page?

Closing this PR in favour of adding to the website in some way.

PR #1854, opened 2026-09-07
Render refactor refactor only (part 1 of ?)

RicardoLuis0:
no changes (intended) to engine behavior (if there _are_ any changes from this are bugs and should be fixed)

Edit: oh wait, there's _one_ change, the fix to autotextures, should i remove that commit and send it separately or just keep it here?

## Checklist


the-phinet (approved):
All the changes look good to me, but this is not an area I'm particularly knowledgeable in. I would prefer if someone who understands the render stack could have a look before merging.

the-phinet:
Offtopic to this pr:

I see this is adding Boost libraries! Can we go all-in on it? I would love to have access to their string manip stuff.

dileepvr:
How urgent is this for getting in? I'd like to test with some 3d-model heavy maps and some shader pk3s.

RicardoLuis0:
> Offtopic to this pr:
> 
> I see this is adding Boost libraries! Can we go all-in on it? I would love to have access to their string manip stuff.

it's only adding it because it's a header-only library, pulling the whole of boost would need a lot of changes to CMake that i'm not comfortable with doing

RicardoLuis0:
> How urgent is this for getting in? I'd like to test with some 3d-model heavy maps and some shader pk3s.

can't continue on actual feature work for the renderer without this, since i was told that it'd only be mergeable if the refactors were PR'd separately

dileepvr (approved):

PR #1862, opened 2026-09-09
use native fullscreen by default on macos

### LINKED ISSUE #1523: MacOS version not using Spaces for fullscreen
### UZDoom Version

5.0.0-pre-405-gd1423315af-m

### Game

All

### Operating System

Mac OS

### OS Version

27.0 DP2

### Hardware

M4 Pro, 16 inch 2024

### Duplicate Confirmation

### Bug Description

Currently, fullscreen UZDoom does not use Spaces. This leads to a couple of issues:

- Switching away will destroy the window, which breaks the program I use to switch tasks, creatively titled AltTab, since there is no longer a window.
- Compared to the current release, fullscreen now appropriately takes into account the notch. However, it does this by changing the entire space UZDoom is running on, so if you Cmd+Tab out, you now have less of a screen because the menu bar has been shifted below the notch.
- At least on MacOS 27, and probably 26, the fake fullscreen is probably causing a thin semi-transparent border to be drawn around the game. See attached screenshot, may need to open in a tab and zoom in and see it.

I'm not sure if there's any reason to be not using Spaces for fullscreen. Almost all major Mac game releases I've seen have used it for a long while it now, and the ones that don't tend to cause weird issues. Perhaps it can be gated via a default on cvar if someone strongly disagrees, but IMO this is the expected behavior for the platform nowadays.

I took a look around trying to see if I could fix it myself, but I only found a reference in `src/common/platform/posix/sdl/hardware.cpp`

```cpp
void I_InitGraphics ()
{
#ifdef __APPLE__
	SDL_SetHint(SDL_HINT_VIDEO_MAC_FULLSCREEN_SPACES, "0");
```

When I changed this to 1, or even just created an intentional syntax error, it still compiled successfully. I'm not sure if this code just isn't being called now, or what. Otherwise, I didn't see anything obvious as to why it wouldn't use Spaces. SDL_WINDOW_FULLSCREEN_DESKTOP should be using Spaces if specified, and that is appears to be what is being used.

(Also, I've got a paid Apple Developer account. I think y'all are set with notarization already since it didn't yell at me, but if it can be of any use to anyone, please get in touch!)

Image

### Replication

_No response_

### Configuration

```ini

```

### Log

_No response_

Yuki-Codes:
This change flips the default behaviour to use real fullscreen on MacOS as it gives a much better and more consistent OS experience. Using a borderless window is pretty rare in macOS games, and breaks several system features like Game mode and Spaces.

This is window code, not rendering, but to be safe I checked with both the OpenGL and Vulkan backends and native fullscreen works as expected on both.

Fixes #1523

## Checklist

### DIFF
--- a/src/common/platform/posix/cocoa/i_video.mm
+++ b/src/common/platform/posix/cocoa/i_video.mm
@@ -93,7 +93,7 @@ EXTERN_CVAR(Int,  vid_defheight)
 EXTERN_CVAR(Bool, vk_debug)
 
 CVAR(Bool, mvk_debug, false, 0)
-CVAR(Bool, vid_nativefullscreen, false, CVAR_ARCHIVE | CVAR_GLOBALCONFIG)
+CVAR(Bool, vid_nativefullscreen, true, CVAR_ARCHIVE | CVAR_GLOBALCONFIG)
 
 CUSTOM_CVAR(Bool, vid_autoswitch, true, CVAR_ARCHIVE | CVAR_GLOBALCONFIG | CVAR_NOINITCALL)
 {
--- a/src/gameconfigfile.cpp
+++ b/src/gameconfigfile.cpp
@@ -719,6 +719,16 @@ void FGameConfigFile::DoGlobalSetup ()
 				}
 			}
 		}
+
+		if(EngineLastRunVer < 234)
+		{
+			// Native fullscreen is now enabled by default, this is only used for MacOS.
+			var = FindCVar("vid_nativefullscreen", NULL);
+			if(var != NULL)
+			{
+				var->SetGenericRep(1, CVAR_Bool);
+			}
+		}
 	}
 
 	OkayToWrite = true;
--- a/src/version.h
+++ b/src/version.h
@@ -45,7 +45,7 @@
 // Version stored in the ini's [LastRun] section.
 // Bump it if you made some configuration change that you want to
 // be able to migrate in FGameConfigFile::DoGlobalSetup().
-#define ENGINELASTRUNVERSION "233"
+#define ENGINELASTRUNVERSION "234"
 #define GAMELASTRUNVERSION "1"
 
 // Protocol version used in demos.


RicardoLuis0 (changes requested):
this needs a save file migration as well, or else anyone who's upgrading from a previous config won't see any changes

save migration would be something along the lines of:

in `src/version.h`:
bump ENGINELASTRUNVERSION to 234
in `FGameConfigFile::DoGlobalSetup` in `src/gameconfigfile.cpp`, add:
```cpp
if(EngineLastRunVer < 234)
{
    var = FindCVar("vid_nativefullscreen", NULL);
    if(var != NULL)
    {
        var->SetGenericRep(1, CVAR_Bool)
    }
}
```

Yuki-Codes:
added that in, thanks!

RicardoLuis0 (approved):

RicardoLuis0:
should probably get an approval from @the-phinet as well, since i don't have a mac machine to test it personally

the-phinet:
It'll take some time for me to be able to test. My mac doesn't have a screen, so I need to borrow one to test notch-related stuff.

Yuki-Codes:
in my testing both options leave a black bar across the top of the screen to avoid the notch.
on 'borderless' you get a thin window border along that edge, and on native fullscreen you don't, but otherwise the same.

I'd like it if there was an option to go under the notch, actually. 🤔

the-phinet:
> in my testing both options leave a black bar across the top of the screen to avoid the notch.
on 'borderless' you get a thin window border along that edge, and on native fullscreen you don't, but otherwise the same.

I'm comfortable taking your word for it and merging it as-is. I'll let Jay make that call. If not, I'll merge once I can test.

> I'd like it if there was an option to go under the notch, actually. 🤔

Same. The current implementation is a bit of a hack until we get the engine to be aware of safe areas.

the-phinet (approved):
Works perfectly, thank you

Kinda annoying that this is an issue in the first place -- the behavior feels like a macos bug

Yuki-Codes:
its actually pretty common, MacOS is built to handle 'fullscreen' apps really elegantly, Windows isn't. borderless fullscreen is a hack to get around a windows limitation, it shouldn't be applied to Mac games. haha

PR #1866, opened 2026-09-09
add always skip intermission option

Yuki-Codes:
I'm not sure why 'always' wasn't an option for skipping the intermission screen, I couldn't find any discussion of it here, so I've added it in. please let me know if there was a reason.

This option (sv_alwaystally 3) always skips the intermission score screen (except in deathmatch). it does not affect the cutscenes or text screens in my testing.

https://github.com/user-attachments/assets/664416ef-cbf3-4d04-8cd7-52dc52c5a489

(The level name on load is a mod, and not part of this change)

## Checklist

### DIFF
--- a/src/g_level.cpp
+++ b/src/g_level.cpp
@@ -703,6 +703,9 @@ bool FLevelLocals::ShouldDoIntermission(cluster_info_t* nextcluster, cluster_inf
 	if ((sv_alwaystally == 2) || (deathmatch))
 		return true;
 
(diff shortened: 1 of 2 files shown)


the-phinet (approved):

the-phinet:
No need to do merges unless there's a conflict. We'll do that when we merge in the pr. (Also try to use rebase instead of merge when syncing)

the-phinet (approved):

Boondorl (approved):

PR #1869, opened 2026-09-10
Make magic color numbers less magic

the-phinet:
Makes hard-coded colors display as a color, not as a number.

It doesn't change any handling of the colors, just how the are input. `#rgb` and `#rrggbb` forms of entry are supported, which many text editors will render as an inline color.

The string processing happens at compile time, so there is no runtime cost to this.

CI needed to be updated, since the specific version of msvc the arm runner was using has a bug where consteval stuff in an array initializer is falsely accused of not being consteval.

|Old|New|
|---|---|
| image image | image image |

Every color I could find in the codebase that was in the form of a single int (in the correct byte order) has been replaced. The only candidates that I did not touch was the color white when it was being used as a bitmask.

A later pr will tackle integer triplets to make them render as colors.

## Checklist


the-phinet:
Rebased to fix conflict. No other changes

TehRealSalt (approved):

TehRealSalt:
Awesome!!

PR #1870, opened 2026-09-10
Save Load screen adjustments + graphics

Yuki-Codes:
Before I get too far into messing with this screen, can we use graphics here, or were they intentionally excluded for compatibility reasons? 🤔 

Screenshot 2026-09-10 at 22 27 52

I'm also looking for feedback on these graphics and implementation details specifically:

- Using a m_warn icon when a save file is missing wads, with a description in the info box.
- Using a m_error icon when a save file is an old version, also with a description in the info box.
- Adding some m_corner decorations to the frames:

I've drawn these myself, using dooms original color palette (or something close to it) so they likely could use improvement.

## Checklist


RicardoLuis0:
IMHO it should still use text colors as well

SanyaWaffles:
I like this, I do think using the textcolors as well would make this solid. :)

mc776:
I like the warning symbols though the corners seem a bit unnecessary and distracting...

Yuki-Codes:
🫡 
I had initially removed the coloured text because the gradient the console font uses made it look a bit weird imo.
happy to put it back in, but does anyone have thoughts on using the NewSmallFont instead of the NewConsoleFont for the list elements? the colours come out much better on the solid font:
Screenshot 2026-09-11 at 10 05 57
Screenshot 2026-09-11 at 10 05 19

the-phinet:
I prefer it without the text colour. The coloured text looks like it's trying to highlight it, not mark it as a potential issue.

I also like the corners, especially if they easily overwritten.

mc776:
What about reversing the colours so that the brighter yellow indicates something that's ready to go?

I also prefer the smallfont2 though I wonder if there's any technical concern to the contrary...

Yuki-Codes:
I'm quite fond of using white text with the brick highlight colour (to match the options menu), perhaps tinting the warning text itself yellow to be a bit more attention grabbing is the go?

Screenshot 2026-09-11 at 13 00 30

or we could try to emulate the options menu style more closely by dropping the background of the file list entirely and using the selection cursor instead of the row highlight, which I also quite like:
Screenshot 2026-09-11 at 13 20 33

the-phinet:
> I'm quite fond of using white text with the brick highlight colour (to match the options menu), perhaps tinting the warning text itself yellow to be a bit more attention grabbing is the go?

I like this one the best!

> or we could try to emulate the options menu style more closely by dropping the background of the file list entirely

And I like this one too

TimRichter777:
> I'm quite fond of using white text with the brick highlight colour (to match the options menu), perhaps tinting the warning text itself yellow to be a bit more attention grabbing is the go?

I think this looks really good

>or we could try to emulate the options menu style more closely by dropping the background of the file list entirely and using the selection cursor instead of the row highlight, which I also quite like:

Still looks good, but i think that the first option looks better

Boondorl:
Only complaint I have is the warning and old version icons are the same. Imo these should have different graphics entirely to be more color blind friendly (e.g. flipping the old save icon upside down to make it clearer it's a different icon)

the-phinet:
> Only complaint I have is the warning and old version icons are the same. Imo these should have different graphics entirely to be more color blind friendly

Good call.

Maybe something like this for the version error? (I'm not much of a pixel artist, so it probably needs some tlc)
image image

Edit:

If a round icon gets used for one, I would sharpen the triangular one out like this maybe
image image

Yuki-Codes:
Screenshot 2026-09-13 at 14 57 32

Cool! I'm happy to open this up for full review. 😄 
thanks for the valuable input all!

the-phinet:
Here's the version error icon in-situ
Screenshot_20260913_223035

the-phinet (approved):

heiko123abc:
Looks really great. But I think the icons aren't centered in the line in the save file selector

TehRealSalt (approved):
Fabulous work!!

PR #1875, opened 2026-09-11
Automap match statusbar (+text padding)

Yuki-Codes:
This change introduces a new automap option and cvar: am_matchStatusbar, which when enabled, causes the automap screen to use the same status bar setting as the game screen, so fullscreen hud or no hud can be used on the automap:

Screenshot 2026-09-12 at 02 31 29
Screenshot 2026-09-12 at 02 28 37
Screenshot 2026-09-12 at 02 30 26

This change does not affect the alternative HUD, that always draws the automap fullscreen and uses its own overlay text.

(also also adjusted the top padding of the upper text while I was in here)

## Checklist

### DIFF
--- a/src/am_map.cpp
+++ b/src/am_map.cpp
@@ -162,6 +162,7 @@ CVAR(Bool, am_showkeys, true, CVAR_ARCHIVE);
 CVAR(Int, am_showtriggerlines, 0, CVAR_ARCHIVE);
 CVAR(Int, am_showthingsprites, 0, CVAR_ARCHIVE);
 CVAR (Bool, am_showkeys_always, false, CVAR_ARCHIVE);
+CVAR(Bool, am_match_statusbar, true, CVAR_ARCHIVE)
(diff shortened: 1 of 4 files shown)


src/am_map.cpp
```diff
@@ -162,6 +162,7 @@ CVAR(Bool, am_showkeys, true, CVAR_ARCHIVE);
 CVAR(Int, am_showtriggerlines, 0, CVAR_ARCHIVE);
 CVAR(Int, am_showthingsprites, 0, CVAR_ARCHIVE);
 CVAR (Bool, am_showkeys_always, false, CVAR_ARCHIVE);
+CVAR(Bool, am_matchStatusbar, false, CVAR_ARCHIVE)
```
the-phinet: Defaulting to true probably makes sense. Also we try to keep cvars in snake_case

the-phinet (approved):

PR #1884, opened 2026-09-12
-continuegame command line argument

Yuki-Codes:
this change introduces a new command line argument that loads the last saved game automatically.
`uzdoom.app -continuegame`

## Checklist


RicardoLuis0 (commented):
this needs to actually check if the loaded mods and load order match to be useful

Yuki-Codes:
oh like: "load the last game that matched the mod settings"? 🤔 

Edit: oh duh it would need to at the very least match the iwad, right. 🤦‍♀️

Ya, let's give up on this one for now.

PR #1886, opened 2026-09-12
Option search improvements

the-phinet:
- Add `*` as a way to see all options. This is to help people looking for something that they have no idea what it might even be called. It's also for me to be able to find untranslated options quickly.
- Searches will include results when searching in English when the engine is using a different language. This is to allow us to give tech support to people when we don't know what an option might be called in their language.

## Checklist


src/common/scripting/interface/stringformat.cpp
```diff
@@ -270,6 +270,26 @@ DEFINE_ACTION_FUNCTION(FStringStruct, DeleteLastCharacter)
 //
 //=====================================================================================
 
+static void GetString(const FString &label, bool prefixed, FString *result)
```
Boondorl, line 273: Is this legal? I know references are just pointers under the hood, but I'm not sure if the VM and JIT will play nicely with using a reference directly

the-phinet: Should be fine. It's the same as LocalizeString right below it.

Boondorl (approved):

PR #1896, opened 2026-09-13
Clean up external ZMusic usage and force BUILD_SHARED_LIBS=OFF

chewi:
Using an external ZMusic is not supported anymore. This remaining code was causing libzmusic.a to be installed with the rest of UZDoom.

Disabling shared libraries was previously done for ZMusic, but it makes no sense to build them at all, and this should therefore be enforced as early as possible. UZDoom has no libraries, and if not enforced, bundled dependencies like cppdap and abseil will be built as shared libraries that don't get installed, leaving UZDoom broken.

## Checklist


RicardoLuis0:
why is this removing FModule? that's also used for curl/openal/etc

the-phinet:
> why is this removing FModule? that's also used for curl/openal/etc

I was surprised that it was still compiling. Looks like the i_module file in zmusic gets used instead. Maybe the wrong one was removed?

chewi:
That was intentional. The one in ZMusic is identical except it adds `FModule_GetProgDir`. ZMusic isn't optional, so surely we only need one of them?

the-phinet:
Okay, we should just remove the one in zmusic instead. Kinda weird to be relying on a part of zmusic in a non-zmusic place.

chewi:
Okay, but `FModule_GetProgDir` is called from a few places within ZMusic. I'm not familiar enough with this to suggest how this should be addressed.

RicardoLuis0:
at least it should be fully removed, headers included, and the zmusic headers included anywhere that the then-deleted headers were used, that way the fact that it's taken from zmusic is obvious instead of just because it happens to match the source in both headers

chewi:
Ah, I hadn't thought to check for a header. I've now dropped that and added libraries/ZMusic/source/loader to the library's interface includes. If you're happy with that, I can submit the same change in the ZMusic repo. I'm assuming you don't want to actually install i_module.h alongside zmusic.h, not that anyone would install ZMusic system-wide now.

The i_module changes have now been made in a separate commit for clarity.

chewi:
Okay, submitted that change in UZDoom/ZMusic#20.

the-phinet (approved):

PR #1916, opened 2026-09-17
Configuration fixes to be able to compile in all configs on macOS

the-phinet:
When compiled directly through xcode `_LIBCPP_HARDENING_MODE` is enabled, which causes compilation failure due to large static assertions creating too many instructions. This enables it for all configurations, but raises the constexpr instruction limit to account for it.

It also makes uzdoom aware of macos 27.

## Checklist


TehRealSalt (approved):
(As usual, my Mac related approvals are just under the pretense that it looks okay & I have no means to test ^^;)

PR #1921, opened 2026-09-20
Add workaround to allow compilation on older MSVC versions

the-phinet:
## Checklist


RicardoLuis0:
why not just use an `#if _MSC_VER < 1951` check directly?

the-phinet:
> why not just use an `#if _MSC_VER < 1951` check directly?

It causes an annoying amount of spam in the build log

https://github.com/the-phinet/UZDoom/actions/runs/35470142178/job/105969495978#step:12:1

RicardoLuis0:
> > why not just use an `#if _MSC_VER < 1951` check directly?
> 
> It causes an annoying amount of spam in the build log
> 
> https://github.com/the-phinet/UZDoom/actions/runs/35470142178/job/105969495978#step:12:1

it should probably just be set automatically based on version on cmake though, rather than being an explicit option

RicardoLuis0:
older msvc versions _will not_ get this as a backport, microsoft never does that to force people to use newer visual studio versions, all it backports are security fixes, so having a way to turn it off isn't necessary

the-phinet:
> older msvc versions will not get this as a backport

That's really annoying. I added the option so I can check in a couple months if the workaround can be removed before 2022 goes eol. I'll remove the option 🫡

the-phinet:
Should I switch back to the default ARM windows runner? It was swapped to the explicit "I want VS2026" one because of the bug. 25a390263bd15cea4a5773cc39259065395d71f1

RicardoLuis0:
> Should I switch back to the default ARM windows runner? It was swapped to the explicit "I want VS2026" one because of the bug. [25a3902](https://github.com/UZDoom/UZDoom/commit/25a390263bd15cea4a5773cc39259065395d71f1)

if there's a speed difference, we should go with the fastest one, otherwise, if they're the same speed the vs22 one might be good that way we have coverage for both vs version

RicardoLuis0 (approved):

the-phinet:
> if there's a speed difference, we should go with the fastest one, otherwise, if they're the same speed the vs22 one might be good that way we have coverage for both vs version

There's only a minor cost at startup. I'll switch it back in a different pr.

PR #1922, opened 2026-09-20
Fixed FTagManager::SectorHasTag erroneously returning false when checking if an untagged sector had a tag of zero.

AKMDM:
This particularly fixes a regression with the **ThingCountSector** and **ThingCountNameSector** ACS functions in which both of them always returned zero if the tag argument was zero, even when there's actors with the correct name and TID in those sectors. This regression first originated when support for multi-tag sectors/lines was added many years ago. Here's an minimal example WAD to test the fix: [ThingCountTest.zip](https://github.com/user-attachments/files/32429461/ThingCountTest.zip)

1. Run **ThingCountTest.wad** and load **MAP01**.
2. Test each of the three switches. The non-zero tagged sectors with imps and hell knights in them will return the correct counts for each, but the zero-tagged sectors with zombies in them will always return zero, even though there's five of them.

## Checklist
- [✓] I have reviewed [CONTRIBUTING.md](../blob/trunk/CONTRIBUTING.md) and agree to its terms.


RicardoLuis0 (approved):

Gutawer:
Kinda curious about the use case here, because in my thinking "Tag 0" doesn't really exist, it's kinda just a sentinel value for "no tag" and so the current behavior seems correct already. But if this is causing niche regressions compared to other source ports without the multi-tag support then that's fair but fixing this on the SectorHasTag level to make a couple functions change behavior feels a bit risky when those functions are called in multiple other places (and we have the SectorHasTags function which implies that the semantics already are "a sector can be completely untagged", not "a sector will always have an implicit 0 tag")

RicardoLuis0 (changes requested):
> Kinda curious about the use case here, because in my thinking "Tag 0" doesn't really exist, it's kinda just a sentinel value for "no tag" and so the current behavior seems correct already. But if this is causing niche regressions compared to other source ports without the multi-tag support then that's fair but fixing this on the SectorHasTag level to make a couple functions change behavior feels a bit risky when those functions are called in multiple other places (and we have the SectorHasTags function which implies that the semantics already are "a sector can be completely untagged", not "a sector will always have an implicit 0 tag")

oh, good point, a better fix for this would only change it for ACS

PR #1924, opened 2026-09-21
Remove c++14 detection for zmusic

the-phinet:
The project already specifies c++17, making the checks pointless. Also there are some warnings caused by the compiler settings being forcefully overwritten

I'll add this to the zmusic repo if/when this gets merged

## Checklist


RicardoLuis0:
zmusic is still meant to be usable as a separate library though (as in, for non-uzdoom things, not as in a dll for uzdoom), this should only be considered if it's gonna be fully folded into uzdoom

the-phinet:
> zmusic is still meant to be usable as a separate library though (as in, for non-uzdoom things, not as in a dll for uzdoom), this should only be considered if it's gonna be fully folded into uzdoom

zmusic itself enforces c++17 (UZDoom enforces c++20). The c++14 checks are from before that

RicardoLuis0 (approved):

PR #1936, opened 2026-09-23
Don't save `bigstate`

### LINKED ISSUE #1935: Loading save from previous UZDoom version gives "boolean type expected for 'bigstate'"
### UZDoom Version

5.0.2

### Game

Doom

### Operating System

Windows 11

### OS Version

Windows 11 Pro

### Hardware

Ryzen 7 7800X3D, Sapphire PULSE Radeon RX 7900XTX 24GB

### Duplicate Confirmation

### Bug Description

I was doing a run of TNT: Evilution and had to update, so I let the launcher do that. I was on version 5.0.1 at the time. I recently updated, and all the sudden my save refuses to load and gives me the error of:

boolean type expected for 'bigstate'
1 errors parsing JSON

### Replication

1. Open UZDoom
2. Load the included save file (leaving at as save00.zds, it should appear in game as TNT. Not renaming it because I don't know if that'll break anything)
3. Read em and weep

[tnt save00 zds.zip](https://github.com/user-attachments/files/32546563/tnt.save00.zds.zip)

### Configuration

```ini
https://gist.github.com/YKWA/3ebff96f392c560113146ace3d351e37
```

### Log

https://gist.github.com/YKWA/c081e6e6357a621e369e12d1f851967e

the-phinet:
Fixes #1935 

Saving map ui state is weird. This feels like it should be at most saved as a global cvar? The only automap stuff that makes sense in a save file is map progress.

## Checklist


RicardoLuis0:
bigstate isn't a cvar, it's the current state of the automap, a proper fix would be to read it manually, not to not save it, not saving it needs to be done as a later refactor (ex. you open the automap, save, and when you reload it stays open, it's not global state)

RicardoLuis0:
> a proper fix would be to read it manually

ex.:
```cpp
if(arc.isReading() && arc.HasKey("bigstate") && arc.r->FindKey("bigstate")->IsInt())
{
	int tmp;
	arc("bigstate", tmp);
	bigstate = !!tmp;
}
else
{
	arc("bigstate", bigstate);
}
```
(might need to have an IsInt(key) added to FSerializer that does `return HasKey(key) && r->FindKey(ket)->IsInt()`)

the-phinet:
If that's the fix you want, open a PR with it. I'd rather not save engine ui state at all in a game savefile

RicardoLuis0:
> If that's the fix you want, open a PR with it. I'd rather not save engine ui state at all in a game savefile

i'm saying, that's well and good but it isn't just a fix, it's removing only one part of the state while still saving the rest, if the ui is to be decoupled from saves, it should be done fully, not halfway with a hacksaw, and that isn't fit for a hotfix

the-phinet:
Jay, there's no reason to throw accusations around. Keep things professional please.

the-phinet:
Closing in favour of #1937

